PR #48, opened 2021-06-02
Support symbol rename refactoring using rascal-core

DavyLandman:
Just a PR to track the progress.

Todo:

- [ ] adapt rascal-core to support the [`rename` function](https://github.com/usethesource/rascal-core/commit/e3819eb15569e5d373de778a49eb2ebc67134b8a) (@PaulKlint is aware of this)
- [ ] make a release of rascal-core we can use in for rascal-vscode (this depends on a few bugfixes)
- [ ] test the integration (especially around surrogate pairs)

Note that the LSP api also supports create/rename/deletion of files, but that would require we start tracking Document versions throughout our system. That is a good thing to have, but for now, this interface is simpler (yet a bit less powerful).


DavyLandman:
Closing this PR, since @toinehartman has been working on a proper rename refactoring the past few monhts. See tracking issue #382.

PR #54, opened 2021-06-28
Implementing File System Provider

DavyLandman:
I opened up a PR for easy reviewing comments.

@jurgenvinju did all the work in this branch


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -260,5 +274,120 @@ public void connect(LanguageClient client) {
...
+        public CompletableFuture<Void> createDirectory(URIParameter uri) throws IOException, URISyntaxException {
+            ISourceLocation loc = uri.getLocation();
+            reg.mkDirectory(loc);
+            return CompletableFuture.completedFuture(null);
+        }
+
+        @Override
+        public CompletableFuture<LocationContent> readFile(URIParameter uri) throws URISyntaxException {
+            ISourceLocation loc = uri.getLocation();
+            return CompletableFuture.completedFuture(new LocationContent(Prelude.readFile(IRascalValueFactory.getInstance(), false, loc).getValue()));
```
DavyLandman, line 345: A bit wastefull to build a `IString` just to do `getValue` on it (I mean, I like the reuse of prelude library)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -260,5 +274,120 @@ public void connect(LanguageClient client) {
...
+
+            if (!reg.exists(loc) && !params.isCreate()) {
+                throw new FileNotFoundException(loc.toString());
+            }
+
+            if (!reg.exists(URIUtil.getParentLocation(loc)) && params.isCreate()) {
+                throw new FileNotFoundException(URIUtil.getParentLocation(loc).toString());
+            }
+
+            if (reg.exists(loc) && params.isCreate() && !params.isOverwrite()) {
```
DavyLandman, line 360: file checks like "exist"  are a bit expensive, we could ask it once at the top of the method to reduce IO.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -260,5 +274,120 @@ public void connect(LanguageClient client) {
...
+                throw new FileAlreadyExistsException(loc.toString());
+            }
+
+            reg.getOutputStream(loc, false).write(params.getContent().getBytes(Charset.forName("UTF8")));
```
DavyLandman, line 364: OutputsStream is not closed!

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -260,5 +274,120 @@ public void connect(LanguageClient client) {
...
+            return CompletableFuture.completedFuture(null);
```
DavyLandman, line 365: All these exceptions are lost to the client, rewriting the code to schedule the write on the executor, and make a completable future that either returns void, or throws an exception, they bubble through better.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -260,5 +274,120 @@ public void connect(LanguageClient client) {
...
+        }
+
+        @Override
+        public CompletableFuture<Void> delete(DeleteParameters params) throws IOException, URISyntaxException {
+            ISourceLocation loc = params.getLocation();
+            reg.remove(loc, params.isRecursive());
+            return CompletableFuture.completedFuture(null);
```
DavyLandman, line 372: similarly, the exceptions should not be lost, they should be connected to the completablefuture.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IBaseLanguageServerExtensions.java
```diff
@@ -42,16 +49,208 @@
...
+    default CompletableFuture<Void>  delete(DeleteParameters params) throws IOException, URISyntaxException {
         throw new UnsupportedOperationException();
     }
 
+    @JsonRequest("rascal/filesystem/rename")
+    default CompletableFuture<Void>  rename(RenameParameters params) throws IOException, URISyntaxException {
+        throw new UnsupportedOperationException();
+    }
+
+    public static class DeleteParameters {
```
DavyLandman, line 107: Are these not build-in to lsp4j? or are there a  mirror of the vscode api? (in that case, add a link to the classes they mirror)

rascal-vscode-extension/src/RascalFileSystemProviders.ts
```diff
@@ -0,0 +1,113 @@
...
+     * Registers a single FileSystemProvider for every URI scheme that Rascal supports, except
+     * for file, http and https.
+     *
+     * @param client to use as a server for the file system provider methods
+     */
+    constructor (client:LanguageClient) {
+        this.client = client;
+
+        client.onNotification("rascal/filesystem/onDidChangeFile", (event:vscode.FileChangeEvent) => {
+            this._emitter.fire([event]);
```
DavyLandman, line 49: Why is this happening? what does it do?

rascal-vscode-extension/src/RascalFileSystemProviders.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    registerSchemes(schemes:string[]):void {
+        schemes
+            .filter(s => !this.protectedSchemes.includes(s))
+            .forEach(s => vscode.workspace.registerFileSystemProvider(s, this));
+    }
+
+    watch(uri: vscode.Uri, options: { recursive: boolean; excludes: string[]; }): vscode.Disposable {
+        // TODO: fix this
```
DavyLandman, line 60: what should be fixed?

rascal-vscode-extension/src/extension.ts
```diff
@@ -34,7 +34,10 @@ import * as os from 'os';
 
 import {LanguageClient, LanguageClientOptions, ServerOptions, StreamInfo, integer } from 'vscode-languageclient/node';
 import { TextDocumentContentProvider } from 'vscode';
+import { RascalFileSystemProvider } from './RascalFileSystemProviders';
 import { RascalTerminalLinkProvider } from './RascalTerminalLinkProvider';
+import { ClientRequest } from 'node:http';
+import { SSL_OP_SSLEAY_080_CLIENT_DH_BUG, WSAEFAULT } from 'node:constants';
```
DavyLandman, line 40: I think this import can be dropped?

rascal-vscode-extension/src/extension.ts
```diff
@@ -134,6 +118,12 @@ export function activateLanguageClient(context: vscode.ExtensionContext, languag
                 registerLanguage(context, parametricServer, lang);
             });
         }
+
+        let schemesReply:Promise<string[]> = client.sendRequest("rascal/filesystem/schemes");
+
+        schemesReply.then( (schemes:string[]) => {
```
DavyLandman, line 124: nitpick: Th `:string[]` part can be dropped, as it's inferred from the type.

DavyLandman (commented):
A few bugs, and a few nitpicks, but looks good to me, quite clean overal..

rascal-lsp/pom.xml
```diff
@@ -59,12 +59,12 @@
...
       <groupId>org.rascalmpl</groupId>
       <artifactId>rascal</artifactId>
-      <version>0.19.4-RC1</version>
+      <version>0.19.4-RC2</version>
     </dependency>
     <dependency>
       <groupId>org.rascalmpl</groupId>
       <artifactId>rascal-core</artifactId>
-      <version>0.4.16</version>
+      <version>0.4.17</version>
```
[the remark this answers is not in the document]
jurgenvinju: the old version is also very much broken. it has cyclic extends, it reports the errors of other modules instead of the current one. etc.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -150,18 +165,37 @@ private static void startLSP(Launcher<IBaseLanguageClient> server) {
...
+            URIResolverRegistry.getInstance().registerLogical(new ProjectURIResolver(this::resolveProjectLocation));
+            URIResolverRegistry.getInstance().registerLogical(new TargetURIResolver(this::resolveProjectLocation));
+        }
+
+        private ISourceLocation resolveProjectLocation(ISourceLocation loc) {
+            try {
+                for (WorkspaceFolder folder : workspaceFolders) {
+                    if (folder.getName().equals(loc.getAuthority())) {
+                        ISourceLocation root = URIUtil.createFromURI(folder.getUri());
+                        return URIUtil.getChildLocation(root, loc.getPath());
```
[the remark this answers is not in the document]
jurgenvinju: no the registry takes care of carrying over the offsets when resolving the logical schemes

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -260,5 +274,120 @@ public void connect(LanguageClient client) {
             this.ideServicesConfiguration = IDEServicesThread.startIDEServices(this.client);
             getTextDocumentService().connect(this.client);
         }
+
+        private final URIResolverRegistry reg = URIResolverRegistry.getInstance();
+
+        // BELOW THE FILESYSTEM SERVICE:
```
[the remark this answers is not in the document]
jurgenvinju: yes, but first I want to see the IDE fully working.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IBaseLanguageServerExtensions.java
jurgenvinju: nope.

rascal-vscode-extension/src/RascalFileSystemProviders.ts
jurgenvinju: this is when a folder is being watched and the watch implementation reports that a change has happened. The listeners inside vscode expect a FileChangeEvent to be triggered.

jurgenvinju: so it's basically passing on the event from the Java based part back into the TypeScript world.

rascal-vscode-extension/src/RascalFileSystemProviders.ts
jurgenvinju: it was fixed, but the comment was not removed.

rascal-vscode-extension/src/extension.ts
jurgenvinju: I have no idea what this is.

rascal-vscode-extension/src/extension.ts
jurgenvinju: sure. I needed it to write the code, but if it can be dropped, by all means.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -150,18 +165,37 @@ private static void startLSP(Launcher<IBaseLanguageClient> server) {
...
-
-    private static class ActualLanguageServer implements LanguageServer, LanguageClientAware, IBaseLanguageServerExtensions {
+    private static class ActualLanguageServer  implements IBaseLanguageServerExtensions, LanguageClientAware {
         static final Logger logger = LogManager.getLogger(ActualLanguageServer.class);
         private final IBaseTextDocumentService lspDocumentService;
         private final BaseWorkspaceService lspWorkspaceService = new BaseWorkspaceService();
         private final Runnable onExit;
         private IBaseLanguageClient client;
         private IDEServicesConfiguration ideServicesConfiguration;
+        private List<WorkspaceFolder> workspaceFolders = Collections.emptyList();
```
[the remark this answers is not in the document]
jurgenvinju: currently the only assignment is in the initialize method, so I think this would be safe no?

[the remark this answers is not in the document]
jurgenvinju: nevermind, somebody will start updating it in the future so let's fix it.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -150,18 +165,37 @@ private static void startLSP(Launcher<IBaseLanguageClient> server) {
...
             this.lspDocumentService = lspDocumentService;
+            URIResolverRegistry.getInstance().registerLogical(new ProjectURIResolver(this::resolveProjectLocation));
+            URIResolverRegistry.getInstance().registerLogical(new TargetURIResolver(this::resolveProjectLocation));
+        }
+
+        private ISourceLocation resolveProjectLocation(ISourceLocation loc) {
+            try {
+                for (WorkspaceFolder folder : workspaceFolders) {
+                    if (folder.getName().equals(loc.getAuthority())) {
+                        ISourceLocation root = URIUtil.createFromURI(folder.getUri());
```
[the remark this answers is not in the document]
jurgenvinju: this code was expensive when we communicated for every resolution with the LSP client, but now we get the folders once at init time so this point is resolved.

PR #97, opened 2021-11-30
Store ExtensionContext upon activation

rodinaarssen:
The extension context is no longer propagated to most functions; instead, it is stored upon activation and used throughout `extension.ts`.


DavyLandman:
Nice clean approach 👍🏼

DavyLandman:
maybe we should clear the context variable (and all the others) when the extension is disposed? to avoid leaking memory?

jurgenvinju (commented):
looks clean! thanks

PR #100, opened 2021-12-02
Changing language mode of open editors for newly registered languages

rodinaarssen:
When a language is registered, open editors are not automatically updated. This change checks whether there are active editors at registration time


rodinaarssen:
Oops, branching fail.

PR #103, opened 2021-12-02
Delaying launching Rascal terminals until the language server is ready

rodinaarssen:
Currently, if the Start Rascal Terminal command is issued while the Rascal server is not already running, no terminal appears. This change checks whether the language server is ready first.

I would not be surprised if there would be a somewhat less convoluted way to express this though.

### DIFF
--- a/rascal-vscode-extension/src/extension.ts
+++ b/rascal-vscode-extension/src/extension.ts
@@ -215,26 +215,23 @@ function registerImportModule() {
     rascalExtensionContext!.subscriptions.push(command);
 }
 
-function startTerminal(uri: vscode.Uri, ...extraArgs: string[]) {
+async function startTerminal(uri: vscode.Uri, ...extraArgs: string[]) {
     if (!rascalClient) {
         rascalClient = activateRascalLanguageClient();
     }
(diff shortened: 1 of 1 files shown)


rascal-vscode-extension/src/extension.ts
```diff
@@ -220,19 +220,21 @@ function startTerminal(uri: vscode.Uri, ...extraArgs: string[]) {
...
-    ]).then(async cfg => {
-        let terminal = vscode.window.createTerminal({
-            cwd: path.dirname(uri.fsPath),
-            shellPath: await getJavaExecutable(),
-            shellArgs: buildShellArgs(cfg[1], cfg[0], ...extraArgs),
-            name: 'Rascal Terminal',
-        });
+    rascalClient.then(c => c.onReady().then(() => {
+        Promise.all([
+            rascalClient!.then(c => c.sendRequest<IDEServicesConfiguration>("rascal/supplyIDEServicesConfiguration")),
```
DavyLandman, line 225: You could move the on ready part to this then chain.

We could also make this functions async, so we can just call them with an await and avoid al this chaining of then calls.

Issue #145, opened 2022-03-07
Provide TextDecorations options for DSLs and other files


DavyLandman:
In eclipse we had the [`util::Editors`](https://github.com/usethesource/rascal-eclipse/blob/main/rascal-eclipse/src/org/rascalmpl/eclipse/library/util/Editors.rsc) module, which would allow the user to add background coloring to any file in the Eclipse IDE.

VSCode has a similar feature called [`TextEditorDecorationType`](https://code.visualstudio.com/api/references/vscode-api#TextEditorDecorationType).  We're already using them for the InlayHint (#119) feature.

There are two ways I want to support this:

1. A user can register a function for a arbitrary file, after that point, VSCode will ask decorations options specific for that file to this function. 
2. DSLs can provide a decorator as a `LanguageService`. This is a function that is called to decorate the file.

The DSL case is quite easy to add, but the first one is a bit harder, since we would be connecting the REPL more tightly with the VSCode IDE.

Secondly, there is an API question. VSCode api allows a subset of css styling for the decorations. The advice is to make a new class/type per different style, so that VSCode won't get to slow with all the inline styling. I see two options:

1. Keep the old info/warning/error styles from eclipse
2. Provide a map during registration of the decorator the the different styles we can expect.

jurgenvinju:
Don't think we need option 2 if we have option 1. This could be a simple side effect of summarizer or parser. We could add it for clarity though.

jurgenvinju:
It would be good if we register a default set of styles (info, warning, error) that people can just use and then optionally let them contribute more styles via one of the Contributor constructors

DavyLandman:
To be clear, option 1 was, you could register a decoration for _any_ file, even one that's not a file that we're producing any services on. Like a java file or a css file. This is very generic, and a lot harder to get the plumbing right.

Option 2 would fit with the summaries indeed.

jurgenvinju:
It doesn't matter from our LSP side for which file the registration is. IDEservices will forward the request to the client asynchronously. If this is not an actual IDE service from VScode yet, then I understand this will be harder to plumb. In fact, we should consider not doing that since adding such a feature does imply that for every new client we will have to port it. I'd rather stay as close to the LSP and basic services as possible.

jurgenvinju:
[...]

PR #152, opened 2022-04-21
Exposing VS Code decorators for DSLs

DavyLandman:
As discussed in #145 we would like to bring decorations features to the DSLs in rascal.

Here is my proposal, please review.

The "biggest" problem I have with this design (and why it's most likely not part of LSP) is that it's quite specific to how VS Code works. What do you think of this trade-off?

The API models how VS code roughly want's you to define css-classes, and then annotate parts of the tree with css classes, and then if needed, you are allowed to customize some property per range on the tree. This is done for performance reasons.

### DIFF
--- a/rascal-lsp/src/main/rascal/util/LanguageServer.rsc
+++ b/rascal-lsp/src/main/rascal/util/LanguageServer.rsc
@@ -35,6 +35,7 @@
 module util::LanguageServer
 
 import util::Reflective;
+import util::Maybe;
 
 import ParseTree;
 
@@ -50,6 +51,11 @@ alias LensDetector     = rel[loc src, Command lens] (Tree /*input*/);
 alias CommandExecutor  = void (Command /*command*/);
 alias InlayHinter      = list[InlayHint] (Tree /*input*/);
 
+// a decorator returns the current decorations for this tree,
+// per invocation you have to repeat all styling for this file,
+// otherwise they are cleared.
+alias Decorator        = map[str baseStyle, lrel[loc range, Maybe[MarkdownString] hoverMessage, set[InstanceDecoration] extraDecorations]] (Tree /*input*/);
+
 @synopsis{Each kind of service contibutes the implementation of one (or several) IDE features.}
 data LanguageService
     = parser(Parser parser)
@@ -60,6 +66,7 @@ data LanguageService
     | lenses(LensDetector detector)
     | inlayHinter(InlayHinter hinter)
     | executor(CommandExecutor executor)
+    | decorator(map[str styleCategory, set[Decoration] definitions] baseStyles, Decorator decorator)
     ;
 
 @synopsis{A model encodes all IDE-relevant information about a single source file.}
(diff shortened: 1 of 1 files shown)


rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -50,6 +51,11 @@ alias LensDetector     = rel[loc src, Command lens] (Tree /*input*/);
 alias CommandExecutor  = void (Command /*command*/);
 alias InlayHinter      = list[InlayHint] (Tree /*input*/);
 
+// a decorator returns the current decorations for this tree,
+// per invocation you have to repeat all styling for this file,
+// otherwise they are cleared.
+alias Decorator        = map[str baseStyle, lrel[loc range, Maybe[MarkdownString] hoverMessage, set[InstanceDecoration] extraDecorations]] (Tree /*input*/);
```
jurgenvinju, line 57: does this mean that the strings have to be in the order of appearance in the file? Otherwise better to use `rel`. If this is required, then runtime asserts in debug mode should help the user finding issues

jurgenvinju: can the Maybe be avoided?

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -137,5 +144,111 @@ data InlayKind // this determines style
...
+    ;
+
+data InstanceDecorationOptions
+    = instanceAfter(AttachableDecorationOptions tdaro)
+    | instanceBefore(AttachableDecorationOptions tdaro)
+    ;
+
+// css properties that decorators can change (in relation to the active theme and other decorations)
+// mirrors: https://code.visualstudio.com/api/references/vscode-api#ThemableDecorationRenderOptions
+data DecoratorOptions
```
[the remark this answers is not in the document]
jurgenvinju: Isn't this called a "property" in CSS?

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -137,5 +144,111 @@ data InlayKind // this determines style
...
+    | gutterIconPath(loc iconPath)
+    | gutterIconSize(GutterIconSize size)
+;
+
+// mirrors https://code.visualstudio.com/api/references/vscode-api#MarkdownString
+data MarkdownString (
+    bool isTrusted = false,
+    bool supportHtml = false,
+    bool supportThemeIconts = false,
+    loc baseUri = |invalid:///|)
```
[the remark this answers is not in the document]
DavyLandman: is that specific for this API? or should that be an issue we should deal with in our generic mapping of locations to URIs?

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -137,5 +144,111 @@ data InlayKind // this determines style
...
+    | backgroundColor(Color col)
+    | color(Color col)
+    | overviewRulerColor(Color col) // use transparant colors
+    | opacity(str opacity)
+
+    | border(str style)
+    | borderColor(Color col)
+    // subparts of border, useful for only overriding a single aspect of a parent definition
+    // try to prefer generic border defititions
+    | borderRadius(str style)
```
jurgenvinju, line 185: what kind of styles are there and can they be enumerated? In that case use another `data` constructor

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -137,5 +144,111 @@ data InlayKind // this determines style
...
+    | borderSpacing(str style)
+    | borderStyle(str style)
+    | borderWidth(str width)
```
jurgenvinju, line 188: width is a string?

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -137,5 +144,111 @@ data InlayKind // this determines style
...
+
+    | outline(str style)
+    | outlineColor(Color col)
+    // subparts of outline, if possible, use outline.
+    | outlineStyle(str style)
+    | outlineWidth(str style)
+
+    | cursor(str cursor)
+    | fontStyle(str style)
+    | fontWeight(str style)
```
jurgenvinju, line 198: weight is a string?

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -137,5 +144,111 @@ data InlayKind // this determines style
...
+    | letterSpacing(str spacing)
```
jurgenvinju, line 199: is spacing not an "enum"?

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
DavyLandman: again, to avoid full css model, yes: https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#syntax

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
DavyLandman: it is, but it would be a wide one: https://developer.mozilla.org/en-US/docs/Web/CSS/letter-spacing#syntax

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -137,5 +144,111 @@ data InlayKind // this determines style
...
+data InstanceDecorationOptions
+    = instanceAfter(AttachableDecorationOptions tdaro)
+    | instanceBefore(AttachableDecorationOptions tdaro)
+    ;
+
+// css properties that decorators can change (in relation to the active theme and other decorations)
+// mirrors: https://code.visualstudio.com/api/references/vscode-api#ThemableDecorationRenderOptions
+data DecoratorOptions
+    // option to insert content after & before the decorated text
+    = after(AttachableDecorationOptions tdaro)
```
jurgenvinju, line 173: these smaller sets of options generate clones and overlap. Another way of solving this is adding 

`data DecorationOptions(bool attachable=false);`

and then for every constructor that is attachable:

`data DecorationOption = color(Color col, bool attachable=true)`

that way you have documented clearly which constructors are applicable but no need for a clone of the definition and the documentation for every constructor.

jurgenvinju:
Proposal: let all CSS-related information be encoded as strings, in CSS notation. So write `style("opacity: 50%")` as opposed to `opacity(50)`. This way we remain compatible with all other uses of CSS in the Rascal context, and it also does not become too VScode specific. 

Styles which are not supported will be ignored and we could refer to the VScode documentation for up-to-date information in that regard.

DavyLandman:
I like the simplicity, but that would mean we would have to parse those css string, as the vscode API does not allow for generic css. We would have to map it back to the right json elements. Quite some overhead I would also not favor.

jurgenvinju:
I see. Why did they make this so complex? Is this accidental?

DavyLandman:
I think it's by design, to limit which changes extensions can do to the IDE. since CSS is super powerful.

Also, I suspect this allows for more direct DOM manipulation, so it might be a performance benefit as well.

jurgenvinju:
Let's find an optimal balance then. 

* We could reuse `javafx.css.CssParser` to:
   * filter property types, and generate a nice error message for unsupported styles, and 
   * select the right objects to wrap each property in by their name according to the VScode API

Thereby pushing some of the complexity into our own code rather than exposing it to our users.

DavyLandman:
Thinking a bit more about it, I think the approach on only having css inside of rascal-lsp does at least make it a more generic solution.

I think it would be better to move the parsing to vscode, so we can keep the rascal-lsp bridge without too specific vscode features. It's just an extension to the lsp protocol, that the rascal-vscode extension has to map onto the vscode api.

I've looked, and a common parser for css is this library: https://www.npmjs.com/package/cssom

jurgenvinju:
That makes sense. We can keep CSS as strings in Rascal, also for other API such as Salix and Figure.

DavyLandman:
I was just thinking of a simpler scheme. We could do

`map[str,str]` so we don't even have to parse css on both sides. Just check on vscode side if the keys are valid.

Bonus: this is also how salix defines styles.

jurgenvinju:
Sure that's what I want too, but the model on the VScode side requires different object types to wrap different kinds of CSS properties. So either we regex or we parse, but I think on the VScode side we have to look into the data a bit to be able to call the right platform methods.

DavyLandman:
Closing this until at a later point we want to pick this up again.

PR #158, opened 2022-05-11
Expose all VSCode virtual file systems to Rascal

DavyLandman:
This PR connects the (optionally) virtual file systems from VSCode to the URIResolverRegistry in rascal.

We run this over a seperate json-rpc channel for two reasons:
- avoid deadlocking due to complex multiplexing issues on the same channel (vscode asking for a summary, and rascal asking back for certain files)
- we only want 1 instance, that multiple clients can connect to (for example, the REPL, rascal lsp, parametric lsp)

For the rest the most work goes in to marshaling the requests of the json-rpc layer.


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/jsonrpc/messages/ISourceLocationChangeType.java
```diff
@@ -23,15 +25,15 @@ public static ISourceLocationChangeType forValue(int value) {
         return allValues[value - 1];
     }
 
-    public static org.rascalmpl.uri.ISourceLocationWatcher.ISourceLocationChangeType translate(
```
jurgenvinju: I have the feeling that I wrote this code before. Probably a translation in the other direction ;-) But let's check if we didn't duplicate a whole hierarchy.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/jsonrpc/impl/Client.java
```diff
@@ -0,0 +1,116 @@
...
+        public Thread newThread(Runnable r) {
+            SecurityManager s = System.getSecurityManager();
+            ThreadGroup group = (s != null) ? s.getThreadGroup() : Thread.currentThread().getThreadGroup();
+            Thread t = new Thread(group, r, "FallbackResolver watcher thread-pool");
+            t.setDaemon(true);
+            return t;
+        }
+    });
+
+    private static class Watchers {
```
jurgenvinju: Please add some source code comments to explain the duplicate management of watchers. In principle the watchers are supposed to be registered on the VScode side, right? Is this copy here because we can not send the callbacks over the wire and so we look them up here?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/jsonrpc/messages/ISourceLocationChangeType.java
DavyLandman: This is by design a duplicate of the enum inside the ISourceLocationWatcher. Specifically, for enums that have to be serialized, we want to have them have a specific (predictable) numeric value, that translates to the enums on the vscode side.

Alternativly we could change the definition of the enums inside the ISourceLocationWatcher, but this way would allow for parallel evolution.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/jsonrpc/impl/Client.java
DavyLandman: Good point, I'll add some comments. And exactly, we cannot keep track of this on the VSCode side, since we use reference equality for the closures. The same closure instance is used later to "unwatch". Initially I designed something around sending over a Identity hashcode, but that was just messy. Also, this way the jsonrpc bridge is not spammed by the same event multiple times. We do the "broadcast" inside java.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -285,5 +287,10 @@ public void connect(LanguageClient client) {
             lspDocumentService.connect(this.client);
             lspWorkspaceService.connect(this.client);
         }
+
+        @Override
+        public void registerVFS(VFSRegister registration) {
+            VSCodeVFSClient.buildAndRegister(registration.getPort());
+        }
```
rodinaarssen, line 294: All changes in this file except this addition looks like autoformatting gone wrong

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/FallbackResolver.java
```diff
@@ -0,0 +1,269 @@
...
+    private final Cache<ISourceLocation, Lazy<Map<String, Boolean>>> cachedDirectoryListing
+        = Caffeine.newBuilder()
+            .expireAfterAccess(Duration.ofMinutes(2))
+            .maximumSize(10000)
+            .build();
+
+    @Override
+    public String[] list(ISourceLocation uri) throws IOException {
+        var result = call(s -> s.list(param(uri)));
+        // we store the entries in a cache, for consequent isDirectory/isFile calls
```
rodinaarssen, line 179:
```suggestion
        // we store the entries in a cache, for consecutive isDirectory/isFile calls
```

rascal-vscode-extension/.vscode/launch.json
```diff
@@ -16,9 +16,24 @@
 			"outFiles": [
 				"${workspaceFolder}/out/**/*.js"
 			],
-            "env": {
-                "RASCAL_LSP_DEV": "true"
-            },
+				"env": {
+					"RASCAL_LSP_DEV": "true"
+				},
```
rodinaarssen, line 21: formatting went haywire

DavyLandman: Ah, this was already a mess with mixed tabs and spaces, I thought I fixed it.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/jsonrpc/impl/Client.java
jurgenvinju: ok

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/jsonrpc/messages/ISourceLocationChangeType.java
DavyLandman: ah, src\main\java\org\rascalmpl\vscode\lsp\IRascalFileSystemServices.java contains the same enum. okay, I'll try to merge it.

jurgenvinju: I see; it's layering and cloning. too bad but this is Java land so let's keep the current design. please add a comment that refers with @see to the API that we are mirroring here.

jurgenvinju: ok. I'd rather move ahead now; let's add an issue in the current project to remind us of this TD.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/jsonrpc/impl/Client.java
DavyLandman: added comment to explain

jurgenvinju:
@DavyLandman zie je de reden voor die compilatie error?

DavyLandman:
Yup, rascal changed the interface for repls, so now that the new rascal version is out, I can finish this PR.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/jsonrpc/messages/ISourceLocationChangeType.java
DavyLandman: Made an issue #160 .

PR #164, opened 2022-06-10
More granular summaries

DavyLandman:
This PR is rewriting how DSLs do summaries.

Especially, it allows a DSL user to fine short and fast functions for dedicated questions, like: where is this specific thing defined.


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/model/ParametricFileFacts.java
```diff
@@ -77,34 +74,37 @@ private FileFact getFile(ISourceLocation l) {
...
+            present.invalidate();
+        }
     }
 
     private class FileFact {
         private final ISourceLocation file;
         private volatile List<Diagnostic> parseMessages = Collections.emptyList();
         private volatile List<Diagnostic> typeCheckerMessages = Collections.emptyList();
-        private final ReplaceableFuture<ParametricSummaryBridge> summary;
+        private final ParametricSummaryBridge summary;
```
jurgenvinju, line 103: mooi!

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/Locations.java
```diff
@@ -35,7 +35,7 @@
 import org.eclipse.lsp4j.TextDocumentItem;
 import org.rascalmpl.uri.URIResolverRegistry;
 import org.rascalmpl.uri.URIUtil;
-
+import org.rascalmpl.values.parsetrees.TreeAdapter;
```
jurgenvinju, line 38: unused?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/ArrayLineOffsetMap.java
```diff
@@ -34,10 +34,12 @@
 public class ArrayLineOffsetMap implements LineColumnOffsetMap {
```
jurgenvinju, line 34: changes here seem unrelated to the PR' topic? I'll try to find the commit message.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -49,6 +49,9 @@ alias Builder          = list[Message] (list[loc] /*sources*/, PathConfig /*pcfg
 alias LensDetector     = rel[loc src, Command lens] (Tree /*input*/);
 alias CommandExecutor  = void (Command /*command*/);
 alias InlayHinter      = list[InlayHint] (Tree /*input*/);
+// these single mappers get caller for every request that a user makes, they should be quick as possible
+// carefull use of memo can help with caching
+alias SingleMapper     = set[loc] (loc /*origin*/, Tree /*fullTree*/, Tree /*lexicalAtCursor*/);
```
jurgenvinju, line 54: Please change to `Definer` and `Referrer` and `Implementer`, in style with the others (`InlayHinter`). The fact that they all share the same function type is not the point of the alias, and if one of them gets an extra argument in the future, this SingleMapper does not hide anything.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -60,6 +63,14 @@ data LanguageService
     | lenses(LensDetector detector)
     | inlayHinter(InlayHinter hinter)
     | executor(CommandExecutor executor)
+    | definer(SingleMapper define)
+    | referrer(SingleMapper reference) // this
+    | implementer(SingleMapper implementations)
+    | disabledSummary( // tell rascal that the summary should not be run to calculate these
```
jurgenvinju, line 69: I'm not happy with this "disabler" semantics. If someone does not want to calculate something, then they can simply decide not to calculate something. We do not need booleans for that. A summary that does not have `definitions`, does not have definitions.

jurgenvinju: we can just have the definer, referrer and implementer, and a summary provider. All we have to do is decide what to do when we have them both. In my view the faster single functions would run first and do a fast approximation, while the summary provider would catch up sometimes and provide more accurate relations.

jurgenvinju: it makes no sense to calculate the summary references variable with a lot of detail, only to turn the information off in the IDE.

jurgenvinju (changes requested):
nice work! complex stuff with the futures as always; glad you took that one on your shoulders. I have some issues about the Rascal interface in `util::LanguageServer` that I'd like us to agree on and consolidate before we merge.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/ArrayLineOffsetMap.java
DavyLandman: I needed the inverse logic, since I have to find the cursor position in the ITree.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
DavyLandman: okay, I will.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
DavyLandman: It should maybe be called differently, but I want to allow the user to say (which we also had in eclipse ide) I'm not providing this information in the summary, so you can save requesting it, and just send an empty reply back.

DavyLandman: For example, if your DSLs can't implement references via the summary, but only on demand, you want to communicate this.

Similarly for `documentation` if your summary doesn't contain it, why wait for it to finish calculating. (or why even request it).

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/Locations.java
DavyLandman: yup, this code was moved, I'll cleanup :)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/InterpretedLanguageContributions.java
```diff
@@ -80,6 +84,29 @@ public class InterpretedLanguageContributions implements ILanguageContributions
     private final CompletableFuture<@Nullable IFunction> lenses;
     private final CompletableFuture<@Nullable IFunction> commandExecutor;
     private final CompletableFuture<@Nullable IFunction> inlayHinter;
+    private final CompletableFuture<@Nullable IFunction> definer;
+    private final CompletableFuture<@Nullable IFunction> referrer;
+    private final CompletableFuture<@Nullable IFunction> implementer;
+    private final CompletableFuture<@Nullable DisabledSummary> summaryConfig;
+
+    private static class DisabledSummary {
```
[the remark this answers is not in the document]
DavyLandman: it's just a mapping of the ADT in `disabledSummary` config inside the LanguageService.

DavyLandman:
Thanks for the review @jurgenvinju I've processed most, I think the only true-ly open one about the `disabledSummary` feature.

I'll also need to test it, hopefull tomorrow I can find some time for that.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
jurgenvinju: I want to find a different solution for this problem, if it is a problem. The current solution couples 5 different contributors, semantically, via boolean flags. That is not good. The contributors should remain independent functional callbacks and the protocol that uses them should remain "under the hood". Here things start to tangle in a way that is too ad-hoc. 

To indicate that something is not there, the design principle here is to simply not provide it. So: no callback contributed, or no relation added to the summary.

DavyLandman: this was rewritten after discussion.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/ArrayLineOffsetMap.java
DavyLandman: yup, that is also used, but we needed to first translate form LSP utf16 offsets to rascal utf32 offsets.

jurgenvinju (approved):
excellent; an important improvement for the usability of the entire design.

Issue #168, opened 2022-07-12
add function execution next to command execution contributor


jurgenvinju:
We need to be able to execute functions in the LSP server side, and receive a value back in TypeScript.

Design parallel to the CommandExecutor, but now a FunctionExecutor that returns a Rascal value which will be
encoded as JSON.

DavyLandman:
This was fixed in #171

PR #171, opened 2022-07-19
Added return values to rascal commands executed from vscode.

PieterOlivier:
Rascal commands executed from typescript over the lsp bridge could can now return values to the typescript side.
Return values are serialized to json so they are easy to work with in typescript.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
@@ -371,6 +371,10 @@ public class ParametricTextDocumentService implements IBaseTextDocumentService, 
         throw new UnsupportedOperationException("Rascal Parametric LSP has no support for this file: " + doc);
     }
 
+    private ParametricFileFacts facts(TextDocumentItem doc) {
+        return facts(doc.getUri());
+    }
+
     private TextDocumentState open(TextDocumentItem doc) {
         return files.computeIfAbsent(Locations.toLoc(doc),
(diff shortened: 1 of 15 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -357,10 +359,6 @@ private ParametricFileFacts facts(TextDocumentIdentifier doc) {
         return facts(doc.getUri());
     }
 
-    private ParametricFileFacts facts(TextDocumentItem doc) {
```
DavyLandman, line 360: while I get that this one is removed (most likely due to nobody using it). For the next time I have a preference to keep cleanup in separate PRs.

DavyLandman (commented):
I like it, I think @jurgenvinju should review it. Since @PieterOlivier and I worked together a bit on this.

But I like this solution for #168 since this was just missing about the command pattern.

Also: it's not truly optional, as there is no "null" for `value` in rascal.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
PieterOlivier: I have reverted all changes not related to core purpose of this PR.

jurgenvinju (approved):
Perfect extension :-) Thanks! We have to document somewhere (let's wait until rascal2doc is released with the rascal-maven-plugin, that normally _nothing_ is done with the return value.

PR #176, opened 2022-08-31
Moved to lsp 3.17 and switch away our own inlayhint implementation

### LINKED ISSUE #148: Migrate to vscode native inlayhint provider feature
[Version v1.65 exposed the InlayHint feature](https://code.visualstudio.com/updates/v1_65#_inlay-hints) for all extensions, so we should switch to that.

### LINKED ISSUE #162: Migrate to lsp4j 0.13.0 & vscode-languageclient 8.0.0
We should update to latest version of LSP (3.17.0) by updating both lsp4j & vscode-langaugeclient.

This also enables #148 since it was added to the latest LSP version. So no more manual work.

DavyLandman:
In this PR two things happened:

- bumped to LSP 3.17 (and corresponding lsp4j & vscode-languageclient node)
- migrate our custom lsp extension for inlayHint to the LSP based on.

Note that in the future we most likely want to adopt more [features from LSP 3.17](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#version_3_17_0) like for example markdown in hints & labels. But for now, this is the minimal and removes decreases our own codebase.

This fixes #162 and fixes #148 .


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -313,6 +313,17 @@ public BaseWorkspaceService getWorkspaceService() {
             return lspWorkspaceService;
         }
 
+        @Override
+        public NotebookDocumentService getNotebookDocumentService() {
+            return null; // can be removed after lsp4j v0.16.0 is released (https://github.com/eclipse/lsp4j/issues/658)
+        }
+
+        @Override
+        public void setTrace(SetTraceParams params) {
```
jurgenvinju, line 322: what are we ignoring here? perhaps add a little comment for future reference

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -294,17 +293,25 @@ private static <T> CompletableFuture<T> recoverExceptions(CompletableFuture<T> f
...
+        // unpack rascal value
+        var t = (IConstructor) v;
+        var loc =(ISourceLocation) t.get("position");
+        var label = ((IString) t.get("label")).getValue();
+        var kind = (IConstructor) t.get("kind");
+        var toolTip = (IString)t.asWithKeywordParameters().getParameter("toolTip");
+
+
+        // translate to lsp
+        var spaceBefore = label.startsWith(" ");
```
jurgenvinju, line 305: let's inline these vars. single use expressions require less mental space.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -139,16 +139,15 @@ data Command(str title="")
 
 data InlayHint
     = hint(
-        loc range, // should point to a cursor position (begin and end column same)
+        loc position, // should point to a cursor position (begin and end column same)
```
jurgenvinju, line 142: this constraint affects all DSL implementations, such that they have to make their start positions equal to their end. I propose we use the start position of _any_ location instead and solve this constraint under-the-hood.

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
```diff
@@ -71,10 +71,12 @@ data Command
 rel[loc,Command] picoLenses(start[Program] input) = {<input@\loc, renameAtoB(input, title="Rename variables a to b.")>};
 
 
+loc atEnd(loc l) = l[begin = l.end][offset=l.offset+l.length - 1][length = 1];
+
 list[InlayHint] picoHinter(start[Program] input) {
     typeLookup = ( "<name>" : "<tp>" | /(IdType)`<Id name> : <Type tp>` := input);
     return [
-        hint(name.src, ": <typeLookup["<name>"]>", \type()) | /(Expression)`<Id name>` := input
+        hint(atEnd(name.src), " : <typeLookup["<name>"]>", \type()) | /(Expression)`<Id name>` := input, "<name>" in typeLookup
```
jurgenvinju, line 79: see below. I'd like to see `name.src` without the `atEnd` wrapper.

rascal-vscode-extension/src/extension.ts
```diff
@@ -242,7 +236,6 @@ async function startTerminal(uri: vscode.Uri | undefined, ...extraArgs: string[]
         uri = undefined;
     }
     const rascal = await rascalClient;
-    await rascal.onReady();
```
jurgenvinju, line 245: wasn't this essential? no source code comment to claim so, but why is it gone and what can happen if we remove this?

jurgenvinju (commented):
nice code. two minor things: remove need to `atEnd` wrappers and I am slightly worried about removing the call to `onReady()`.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
DavyLandman: yeah, I should word this differently. The comment was more right than the name of the variable. I'll update the comment a bit. but true, what we do is we put it at the start of the location.

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
DavyLandman: Okay, so for now the design is close to LSP's inlayhint feature. Where the start of the source location determines where the hint it put. We could add a feature where you can pass in a location, and a special bool to say: take the end of this location, not the beginning of this location to place this hint?

rascal-vscode-extension/src/extension.ts
DavyLandman: It's discussed here: https://github.com/Microsoft/vscode-languageserver-node#3170-protocol-800-json-rpc-800-client-and-800-server

Roughly, the `onReady` stuff was bad design, you can no just register/send stuff before the client is connected, and the library will take care of the rest. This simplified some of our init code.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
DavyLandman: It's a new feature, we aren't using it yet, I'll add a comment (and issue) to say we should do this in the future.

DavyLandman: Added #177 for this.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
DavyLandman: I've changed it, this is better to use. thanks for helping not just follow LSP spec ;)

jurgenvinju: that's what I'm here for ;-)

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
jurgenvinju: ok

rascal-vscode-extension/src/extension.ts
jurgenvinju: It may also be the trigger for the infinite recursion you were talking about earlier this week?

rascal-lsp/pom.xml
```diff
@@ -187,7 +187,7 @@
         <artifactId>maven-dependency-plugin</artifactId>
         <version>3.1.2</version>
         <executions>
-          <execution>
+          <execution> <?m2e ignore?>
```
jurgenvinju, line 190: can I ignore this, or is this an accident?

jurgenvinju (approved):
Thanks for the new version. There was only a small question about the pom file, but I think we should merge this contribution.

rascal-vscode-extension/src/extension.ts
DavyLandman: nope, that's about setting the language of a text editor. this code is only run for the very first time when we start-up the lsp server.

Issue #185, opened 2022-09-27
showMessage and logMessage from util::IDEServices have no effect anymore


jurgenvinju:
**Describe the bug**

In util::IDEServices two very similar functions have lost their effect:
* showMessage would show a "popup" in the bottom right corner for alerting the user to certain events
* logMessage would add a log entry to the Rascal MPL LSP log, or the parametrized LSP log, respectively

Both have worked before. So it could be our upgrade to a higher version of VSCode or the LSP that has introduced this
issue. We can also add (manual) tests for these two function to the release issue template while we are at it.

DavyLandman:
Can you point to the commit that added that functionality? Since I cannot find it in our history.

I agree we need more tests.

jurgenvinju:
I'll have to dig too. That was around the crunch time for the Strumenta demo with the graph language I believe.

DavyLandman:
I've only been able to find places where we would report unsupported features using the `languageClient.showMessage` call.

jurgenvinju:
I'll look it up tomorrow with `git reflog -S showMessage`

jurgenvinju:
I only see the addition of the (default) interface methods but not their implementation in the log. This means that this functionality is still on a branch somewhere, or has never been committed.

jurgenvinju:
Ok, after some deep archeology I found out that these things have worked (a bit), but I removed the code before pushing because while testing several times a deadlock occurred. Next step is to enable only the `showMessage` functionality withou the `logMessage` because it seems the `logMessage` is what is causing the deadlocks.

DavyLandman:
Do you know why showMessage causes no deadlocks?

jurgenvinju:
it is not called during the start-up processes

DavyLandman:
could we add a guard around the `logMessage` & `showMessage` implementation that we only call it once we know that we've finished initializing?

Else someone adds a call to `showMessage` in the future, and introduces a deadlock accidentally.

jurgenvinju:
I want to understand completely what is going on first. And then this could be a solution, with a lot of source code comments :-)

PR #186, opened 2022-10-03
added basic scaffolding that should implement log and show message but it does not work yet.

jurgenvinju:


jurgenvinju:
This should address #185 ; added the initial implementation but I'm debugging now because now the terminal will not start. It's possible I create some kind of infinite recursion.

jurgenvinju:
Or it could be a deadlock.. if `logMessage` was used before it would go to stderr, and now it needs a server to send the message to then the server is still booting.

PR #188, opened 2022-10-04
No longer adding an explicit dependency to rascal-core and typepal for DSL evaluators

rodinaarssen:


DavyLandman (approved):
Nice clean and simple.

jurgenvinju:
This makes total sense.

Issue #189, opened 2022-10-10
Release rascal-lsp & vscode-lsp wrapper as a seperate dependency on npm (for deploying rascal DSLs without dependencies on rascal-vscode extension)


DavyLandman:
**Is your feature request related to a problem? Please describe.**
Currently, if we deploy a VS Code DSL that uses Rascal, we have to make a extension that depends on the rascal vscode extension. But you'll always get updated to the latest version. So 2 years in the future, that DSL extension might break due to a breaking change that happened in the rascal-vscode extension.

**Describe the solution you'd like**
I think properly restricting the project to full backwards compatiblity is not feasible or even wise at this moment. So I want to propose we bundle up the language parametric "side" of rascal-lsp & rascal-vscode-extension in a npm package. This npm package contains everything you need to run a Rascal DSL from outside a rascal repl. So we can deploy a new extension, that has a fixed dependency on this npm package, it starts it's own LSP process and is not depending on any specifics of the rascal vscode extension that is or is not installed.

This would even allow multiple DSL installed, with different versions of rascal & rascal-lsp.

The only problem I forsee is with the VFS, the first rascal jar that is started would win that race. So we have to be very careful to remain backwards compatible there, or think of some complicated redesign of that registry component.

**Describe alternatives you've considered**
Extend the rascal-lsp interface with a way to allow extensions to request older version of rascal & rascal-lsp. This seems like a fragile approach, where we would lock down this api for future changes, and would dynamically be downloading jars & javascript files and loading them, not sure if that would even work.

**Additional context**
This continues the idea presented in #94

jurgenvinju:
* the VFS URI schemes are supposed to implement unique identification mechanisms, and so it should not matter who resolves them. If we run into the problem that it matters which resolver does the work, then we have to re-think that scheme rather than this solution. It could be however that we need to _multiplex_ if some resolvers offer other schemes...

jurgenvinju:
* Deployment on npm. Is that necessary/wanted? What are the pros and cons, and alternatives?

DavyLandman:
> * the VFS URI schemes are supposed to implement unique identification mechanisms, and so it should not matter who resolves them. If we run into the problem that it matters which resolver does the work, then we have to re-think that scheme rather than this solution. It could be however that we need to _multiplex_ if some resolvers offer other schemes...

What if I have a DSL with rascal version X, and it's started first, so we get the VFS from version X.

[...]

PR #198, opened 2022-11-09
Refactored out different components of the rascal extension

DavyLandman:
To prepare for #189 I've refactored out the big tangled `extension.ts` into separate classes.

Now for #189 we only have to figure out how we can export the `TermLanguageServer` into a dedicated npm package.

Feature wise, I tried to not change stuff. The only functionality that is lost is that the starting of the rascal server is now not dynamic anymore. If you start the extension, rascal is started. This makes sense again if we don't have to have a dependency on rascal extension in the future.

I've run the test we normally do for a release, and it all seems to work. But please, if you can, read with me.


rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -0,0 +1,187 @@
...
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+// The module 'vscode' contains the VS Code extensibility API
+// Import the module and reference it with the alias vscode in your code below
```
jurgenvinju, line 28: is een beetje een gek commentaar. kan misschien weg?

rascal-vscode-extension/src/RascalLSPConnection.ts
```diff
@@ -0,0 +1,213 @@
...
+ */
+import * as vscode from 'vscode';
+import * as net from 'net';
+import * as os from 'os';
+import * as path from 'path';
+
+import { integer, LanguageClient, LanguageClientOptions, ServerOptions, StreamInfo } from 'vscode-languageclient/node';
+import { getJavaExecutable } from './auto-jvm/JavaLookup';
+import { RascalFileSystemProvider } from './RascalFileSystemProviders';
+import { VSCodeUriResolverServer } from './VSCodeURIResolver';
```
jurgenvinju, line 35: nice overview of internal dependencies. 👍

rascal-vscode-extension/src/RascalLSPConnection.ts
```diff
@@ -0,0 +1,213 @@
...
+
+
+
+// deploymode = (deployMode || testDeployMode)
+// jarPath = rascalExtensionContext!.asAbsolutePath(path.join('.', 'assets', 'jars', j)))
+// push return onto subscription stack!
```
jurgenvinju, line 41: these three comment lines don't make sense to me (yet). what's up?

rascal-vscode-extension/src/RascalLSPConnection.ts
```diff
@@ -0,0 +1,213 @@
...
+        ? await buildRascalServerOptions(jarPath, isParametricServer)
+        : () => connectToRascalLanguageServerSocket(devPort) // we assume a server is running in debug mode
+            .then((socket) => <StreamInfo> { writer: socket, reader: socket});
+
+    const clientOptions = <LanguageClientOptions>{
+        documentSelector: [{ scheme: '*', language: language }],
+    };
+
+    const client = new LanguageClient(language, title, serverOptions, clientOptions, !deployMode);
+    //rascalExtensionContext!.subscriptions.push(client);
```
jurgenvinju, line 56: commented code

rascal-vscode-extension/src/RascalLSPConnection.ts
```diff
@@ -0,0 +1,213 @@
...
+
+    await client.start();
```
jurgenvinju, line 58: does this work as expected? I get warnings from the typescript checker.

rascal-vscode-extension/src/RascalLanguageServer.ts
```diff
@@ -0,0 +1,69 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+import * as path from 'path';
+import * as vscode from 'vscode';
+
+import { LanguageClient } from 'vscode-languageclient/node';
+import { activateLanguageClient } from './RascalLSPConnection';
+import { LanguageParameter, TermLanguageServer } from './TermLanguageServer';
```
jurgenvinju, line 32: we should pick a term for "TermLanguageServer" or "Language Parameterized Server" and use it everywhere. Now we have two names for the same thing and that becomes confusing. TermLanguageServer is nice and short, but you'd have to come from the ASF+SDF times to understand why it is called like that. Or maybe a third name would be better, like "MultiplexedLanguageServer" or something. The README uses "Language-Parametrized"

rascal-vscode-extension/src/TermLanguageServer.ts
```diff
@@ -0,0 +1,153 @@
...
+    name: string; 			// name of the language
+    extension:string; 		// extension for files in this language
+    mainModule: string; 	// main module to locate mainFunction in
+    mainFunction: string; 	// main function which contributes the language implementation
+}
+
+function languageKey(lang: LanguageParameter) {
+    return `${lang.mainModule}::${lang.mainFunction}`;
+}
+
```
jurgenvinju, line 153: sweet!

rascal-vscode-extension/src/extension.ts
```diff
@@ -26,463 +26,21 @@
...
-        if (!text.document.uri || !moduleName) {
-            return;
-        }
-        startTerminal(text.document.uri, "--loadModule", moduleName);
-    });
-    rascalExtensionContext!.subscriptions.push(command);
+    const jars = context.asAbsolutePath(path.join('.', 'assets', 'jars'));
+    const extension = new RascalExtension(context, jars, (deployMode || testDeployMode));
+    context.subscriptions.push(extension);
+    return extension.backwardsCompatibleExternalConsumers();
```
jurgenvinju, line 42: I don't understand the meaning of the identifier `backwardsCompatibleExternalConsumers`; "consumers" does not speak to me, and "backward compatible" also not. compatible with what? and why? and is this going to disappear soon then? Perhaps rename or add comments with the declaration of the method to explain what is going on?

jurgenvinju (commented):
excellent move forward for future stability and locality. I have some minor questions and I think we should pick a name for the "TermLanguageServer" that we can use both in internal identifiers and in documentation to refer to the language parametrized or "generic" or "multiplexed" language serverm, or "bridged". I'm open to suggestions :-) not "Term" that smells of algebraic specifications and term rewriting that we don't have anymore.

rascal-vscode-extension/src/RascalExtension.ts
DavyLandman: aah, true, die had ik meeste plekken al weggehaald.

rascal-vscode-extension/src/RascalLSPConnection.ts
DavyLandman: refactoring stuff, during rewriting i kept these lines, so that the client code should use this logic. so I gotta cleanup there.

rascal-vscode-extension/src/extension.ts
DavyLandman: currently, we enabled other extensions to call "registerLanguage" from the outside. This refactoring prepares for a different way to deploy your own DSL (namely the design discussed in #189) but to remain backwards compatible with existing extension, we still offer the `register` & `unregister` language as exported functionality.

But we should rename this maybe?

rascal-vscode-extension/src/RascalLSPConnection.ts
DavyLandman: check, will get deleted, now the whole Connection is disposible.

rascal-vscode-extension/src/extension.ts
jurgenvinju: I would expect that we always have `register` and `unregister` during development time, just that we have different ways of deploying DSLs when we are ready. The way in #189 is most attractive from a branding point of view (we should think about branding Rascal there somehow while still allowing the brand of the DSL to float on top).

DavyLandman: During develoment time you don't need this, if #189 is finished. Either you are developing your DSL logic inside of 1st level rascal, then you start everything using the REPL.

If you are in second level of your own extension (with a DSL) this setup is really cumbersome (as you need to enable all other extensions to load, just so that your own extension can find rascal). So #189 will also make second-level development of your extension easier.


But this is not the discussion here. The pre-#189 design is that we export 3 functions, such that external extension can call `register` & `unregister` for any language. This function keeps that working, untill we at a later point in time can truly deprecated it, and even later, remove it.

rascal-vscode-extension/src/RascalLanguageServer.ts
DavyLandman: I agree, the term "term" has run it's course. `ParameterizedLanguageServer` would match better with the java side of the extension. But I also like `MultiplexedLanguageServer`, since that is more a description of what it does.

rascal-vscode-extension/src/extension.ts
jurgenvinju: ok

rascal-vscode-extension/src/RascalLanguageServer.ts
jurgenvinju: and we'd rename the Java classes too, right? in one go.

DavyLandman: Optimizing for less invasive file movements, we could use `ParameterizedLanguageServer`? 

Also, I think Generic/Generalized is to "generic" (no pun intended).

jurgenvinju: I'm ok with that. It says what it does.

rascal-vscode-extension/src/RascalLSPConnection.ts
DavyLandman: I get no warnings in my ide? (and yes, it works, if this one failed, nothing would work).

jurgenvinju: I will update my json modules and see if it goes away...

DavyLandman:
Thanks for the review, I've also moved some files around into subpackages, it became a bit crowded.

PR #199, opened 2022-11-15
Moved to webpack for faster compilation and less files in the vsix

DavyLandman:
I think we've all seen this message:

```
This extension consists of 331 files, out of which 221 are JavaScript files.
For performance reasons, you should bundle your extension: https://aka.ms/vscode-bundle-extension . 
You should also exclude unnecessary files by adding them to your .vscodeignore: https://aka.ms/vscode-vscodeignore
```

We started out with webpack (aka, bundling all files into a single javascript file) but there were some issues with the integration of webpack & vscode. Those are solved now (if you install the `amodio.tsl-problem-matcher` extension that the project now suggests).

So this PR is purely making our packaging more compact (less javascript code). Sadly the jar size is still the thing that gets us ;)

Before: 

```
 DONE  Packaged: ...\rascalmpl-0.5.5-head.vsix (331 files, 128.75MB)
```

Now:
```
 DONE  Packaged: ...\rascalmpl-0.5.5-head.vsix (18 files, 128.22MB)
```

### DIFF
--- a/rascal-vscode-extension/.vscode/settings.json
+++ b/rascal-vscode-extension/.vscode/settings.json
@@ -1,10 +1,12 @@
 // Place your settings in this file to overwrite default and user settings.
 {
     "files.exclude": {
-        "out": false // set this to true to hide the "out" folder with the compiled JS files
+        "out": false, // set this to true to hide the "out" folder with the compiled JS files
+        "dist": false, // set this to true to hide the "out" folder with the compiled JS files
     },
     "search.exclude": {
-        "out": true // set this to false to include "out" folder in search results
+        "out": true, // set this to false to include "out" folder in search results
+        "dist": true, // set this to false to include "out" folder in search results
     },
     // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
     "typescript.tsc.autoDetect": "off",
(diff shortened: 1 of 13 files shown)


rascal-vscode-extension/.vscode/settings.json
```diff
@@ -1,11 +1,13 @@
...
     "search.exclude": {
-        "out": true // set this to false to include "out" folder in search results
+        "out": true, // set this to false to include "out" folder in search results
+        "dist": true, // set this to false to include "out" folder in search results
     },
     // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
     "typescript.tsc.autoDetect": "off",
-}
\ No newline at end of file
+}
```
jurgenvinju, line 13: I find these comments superfluous. the key is called "search.exclude" or "files.exclude" so that says enough.

jurgenvinju (approved):
I'm not the JS or TS expert, but this seems good. Please test the developer debug setup before merging as well. Remove superfluous comments taken from documentation examples?

rascal-vscode-extension/.vscode/settings.json
DavyLandman: true, they came from the tool that generated them. I'm gonna cleanup some.

PR #206, opened 2022-11-30
Allow vscode extensions to use rascal-vscode independently of installed rascal.

DavyLandman:
This work finished the design in #189 and will automatically package the ParameterizedLanguageServer on npm, as a library you can use in your own VS Code extension.


examples/pico-dsl-extension/assets/pico-language-configuration.json
```diff
@@ -0,0 +1,18 @@
+{
+  "comments": {
+    "lineComment": ";",
+    "blockComment": []
+  },
+  "autoClosingPairs": [
+    { "open": "{", "close": "}" },
```
jurgenvinju, line 7: these brackets do not occur in Pico

examples/pico-dsl-extension/src/extension.ts
```diff
@@ -0,0 +1,37 @@
...
+		calcJarPath(context), 
+		true, 
+		"pico", // vscode language ID
+		"Pico", // vscode language Title (visible in the right bottom corner)
+		language);
+	// adding it to the subscriptions makes sure everything is closed down properly
+	context.subscriptions.push(lsp);
+}
+
+function calcJarPath(context: vscode.ExtensionContext) {
```
jurgenvinju, line 30: shouldn't this be folded into the constructor of ParametrizedLanguageServer?

examples/pico-dsl-lsp/src/main/rascal/lang/pico/LanguageServer.rsc
```diff
@@ -0,0 +1,55 @@
...
+    return summary(l,
+        messages = {<src, error("<id> is not defined", src)> | <src, id> <- uses, id notin defs<0>},
+        references = (uses o defs)<1,0>,
+        definitions = uses o defs,
+        documentation = (uses o defs) o docs
+    );
+}
+
+// run this from a REPL while developing the DSL
+int main(int testArgument=0) {
```
jurgenvinju, line 42: remove testArgument

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -161,7 +173,11 @@ public void initializeServerCapabilities(ServerCapabilities result) {
         result.setImplementationProvider(true);
         result.setSemanticTokensProvider(tokenizer.options());
         result.setCodeLensProvider(new CodeLensOptions(false));
-        result.setExecuteCommandProvider(new ExecuteCommandOptions(Collections.singletonList(BaseWorkspaceService.RASCAL_META_COMMAND)));
+        String commandName = BaseWorkspaceService.RASCAL_META_COMMAND;
+        if (!dedicatedLanguageName.isEmpty()) {
+            commandName += "-" + dedicatedLanguageName;
+        }
+        result.setExecuteCommandProvider(new ExecuteCommandOptions(Collections.singletonList(commandName)));
```
jurgenvinju, line 180: I'm missing why we check for prefix when the commandname comes back in a message from the LSP. Here we register the commandName as-is; not its prefix? pls explain.

rascal-vscode-extension/README-lib.md
```diff
@@ -0,0 +1,9 @@
+# Rascal Metaprogramming Language - VSCode DSL Runtime library
+
+This package is intended writing your own VS Code extension while using
+[Rascal's VS Code LSP abstraction](https://github.com/usethesource/rascal-language-servers).
+
+This package extracts the [`ParameterizedLanguageServer`](https://github.com/usethesource/rascal-language-servers/blob/main/rascal-vscode-extension/src/lsp/ParameterizedLanguageServer.ts) class from the [Rascal VS Code extension](https://marketplace.visualstudio.com/items?itemName=usethesource.rascalmpl).
```
jurgenvinju, line 6: don't understand the language. "This package extracts ... " What is it exactly that is encapsulated? The Readme also defines two LSP servers earlier and gives them names. Perhaps that would help if we use the same terms here for the same concepts?

rascal-vscode-extension/src/rewritePackageForNPMPublish.js
```diff
@@ -0,0 +1,54 @@
...
+  "compile" : packageContents.scripts["compile-lib"],
+  "prepare": "npm run compile",
+};
+
+// we add wich files we actually want to add
+packageContents["main"] = "lib/lsp/library.js";
+packageContents["types"] = "lib/lsp/library.d.ts";
+packageContents["files"] = ["lib/**/*", "assets/**/*"];
+
+fs.writeFileSync("package.json", JSON.stringify(packageContents, null, 2));
```
jurgenvinju, line 54: I have truly no clue what this file does and why it exists. pls document with two lines of comments.

jurgenvinju (changes requested):
This is a great new way of using the parametric LSP, without doing any damage to the other way of using the parametric LSP. So with fixing the minor comments I'm all for merging this. I am wondering if we need to adapt the release steps in the issue template? If not please say so, otherwise pls adapt the release template. Thanks for this!

examples/pico-dsl-extension/assets/pico-language-configuration.json
DavyLandman: ah, I copied them from a previous project :)

examples/pico-dsl-extension/src/extension.ts
DavyLandman: no, since it's up to the user how they package their stuff. If you are using typescript/webpack/yarn/browserify, you name it. This is the only end user concern that I couldn't strip away.

It's all due to the fact that npm&webpack is only primed for javascript files. Other assets (like css) are sometimes bundled inside of a javscript file, just to get rid of this. But making a 100MB json/javascript file to encapsulate a jar seemed a bit funky to me.

jurgenvinju: that makes total sense

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
DavyLandman: This is for the `rascal-meta` command feature, where a vs code extension can call a rascal registered command from "the outside". Our current implementation registers `rascal-meta-command`, so we can't have collisions there with "dedicated"  versions of this server. This was the only point where I've had to special case the code for these dedicated instances.

The non-dedicated behaves as before, it's just that in case you are in a dedicated setup, you have to invoke the command with the right suffix (namely, the language identifier, so for example `rascal-meta-command-pico`).

rascal-vscode-extension/src/rewritePackageForNPMPublish.js
DavyLandman: This file rewrites the `package.json` so that it's suited for npm, instead of suited for VS Code marketplace.

I tried to pack it all in the file name, but I'll add some comments.

rascal-vscode-extension/README-lib.md
DavyLandman: I've improved the wording in this file.

rascal-vscode-extension/src/rewritePackageForNPMPublish.js
DavyLandman: I've added comments in this file.

DavyLandman:
Thanks for the review. The are not changes to the release process, the action is already adopted, every release of rascal-vscode-extension immediately is published to the npm store.

jurgenvinju (approved):
thanks

Issue #180, opened 2022-09-12
Add parser and static checker for RASCAL.MF


jurgenvinju:
Proposal: to add `META-INF/RASCAL.MF` to the Rascal LSP implementation with some meaningful checks including the the syntax of the last line problem.

For example, we could also check if people have rewritten `Require-Libraries` instead of `Required-Libraries` or whatever the right incantation is :-) Some auto-completion would also be great.

_Originally posted by @jurgenvinju in https://github.com/usethesource/rascal/issues/1646#issuecomment-1238280719_

PR #215, opened 2022-12-13
Added warnings&quickfixes for RASCAL.MF errors

### LINKED ISSUE #117: Display proper error message if the name of the project in RASCAL.MF does not align with the workspace project folder name
I edited this issue report. Error messages downstream are triggered if `Project-Name` in `RASCAL.MF` is not the same as the workspace folder name. 

The type-checker, for example, can not find .tpl files for its own project that way.

DavyLandman:
Initial implementation of #180 and fixes #117

Note that #180 wanted a full language extension, that would be nice, but until then, let's protect everyone from these common errors.

### DIFF
--- a/rascal-vscode-extension/src/extension.ts
+++ b/rascal-vscode-extension/src/extension.ts
@@ -28,6 +28,7 @@ import * as path from 'path';
 import * as vscode from 'vscode';
 
 import { RascalExtension } from './RascalExtension';
+import { RascalMFValidator } from './ux/RascalMFValidator';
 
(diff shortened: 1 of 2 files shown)


jurgenvinju (approved):
As long as we understand that it is our goal to write such checks in Rascal and not in TS; I'm very happy with these features. Will save us all a lot of time!

DavyLandman:
Agreed, but it got very cyclic (and a lot more work to add). But for #180 a pure rascal one would be better.

PR #218, opened 2022-12-21
Implemented rascal library navigator

### LINKED ISSUE #93: Being able to browse the libraries a project depends on (like `std:///`)
Rascal-eclipse offers the feature of browsing a libraries we depend on in the tree view to the left.
It is something @tvdstorm and I use a lot, and we would like something like that in VScode. 

This includes
the "Show in" feature of the editor, where you could jump from the library file in an editor (for example you'd end up there while debugging or following a declaration reference), to the library file in the tree view.

It's worth investigating what the idiomatic manner for exploring libraries is in VScode/LSP before we start
replicating exactly how Eclipse did it. 

It is also worth investigating how to visualize/make browsable the _full information_ of a PathConfig instance for a project. This includes not only the roots of all libraries, but also other configuration options which are good to make transparant such as the `bin` folder and JVM `classpath`. As often things go wrong around classpaths and libraries, it is very good to be able to see what is going on after the paths have been resolved.

DavyLandman:
This PR implements the rascal library navigator we all love from eclipse. (fixes #93).

image

I think all is working as expected, but I'm not sure yet what we should do with lib dependencies on projects that are in your workspace. Currently it show them, and links to their target folder, but I'm not sure if that is wise. Perhaps we should try and remove them?

I think the logic in rascal-lsp, that translates a `lib://` path to the jars on the classpath, is a bit brittle, and maybe we could add some feature to prelude or somewhere else that does this for us. Although I'm breaking through an abstraction here we rather not expose.


rascal-vscode-extension/src/fs/RascalFileSystemProviders.ts
```diff
@@ -116,3 +117,8 @@ interface WatchParameters {
     recursive: boolean;
     excludes:Array<string>;
 }
+
+interface FileWithType {
```
DavyLandman, line 121: Note, this was a bug, `readDirectory` signature on java side was different.

jurgenvinju:
We had a call about this great idea. I think it will help users understand better what is going on with the search path for modules of the interpreter. The next step is to get the right data out of the PathConfig class and to remove possible non-existing paths from there. 

The other problem with the VFS for `lib` being different between different projects is something to ponder for a bit. The current solution is creative and it works, but it also is brittle against changes in the LibraryResolver code (which is brittle in the first place). This also relates to issues in the lib scheme regarding missing version identifiers. If we'd solve those first, then the lib scheme could become a universal identification scheme for jars under the assumption of the maven repository.

jurgenvinju:
👍 I like it, but I'm biased. Let's ask some other users about the pathConfig view in its entirety? Like @PaulKlint or @PieterOlivier or @rodinaarssen ?

DavyLandman:
A bit more tuning:

- detect current project paths and rewrite them to show `project://`
- make sure we can go into jar's by automatically rewriting them to `jar+file` scheme when collapsing.

image

jurgenvinju:
Check! Are we 100% sure that the rewriting actually represents something that works in the terminal exactly as expected?

jurgenvinju:
`system:/` has only one slash and that does not work in Rascal. Can we print them all?

DavyLandman:
> Check! Are we 100% sure that the rewriting actually represents something that works in the terminal exactly as expected?

well, if you see an absolute path reference, and that points to a path in the workspace of VS Code, I think it's safe to rewrite that to a slightly more readable version.

> Aren't jarfiles already jarified in the srcs path? I'm confused.

Yup, those are, as you can see in the print. it's just that the jars on the compilation path (and others) are just plain jars, so if you try to open them up, I ask rascal (via the VFS) to browse inside. 


> system:/ has only one slash and that does not work in Rascal. Can we print them all?

This is an annoying aspect of VS Code, their handling of URIs are slightly different, I would have to see what the original uri is, and how it went wrong, but I'm not changing the rendering in any way myself.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -278,53 +280,14 @@ private static PathConfig findPathConfig(ISourceLocation path) throws IOExceptio
...
-                            l = RascalManifest.jarify(l);
-                            var libName = new RascalManifest().getProjectName(l);
-                            if (!seen.contains(libName)) {
-                                result.add(new ProjectLibFolder(libName, l));
-                                seen.add(libName);
-                            }
-                        }
-                    }
-                    return result.toArray(new ProjectLibFolder[0]);
+                    return PathConfig.fromSourceProjectMemberRascalManifest(projectFolder.getLocation(), RascalConfigMode.INTERPETER).asConstructor();
```
jurgenvinju, line 290: Could we return the map of keyword parameters here as a generic object? That way the code is 100% robust against changes in PathConfig.

rascal-vscode-extension/src/ux/LibraryNavigator.ts
```diff
@@ -64,9 +64,13 @@ abstract class RascalLibNode extends vscode.TreeItem {
 
 }
 
-interface ProjectLibFolder {
-    libName: string;
-    uri: string;
+interface PathConfig {
```
jurgenvinju, line 67: could avoid the co-evolution problem here. See above.

rascal-vscode-extension/src/ux/LibraryNavigator.ts
```diff
@@ -133,7 +161,7 @@ class RascalFSDirEntry extends RascalLibNode {
 
 class RascalFSFileEntry extends RascalLibNode {
     constructor(readonly fileName: string, readonly parentDir: vscode.Uri, parent: RascalLibNode) {
-        super(parentDir.with({path: posix.join(parentDir.path, fileName)}), vscode.TreeItemCollapsibleState.None, parent);
+        super(parentDir.with({path: posix.join(makeAbsolute(parentDir.path), fileName)}), vscode.TreeItemCollapsibleState.None, parent);
         this.command = <vscode.Command>{command:"vscode.open", title: "Open Rascal file", arguments: [this.resourceUri]};
```
jurgenvinju, line 179: other files in the hierarchy might still be interesting for opening.

DavyLandman: I think this is a workaround for a bug of the uri parser in vscode, `tmp:///` as an empty string as path field. while it needs to be `/`.

rascal-vscode-extension/src/ux/LibraryNavigator.ts
```diff
@@ -133,7 +161,7 @@ class RascalFSDirEntry extends RascalLibNode {
 
 class RascalFSFileEntry extends RascalLibNode {
```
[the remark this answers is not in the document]
DavyLandman:
```
            else if (f.endsWith(".rsc")) {
                result.push(new RascalFSFileEntry(f, loc, parent));
            }
```

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
DavyLandman: ok, changed that.

rascal-vscode-extension/src/ux/LibraryNavigator.ts
```diff
@@ -82,17 +82,13 @@ class RascalProjectRoot extends RascalLibNode {
...
-        function buildPathNode(name: "srcs" | "libs" | "ignores" | "javaCompilerPath" | "classloaders", root: RascalLibNode) {
-            return new RascalPathNode(name, paths[name], root);
-        }
         return [
-            buildPathNode("srcs", this),
-            buildPathNode("libs", this),
-            buildPathNode("ignores", this),
-            buildPathNode("javaCompilerPath", this),
-            buildPathNode("classloaders", this)
+            new RascalPathNode("Sources", paths.srcs, this),
```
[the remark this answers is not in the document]
DavyLandman: I've now moved the labels to the rascal-lsp side.

DavyLandman:
@jurgenvinju I think this is in a place you would like. there is no more magic mapping now (like making a different label for the project entries). 

Although I think you would say that the `PathConfig` is not the one the REPL completly uses. So maybe you could help provide the right information that should instead be shown in this list. Right now it's all happening in `BaseLanguageServer::supplyPathConfig` so whatever you think should be shown in VS Code, could you calculate it there and return it?

DavyLandman:
image

Current state. We would need to do a bit more testing, and figuring out what we should be showing, maybe not the full file name but just the .jar name (and in a hover the full file name or something)

DavyLandman:
@jurgenvinju whats your opinion in this PR? I think this is ready or close to ready for releasing to end users.

jurgenvinju:
I think this is really great. If we could show only the filename and extension of a URI that might help, and the full URI on hover, but this is not essential. Let's do that later in an improvement PR.

jurgenvinju:
Let's change the main label to "Rascal Configuration" from "Rascal Libraries". It's more accurate, and the labels of the fields of pathConfig tell the story of what each collection of URIs is (srcs, libs, etc.)

DavyLandman:
> I think this is really great. If we could show only the filename and extension of a URI that might help, and the full URI on hover, but this is not essential. Let's do that later in an improvement PR.

Currently this is already separated. `RascalLibraryRoot` class gets a label name, and a actual URI. Currently the label name is the full URI, but we could do some special handling there, but what if its: `file:///.../rascal-project/target/classes` we don't want to show `/target/classes` in that case right?

DavyLandman:
image

I've updated the rendering (and naming) based on your suggestion.

Maybe instead of `project://` we should also just do `file://../` for those, as path config doesn't actually contain project uris.

jurgenvinju:
Path configs may contain any URI scheme, including project:// and target://, so we should make sure that can be handled. Such URI can come from manual pom.xml entries or rascal.mf details.

DavyLandman:
> Path configs may contain any URI scheme, including project:// and target://, so we should make sure that can be handled. Such URI can come from manual pom.xml entries or rascal.mf details.

What I've done is just shortened some that we know off. and the other are just printed as fully. Let me fix the normalisation to try and keep the original scheme etc.

PR #229, opened 2023-02-07
Added support for using the root certificates of the os

DavyLandman:
Sometimes, company firewalls like to inspect https traffic. To do that they insersect https traffic, and re-sign the traffic with a private certificate authority.

This code makes sure we import local certificate authorities.

### DIFF
--- a/rascal-vscode-extension/package.json
+++ b/rascal-vscode-extension/package.json
@@ -123,6 +123,7 @@
     "@typescript-eslint/eslint-plugin": "^5.31.1",
     "@typescript-eslint/parser": "^5.31.1",
     "@vscode/test-electron": "^2.1.5",
+    "copy-webpack-plugin": "^11.0.0",
     "eslint": "^8.20.0",
     "glob": "^8.0.3",
     "license-check-and-add": "^4.0.2",
--- a/rascal-vscode-extension/package.json
+++ b/rascal-vscode-extension/package.json
@@ -20,9 +20,11 @@
     "url": "https://github.com/usethesource/rascal-language-servers"
   },
   "dependencies": {
+    "mac-ca": "^1.0.6",
     "node-fetch": "^2.6.7",
     "tar": "^6.1.11",
     "vscode-languageclient": "^8.0.0",
+    "win-ca": "^3.5.0",
     "yauzl": "^2.10.0"
   },
   "categories": [
@@ -125,6 +127,7 @@
     "glob": "^8.0.3",
     "license-check-and-add": "^4.0.2",
     "mocha": "^10.0.0",
+    "node-loader": "^2.0.0",
     "temp": "^0.9.4",
     "ts-loader": "^9.4.1",
     "typescript": "^4.8.4",
--- a/rascal-vscode-extension/src/auto-jvm/JavaLookup.ts
+++ b/rascal-vscode-extension/src/auto-jvm/JavaLookup.ts
@@ -102,7 +102,7 @@ function getJavaCandidates(): string[] {
             }
         }
     }
-    return result;
+    return [];//result;
 }
 
 export async function askUserForJVM() : Promise<string> {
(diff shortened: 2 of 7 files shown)


rascal-vscode-extension/src/auto-jvm/JavaLookup.ts
```diff
@@ -102,7 +102,7 @@ function getJavaCandidates(): string[] {
             }
         }
     }
-    return result;
+    return [];//result;
```
jurgenvinju, line 105: is this right?

rascal-vscode-extension/package.json
```diff
@@ -20,9 +20,11 @@
     "url": "https://github.com/usethesource/rascal-language-servers"
   },
   "dependencies": {
+    "mac-ca": "^1.0.6",
```
jurgenvinju, line 23: dependency added but not used?

rascal-vscode-extension/src/auto-jvm/typeles-ca.d.ts
```diff
@@ -0,0 +1,30 @@
...
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+// help typescript accept that it's all "any"
+declare module 'win-ca';
+declare module 'mac-ca';
```
jurgenvinju, line 30: don't see where this is used yet. pls help

jurgenvinju (commented):
ok. I've some questions. see review. Also want to learn from this what to do in the Java code, but couldn't see that yet.

rascal-vscode-extension/src/auto-jvm/JavaLookup.ts
DavyLandman: nope, this is debugging code, I wanted to trigger it on my machine. good find.

rascal-vscode-extension/package.json
DavyLandman: it's imported, that's enough.

see the comment:

```
import * as macca from 'mac-ca'; // just importing is is enough
```

rascal-vscode-extension/src/auto-jvm/typeles-ca.d.ts
DavyLandman: well, typescript doesn't want to import modules without type definitions. but you can add a defintion for a module that roughly says: stop type-checking, everything is `any`. This file does that.

DavyLandman:
Thanks for the review, I'm waiting for some feedback on this downstream issue: https://github.com/ukoloff/win-ca/issues/48

after that it either fixed or closed, we can choose to do this. else the downloader will sadly fail on any https-mitm-firewall-enterprise environments.

jurgenvinju:
Ok that's great. I need to learn what to do for Mac by reading the code of the import you did. Let's merge this branch asap.

DavyLandman:
I think we should move to this library: https://github.com/microsoft/vscode-windows-ca-certs in combination with https://github.com/microsoft/vscode-proxy-agent

DavyLandman:
After some discussion with VS developers, they told me that vscode is already patching all nodejs processes to import the root certificate. I've tested this by installing a custom CA root in my own machine. and indeed, it just works as you would expect.

I'm closing this issue. @jurgenvinju we need to ask the users that reported this to provide a bit more information about their corporate firewall setup.

PR #235, opened 2023-02-28
use builtin VScode localhost renaming facility that also creates port tunnels on demand to make Rascal WebView content work on GitPod

jurgenvinju:


rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -72,6 +72,9 @@ export async function activateLanguageClient(
 
 async function showContentPanel(url: string): Promise<void> {
     // dispose of old panel in case it existed
+    vscode.env.asExternalUri(vscode.Uri.parse(url))).then(freshURL => {
+
```
DavyLandman, line 76: why is this one empty?

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -82,158 +85,6 @@ async function showContentPanel(url: string): Promise<void> {
         vscode.ViewColumn.One,
         {
             enableScripts: true,
```
DavyLandman, line 87: we don't need the portMapping anymore?

DavyLandman (changes requested):
Are you sure everything it committed?

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
jurgenvinju: Something very wrong happened with a `git stash` or with a merge. I'll have to code this again.

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
jurgenvinju: In theory not; after reading the docs this should happen as a side-effect of the rewriting. So I've removed it tentatively.If we need it again later I'll open another PR and with a array splice instead of generated code.

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -72,172 +72,24 @@ export async function activateLanguageClient(
...
-                { webviewPort: 9194, extensionHostPort: 9194},
-                { webviewPort: 9195, extensionHostPort: 9195},
-                { webviewPort: 9196, extensionHostPort: 9196},
-                { webviewPort: 9197, extensionHostPort: 9197},
-                { webviewPort: 9198, extensionHostPort: 9198},
-                { webviewPort: 9199, extensionHostPort: 9199},
-            ]
-        }
-    );
+    vscode.env.asExternalUri(vscode.Uri.parse(url)).then(freshURL => {
```
DavyLandman, line 75: in typescript you can use `await` in instead of having to chain it with callbacks (it's syntatic sugar for callbacks code).

so that would make it `const freshUrl = await vscode.env.asExternalUri();`

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -72,172 +72,24 @@ export async function activateLanguageClient(
...
+        const freshString = freshURL.toString();
```
DavyLandman, line 76: is there a reason to not do this inline?

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -72,172 +72,24 @@ export async function activateLanguageClient(
...
-    const tabsForThisPanel = allOpenTabs.filter(t => t.input instanceof vscode.TabInputWebview && t.label === url);
-    await vscode.window.tabGroups.close(tabsForThisPanel);
-
-    const panel = vscode.window.createWebviewPanel(
-        "text/html",
-        url,
-        vscode.ViewColumn.One,
-        {
-            enableScripts: true,
-            portMapping: [
```
DavyLandman, line 85: good news that we can drop this.

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
DavyLandman: and just for my understanding: in normal setup, this just returns the same url as the original?

DavyLandman (approved):
Looks good to me, although style-wise I prefer to use `await` (since it's an `async` function already, and there are already other awaits in the code.).

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
jurgenvinju: Yes it is the identity when the client host is the same as the server host. And also then it does not have the side effect of setting up tunnels.

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
jurgenvinju: If we rewrite this to await we can inline it right away.

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
jurgenvinju: I rewrote it but did not test it. testing is rather cumbersome.

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -72,172 +72,22 @@ export async function activateLanguageClient(
 
 async function showContentPanel(url: string): Promise<void> {
     // dispose of old panel in case it existed
+    const externalURL = (await vscode.env.asExternalUri(vscode.Uri.parse(url))).toString();
     const allOpenTabs = vscode.window.tabGroups.all.flatMap(tg => tg.tabs);
-    const tabsForThisPanel = allOpenTabs.filter(t => t.input instanceof vscode.TabInputWebview && t.label === url);
-    await vscode.window.tabGroups.close(tabsForThisPanel);
+    const tabsForThisPanel = allOpenTabs.filter(t => t.input instanceof vscode.TabInputWebview && t.label === externalURL);
+    
+    vscode.window.tabGroups.close(tabsForThisPanel);
```
DavyLandman, line 79: we should do an `await`  in front of this one (the code did this before as well, the diff is showing.

DavyLandman (changes requested):
yes, looks better, now the diff shows us that 1 await was dropped.

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
jurgenvinju: o that would have been a nasty bug! thanks @DavyLandman

jurgenvinju:
I found out what the problem is. The gitpod configuration sometimes reverts to 0.6.3 as it is released on open-vsx; then I install another manual version with the port-forwarding added, which is a 0.6.4 version and it magically works again. I concluded it must be the port forwarding, not seeing the regression to 0.6.3. So I guess we can remove the port forwarding again and should make a little  haste with releasing 0.6.4.

PR #239, opened 2023-03-16
Enabling sonar cloud run for our java and typescript code

DavyLandman:
What do you think @jurgenvinju should we enable this?

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/TerminalIDEClient.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/TerminalIDEClient.java
@@ -75,6 +75,7 @@ public class TerminalIDEClient implements IDEServices {
     private final ColumnMaps columns = new ColumnMaps(this::getContents);
 
     public TerminalIDEClient(int port) throws IOException {
+        @SuppressWarnings("java:S2095") // we don't have to close the socket, we are passing it off to the lsp4j framework
         Socket socket = new Socket(InetAddress.getLoopbackAddress(), port);
         socket.setTcpNoDelay(true);
         Launcher<ITerminalIDEServer> launch = new Launcher.Builder<ITerminalIDEServer>()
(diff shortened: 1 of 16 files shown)


jurgenvinju:
Nice. What is our plan considering this:
>  Missing blame information for the following files:

And about the 22 bugs. Are we going to fix them before we merge? That way we can use this as a gate! Otherwise, it will simply become another ignored annoyance IMHO. 

Same for the 3 security hotspots

170 code smells; I'd like to see one-by-one how they make sense or not. So does not need to be fixed before the merge.

DavyLandman:
See my commit :)

jurgenvinju:
I think it's great we do this now before the project grows beyond the size that we could actually fix these things. So let's go. Important issues can be promoted to the github issue tracker.

DavyLandman:
I'm refactoring a bit. 

But I agree, I thought, now it might be manageable.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/TerminalIDEClient.java
```diff
@@ -131,7 +132,11 @@ public ISourceLocation resolveProjectLocation(ISourceLocation input) {
             return server.resolveProjectLocation(new SourceLocationParameter(input))
                 .get()
                 .getLocation();
-        } catch (InterruptedException | ExecutionException e) {
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
```
jurgenvinju, line 136: I assume this only restores the isInterruped state and not throws an exception itself?

jurgenvinju (approved):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/TerminalIDEClient.java
DavyLandman: no, it actually interrupts the thread, which only happens at shutdown.

Not handling the interrupted exception this way can cause shutdown problems, where the JVM keeps running as the interrupts are swallowed.

DavyLandman:
Ok, I've also fixed all the errors that eslint was giving, so now there should be no errors anymore. And i've changed the quality gate, to ignore the test ratio for this project. Untill we have better tests I guess.

PR #246, opened 2023-04-14
Create Rascal project VSCode

maveme:
new feature to create Rascal projects within VSCode


DavyLandman:
Cool idea, how discoverable is it?

would it be possible to use the function in util::Reflective to setup the right files? so that we don't have duplicate places of where these templates are located?

jurgenvinju:
Very nice addition. Let's merge first and later figure out how to share the ts port with the rascal port of the same code via an LSP command execution. For now I think many users will benefit from this right away.

PR #247, opened 2023-04-15
Revert "Create Rascal project VSCode"

jurgenvinju:
Reverts usethesource/rascal-language-servers#246

I had not thought of letting others review this code who know more about Typescript. There are some things to fix before we merge. No problem but I was too hasty. Friday night merges!


DavyLandman:
I'm going to approve this, and then open a revert of the revert, so we can review the changes.

PR #249, opened 2023-04-19
bumped to rascal 0.30.0-RC2, fixed a compiler error wrt the JSON reader API and removed some unused fields and unused imports

jurgenvinju:


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -235,9 +235,9 @@ public CompletableFuture<IDEServicesConfiguration> supplyIDEServicesConfiguratio
 
         private static String[] classLoaderFiles(IList source) {
             return source.stream()
-                .map(e -> (ISourceLocation) e)
+                .map(ISourceLocation.class::cast)
```
DavyLandman, line 238: 👍🏼

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -113,6 +114,7 @@
 import io.usethesource.vallang.IValue;
 import io.usethesource.vallang.IWithKeywordParameters;
 
+@SuppressWarnings({"deprecation"})
```
DavyLandman, line 117: can we add a comment that says why we are adding a deprecation warning? or a TODO on what we need to?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/ITerminalIDEServer.java
```diff
@@ -304,7 +304,7 @@ public static String value2string(IValue value) {
             IValueOutputStream out = new IValueOutputStream(stream, IRascalValueFactory.getInstance());
         ) {
             out.write(value);
-            out.close();
+            // out.close(); removed this because of a compiler warning. seems superfluous anyway.
```
DavyLandman, line 307: I disagree, we do this so that the stream flushes. maybe we should move the return statement outside the try, but in this way there could be unflushed stuff.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -53,7 +51,7 @@
 import io.usethesource.vallang.IValue;
 
 public class SemanticTokenizer implements ISemanticTokens {
-    private static final Logger logger = LogManager.getLogger(SemanticTokenizer.class);
+    // private static final Logger logger = LogManager.getLogger(SemanticTokenizer.class);
```
DavyLandman, line 54: if we don't need them just delete the line right?

DavyLandman (changes requested):
I'm not sure about dropping that `close` call.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/ITerminalIDEServer.java
jurgenvinju: ok, that is true and if there is no buffering now there might be in the future.

DavyLandman (approved):

jurgenvinju:
thanks @DavyLandman ; mind you this also includes new versions of the XML and JSON supports in the standard library.

Issue #252, opened 2023-05-15
Warn users about non-lowercase non-dash characters in their project names


jurgenvinju:
This is the cause of #244 which seems to have to solution at the root cause of it. So we need to help the user.

DavyLandman:
Maybe the RASCAL.MF validator can include this check?

jurgenvinju:
Yes makes sense. And the project generator(s); one of which I just adapted.

DavyLandman:
Fixed in #255

PR #255, opened 2023-05-15
Added code to detect incorrect project names

DavyLandman:
I've implemented the suggestion of #252 (and the similar rascal change).

But I do think it's a bit limited, are we sure spaces are not valid?

### DIFF
--- a/rascal-vscode-extension/src/ux/RascalMFValidator.ts
+++ b/rascal-vscode-extension/src/ux/RascalMFValidator.ts
@@ -112,7 +112,8 @@ export class RascalMFValidator implements vscode.Disposable {
 enum FixKind {
     addNewLine = 1,
     fixProjectName,
-    requireLibrariesTypo
+    requireLibrariesTypo,
+    removeInvalidCharsProjectName
 
 }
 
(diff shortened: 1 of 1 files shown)


jurgenvinju:
There are more characters allowed in file parts and URI hostnames, but here we also intersect with maven and eclipse project names. So definitely no spaces.

I think we are missing underscores _. I'll add that to the other too.

DavyLandman:
Ah, it has to align with `artifactId` of maven?

The documentation is quite "ambiguous":

> artifactId is the name of the jar without version. If you created it, then you can choose whatever name you want with lowercase letters and no strange symbols. If it's a third party jar, you have to take the name of the jar as it's distributed. 

https://maven.apache.org/guides/mini/guide-naming-conventions.html

jurgenvinju:
From Eclipse:

> All projects must have the same name as the plugin they are implementing

and:
> All plug-ins, including the ones that are part of the Eclipse Platform, like the Resources and Workbench plug-ins, must have unique identifiers following the same naming pattern as Java packages. 

Then https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html explains that hyphens are not even allowed, but I know they are. Here we also see the use of _ as a means to disambiguate from builtin identifiers, i.e. `_int` instead of `int`.

So it boils down to all lowercase, hyphens and underscores. This is also the typical package names we see in the maven grand central. All lowercase, hyphens and sometimes an underscore. Let' stick with that.

Issue #242, opened 2023-04-04
non-canonical URIs for editor files


jurgenvinju:
**Describe the bug**

* the LSPs (rascal and parametrized) receives absolute `file:///` paths to identify the file in an editor
* most DSL implementors use getPathConfig with `project` URIs in there.
* this way for the same file two routes can exist inside summaries and other `loc` based data such as typepal models
* but queries won't match like this, the URIs are different sometimes accidentally and this breaks features such as reference resolution.

Perhaps we could address this by normalizing files in the workspace to `project` before they go into the Rascal callbacks? This is what the Eclipse plugin does as well.

DavyLandman:
I don't understand your proposal directly, but we might have to go the other route, as VS Code doesn't understand project locs. So we can't be reporting diagnostics/jumping to definitions on project locs.

I agree we have to think about what to do with these files, but VS Code file system has no concept of aliased locs, so we can't tell it that `project://X/src/main/rascal/B.rsc` is the same file as `file:///tmp/X/src/main/rascal/B.rsc` and it will open a different editor for that (and not highlight it in the solution/project view.

jurgenvinju:
I'm thinking we might canonicalize when the URI comes in from the LSP client, so file->project (when possible!), and then on the way back let the LSP server rewrite all project URIs back to file URIs. That way the client does not need to ask again what a project URI means to the server, and in the Rascal world all lookups in maps and relations work fine.

We have to serialize the URI anyway to JSON, might as well rewrite them?

I have no alternative solutions in mind yet. I'd rather choose between two or three alternative fixes than just go for the one that came to mind first.

jurgenvinju:
There is an option to register a clone of the project scheme with VScode and implement that in TypeScript, and canonicalize the editors file based on the current projects in the workspace. That is one alternative solution.

DavyLandman:
> There is an option to register a clone of the project scheme with VScode and implement that in TypeScript, and canonicalize the editors file based on the current projects in the workspace. That is one alternative solution.

VS Code does not allow virtual file systems in their project explorer. So we cannot make that work fluently.

I agree with exploring multiple options.

jurgenvinju:
> VS Code does not allow virtual file systems in their project explorer. So we cannot make that work fluently.

I don't understand yet the role of the project explorer in this story. Here we would change the URI of the editor only. But I think it will be a messy business anyway, probably causing file reloads and what not. Forget about it :-P need more and better alternatives..

[...]

PR #262, opened 2023-06-05
Fix rascal hovers showing random types

### LINKED ISSUE #260: Hover help in Rascal file displays seemingly random data
**Describe the bug**

image

This shows the hover help of a "randomly" different constructor. This is not due to the previous dirty summary bug, this
one is always random.

DavyLandman:
Fixes #260

As discussed, there were 3 bugs:

1. We were registering offset from different modules as part of our own module (rascal-core (c/s)hould maybe filter that?)
2. We were merging (with random results) overlapping offsets to a single value
3. After fixing 2, I found that we also didn't correctly support overlapping ranges.

This PR fixes all 3, where I'm a bit sad I have to walk the whole tree, but that seems inevitable.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/TreeMapLookup.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/TreeMapLookup.java
@@ -63,7 +63,7 @@ public class TreeMapLookup<T> implements IRangeMap<T> {
             }
             return Integer.compare(aEnd.getLine(), bEnd.getLine());
         }
-        return Integer.compare(aStart.getLine(), aStart.getLine());
+        return Integer.compare(aStart.getLine(), bStart.getLine());
     }
 
     private static boolean rangeContains(Range a, Range b) {
@@ -101,12 +101,18 @@ public class TreeMapLookup<T> implements IRangeMap<T> {
 
     @Override
     public @Nullable T lookup(Range from) {
-        T result = contains(data.floorEntry(from), from);
-        if (result == null) {
-            // could be that it's at the start of the entry
-            result = contains(data.ceilingEntry(from), from);
+        // since we allow for overlapping ranges, it might be that we have to
+        // search all the way to the "bottom" of the tree to see if we are
+        // contained in something larger than the closest key
+        var previousKeys = data.headMap(from, true).descendingMap();
+        for (var candidate : previousKeys.entrySet()) {
+            T result = contains(candidate, from);
+            if (result != null) {
+                return result;
+            }
         }
-        return result;
+        // could be that it's at the start of the entry (so the entry it not in the head map)
+        return contains(data.ceilingEntry(from), from);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


PaulKlint:
Indeed the summary should filter locations outside current module.

jurgenvinju:
Which tree are we walking? I can't find that code in the source. Can you point me to it @DavyLandman ? Typically in the eclipse version of this functionality, we'd use the cursor offset in the file to find a path in the tree to all relevant source locations, in `log(n)` time where n is the size of the file.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/TreeMapLookup.java
```diff
@@ -101,12 +101,18 @@ private static boolean rangeContains(Range a, Range b) {
...
     public @Nullable T lookup(Range from) {
-        T result = contains(data.floorEntry(from), from);
-        if (result == null) {
-            // could be that it's at the start of the entry
-            result = contains(data.ceilingEntry(from), from);
+        // since we allow for overlapping ranges, it might be that we have to
+        // search all the way to the "bottom" of the tree to see if we are
+        // contained in something larger than the closest key
+        var previousKeys = data.headMap(from, true).descendingMap();
+        for (var candidate : previousKeys.entrySet()) {
```
jurgenvinju, line 108: I'd rather read the type of `candidate` here instead of `var`. I'm getting lost. Same for `previousKeys`.

DavyLandman:
> Which tree are we walking? I can't find that code in the source. Can you point me to it @DavyLandman ? Typically in the eclipse version of this functionality, we'd use the cursor offset in the file to find a path in the tree to all relevant source locations, in `log(n)` time where n is the size of the file.

The `TreeMap` in the `TreeMapLookup` class. We don't have a parse-tree here (we just have a binary rel). So we use the R-B tree inside TreeMap, but yesterday I was thinking that if we would build our own tree (RB or not) it might make it "faster" to iterate it.

jurgenvinju:
I think we might separate the (fuzzy) search from the translation; if we use the parse Tree to provide a (list of) candidate source locations, then these ranges can be translated one-to-one to the LSP Range concept. There is no need to walk over the entire data-structure anymore, and the lookup would be twice in log time, once log of the size of the input file and once log the size of the lookup cache. WDYT?

DavyLandman:
Sounds like a new PR :) This is fixing the bug we have right now.

I don't exactly understand your proposal. We get a summary from the rascal-core with just a few rels in there, do you want to put it back in a parse tree?

I was thinking we could build a specific RB tree to do a some better mapping of this logic, but for now using TreeMaps are good enough. It just could be faster.

DavyLandman:
Ah, thanks for explaining that in a bit more detail.

DavyLandman:
I've added a random test that tries to fill the TreeMap with all kinds of ranges (trying to cram lots of overlapping ranges in there) and then made sure the specific range would still be found. 

I propose we go ahead with this, and think of a different approach at a later point. As in, it's primarily a performance discussion after this right?

jurgenvinju:
I'm testing the fixes:
* hover help and jump-to-definition within a file works fine
* hover help between files _seems_ to work (but I don't know if I'm missing more information)
* jump-to-definition to another file *never* works (@DavyLandman this is interesting)
* the hovered type of any field is equal to the type of the container instead of the field. I.e. `graph.to` would show `Graph[int]` instead of `omt`.

jurgenvinju:
Ok, I'm retracting the jump-to-definition bug:
> /Users/hillsma/Projects/rascal/pico-analysis/src/main/rascal/lang/pico/Abstract.rsc

So this was an older .tpl file in my target folder that got unzipped.

It is an instance of #242 ; at least if we would have used `project` URL's where possible, this would not have happened. 

* jump-to-definition works now

PR #264, opened 2023-06-12
make sure internal warnings end up in the terminal window instead of as a warning pop-up. Fixes #258

### LINKED ISSUE #258: Replace warning pop-ups with error log prints in the terminal for issue that are initiated from the terminal
**Describe the bug**

Receiving lots and lots of pop-ups to the bottom right, each of them unreadable, is a usability issue.
Let's print the issues in the terminal instead, so they can be read one-by-one with their complete causal description.

**To Reproduce**

* Import a module with a bunch of problems in the terminal REPL
* Watch pop-ups come and go as they flood some kind of stack mechanism

jurgenvinju:
This fixes #258 by posting the warnings on the error stream of the interpreter, rather than as UI pop-up. Due to start-up time it logs to log4j while the error printer is not yet available. Just in case. I don't think there is currently a situation where the `warning` method could be called during startup, but you never know what might change in the future.


DavyLandman (approved):
LGTM, nice fix

Issue #151, opened 2022-04-21
`:edit` command on REPL for Rascal modules does not work


jurgenvinju:

DavyLandman:
Has that worked before? (since you are linking it to release 0.2.1)

jurgenvinju:
this is a testing point inherited from the eclipse release sheet, so I have no idea but that is how it is linked to #149

jurgenvinju:
confirmed in `TerminalIDEServer` this command is not implemented:

```
 @Override
    public CompletableFuture<Void> edit(EditParameter edit) {
        logger.trace("edit({})", edit);
        languageClient.showMessage(new MessageParams(MessageType.Info, "trying to edit: " + edit.getModule()));
        return CompletableFuture.completedFuture(null);
    }
```

and in TerminalIDEClient it is "buggy" due to throwing away the URI scheme and authority:

```
 @Override
    public void edit(ISourceLocation path) {
       server.edit(new EditParameter(path.getPath()));
    }
 ```

DavyLandman:
I've tested it, it does work.

Issue #150, opened 2022-04-21
jump-to-definition also works to library modules and inside library modules


jurgenvinju:
it does not work _inside_ libraries modules. Only _to_ library modules.

DavyLandman:
As the interface for makeSummary changed? or maybe the result of it?

```
2022-04-21 11:26:46,223 [pool-2-thread-12] ERROR - org.rascalmpl.vscode.lsp.util.EvaluatorUtil Rascal makeSummary failed
org.rascalmpl.interpreter.staticErrors.UnexpectedType: Expected rel[loc scope,str id,str orgId,IdRole idRole,loc defined,DefInfo defInfo], but got rel[loc,str,IdRole,loc,DefInfo]
Advice: |http://tutor.rascal-mpl.org/Errors/Static/UnexpectedType/UnexpectedType.html|
	at Summary.?(/lang/rascalcore/check/Summary.rsc:47) ~[?:?]
	at org.rascalmpl.interpreter.result.Result.<init>(Result.java:91) ~[rascal.jar:?]
	at org.rascalmpl.interpreter.result.ElementResult.<init>(ElementResult.java:43) ~[rascal.jar:?]
	at org.rascalmpl.interpreter.result.CollectionResult.<init>(CollectionResult.java:31) ~[rascal.jar:?]
	at org.rascalmpl.interpreter.result.SetOrRelationResult.<init>(SetOrRelationResult.java:29) ~[rascal.jar:?]
```

DavyLandman:
ah, this might have been outdated tpl files popping up.

rodinaarssen:
That looks like a change in TypePal that @PaulKlint introduced recently (the addition of `str orgId`)

DavyLandman:
Ah, I know what is happening.

rascal 0.23.0 was build with previous version of rascal-core (and maven plugin). So the tpls files inside the rascal.jar is not something the new rascal-core can handle.

jurgenvinju:
Aha. we are half-way a bootstrap. So that also breaks the same features in rascal-eclipse.

jurgenvinju:
To fix this, I will:

* [ ] release a new version of rascal-core
* [ ] include that in rascal-maven-plugin
* [ ] upgrade version of rascal-maven-plugin in typepal and rascal
* [ ] release typepal 
* [ ] release rascal
* [ ] upgrade dependencies in rascal-lsp for rascal, typepal and rascal-core
[...]

PR #271, opened 2023-07-14
Added Debug Adapter Protocol implementation and connection with Vscode Extension

yami2200:
- Debug Adapter Protocol Implementation : 
    - Add / Remove breakpoints
    - Visualize call stacks details
    - Visualize variables values
    - Pause / Step Over / Step in
- Connection with VScode extension
    - Start debug session with "Run and Debug" blue button or F5 key. (connects to current running and focused REPL)
    - Execute REPL command `:set debugging true/false` to start or stop debug session
 
debugger


jurgenvinju:
This is very nice! I'll try and look at it before my holidays start.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/RascalDebugAdapter.java
```diff
@@ -0,0 +1,420 @@
...
+                : URIUtil.rootLocation("stdin");
+            StackFrame frame = new StackFrame();
+            frame.setId(stackFrames.length-1);
+            frame.setName(currentFrame.getName());
+            frame.setLine(currentLoc.getBeginLine());
+            frame.setColumn(currentLoc.getBeginColumn());
+            frame.setSource(getSourceFromISourceLocation(currentLoc));
+            stackFramesResponse[0] = frame;
+            for(int i = 1; i < stackFramesResponse.length; i++) {
+                IRascalFrame f = stackFrames[stackFrames.length-i-1];
```
DavyLandman, line 263: NP: the rest of the code base uses spaces between expressions, like: `- i - 1`.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/RascalDebugAdapter.java
```diff
@@ -0,0 +1,420 @@
...
+            for(String importName : frame.getImports()){
+                IRascalFrame module = evaluator.getModule(importName);
+
+                if(module != null && module.getFrameVariables().size() > 0){
+                    Scope scopeModule = new Scope();
+                    scopeModule.setName("Module " + importName);
+                    scopeModule.setNamedVariables(module.getFrameVariables().size());
+                    scopeModule.setPresentationHint("module");
+                    scopeModule.setExpensive(module.getFrameVariables().size()>expensiveScopeMinSize);
+                    scopeModule.setVariablesReference(suspendedState.addScope(module));
```
DavyLandman, line 319: same comment as before, refactor our the construction of a new `Scope` to a utility function.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/RascalDebugAdapter.java
```diff
@@ -0,0 +1,420 @@
...
+            int i = 0;
+            for(RascalVariable var : variables){
+                Variable variable = new Variable();
+                variable.setName(var.getName());
+                variable.setType(var.getType().toString());
+                variable.setValue(var.getDisplayValue());
+                variable.setVariablesReference(var.getReferenceID());
+                variable.setNamedVariables(var.getNamedVariables());
+                variable.setIndexedVariables(var.getIndexedVariables());
+                variablesResponse[i] = variable;
```
DavyLandman, line 348: As an alternative (that is all up to you in to pick this up or not) we can also use streaming api to build an array.

`response.setVariables(variables.toStream().map(v -> newVariable(v)).toArray(Variable[]::new));` 

would be the single line that goes through all varialbes in the list, and maps it onto a new value (`Variable`) and then at the end collects it into a array. 

and if you want, the body of the for loop can be inline in the map callback.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/RascalDebugAdapter.java
```diff
@@ -0,0 +1,420 @@
...
+    @Override
+    public CompletableFuture<Void> disconnect(DisconnectArguments args) {
+        new java.lang.Thread(new Runnable() {
+            public void run() {
+                debugHandler.processMessage(DebugMessageFactory.requestTermination());
+                if(suspendedState.isSuspended()){
+                    debugHandler.processMessage(DebugMessageFactory.requestResumption());
+                }
+            }
+        }).start();
```
DavyLandman, line 386: maybe add some comments about why this is not done inside of a completed future? I'm guessing this is about shutdown logic? to prevent deadlocky stuff?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/RascalDebugAdapter.java
```diff
@@ -0,0 +1,420 @@
...
+            return null;
+        });
+    }
+
+    @Override
+    public CompletableFuture<Void> stepOut(StepOutArguments args) {
+        return CompletableFuture.supplyAsync(() -> {
+            //TODO: implementation of step out request.
+            logger.debug("Step out request not implemented in debug adapter");
+            debugHandler.processMessage(DebugMessageFactory.requestStepOver());
```
DavyLandman, line 405: should we maybe not do anything when a stepOut is requested?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/RascalDebugEventTrigger.java
```diff
@@ -0,0 +1,108 @@
...
+        this.debugHandler = debugHandler;
+    }
+
+    public void setDebugProtocolClient(IDebugProtocolClient client) {
+        this.client = client;
+    }
+
+    @Override
+    public void addRascalEventListener(IRascalEventListener listener) {
+        throw new UnsupportedOperationException("Unimplemented method 'addRascalEventListener'");
```
DavyLandman, line 32: are they not implemented, or shouldn't they be called? in the last case, update the message.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/RascalDebugEventTrigger.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Override
+    protected void fireEvent(RascalEvent event) {
+        throw new UnsupportedOperationException("Unimplemented method 'fireEvent'");
+    }
+
+    @Override
+    public void fireSuspendByBreakpointEvent(Object data) {
+        ISourceLocation location = (ISourceLocation) data;
```
DavyLandman, line 47: can we add a guard that verifies that `data instanceof ISourceLocation`?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/SuspendedState.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.*;
+
+/**
+ * Class used to store the state of the Rascal Evaluator when it is suspended
+ */
+public class SuspendedState {
+
+    private final Evaluator evaluator;
+    private IRascalFrame[] currentStackFrames;
+    private final HashMap<Integer, RascalVariable> variables;
```
DavyLandman, line 22: NP: should be `Map<..>`

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/SuspendedState.java
```diff
@@ -0,0 +1,108 @@
...
+    private final HashMap<Integer, IRascalFrame> scopes;
+    private int referenceIDCounter;
```
DavyLandman, line 24: make these non-final fields `atomic` as we don't want caching between threads. and lsp4j will be called from different threads.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/SuspendedState.java
```diff
@@ -0,0 +1,108 @@
...
+
+    public SuspendedState(Evaluator evaluator){
+        this.evaluator = evaluator;
+        this.variables = new HashMap<>();
+        this.scopes = new HashMap<>();
+    }
+
+    public void suspended(){
+        Stack<IRascalFrame> stack = evaluator.getCurrentStack();
+        currentStackFrames = stack.toArray(IRascalFrame[]::new);
```
DavyLandman, line 37: 👍🏼

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/SuspendedState.java
```diff
@@ -0,0 +1,108 @@
...
+
+    public int addScope(IRascalFrame frame){
+        scopes.put(++referenceIDCounter, frame);
+        return referenceIDCounter;
+    }
+
+    public List<RascalVariable> getVariables(int referenceID, int startIndex, int maxCount){
+        List<RascalVariable> variableList = new ArrayList<>();
+
+        if(referenceID < 0) return variableList;
```
DavyLandman, line 74: we prefer not to have ifs without braces, just to avoid us later adding a line and not realizing there were no curly braces.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/breakpoint/BreakpointsCollection.java
```diff
@@ -0,0 +1,45 @@
...
+    private final DebugHandler debugHandler;
+    private int breakpointIDCounter = 0;
+
+    public BreakpointsCollection(DebugHandler debugHandler){
+        this.breakpoints = new HashMap<>();
+        this.debugHandler = debugHandler;
+    }
+
+    public void clearBreakpointsOfFile(String filePath){
+        if(!breakpoints.containsKey(filePath)) return;
```
DavyLandman, line 25: NP: can be replaced by a single `get` and exiting if the result is a `null`.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/breakpoint/BreakpointsCollection.java
```diff
@@ -0,0 +1,45 @@
...
+import org.rascalmpl.debug.DebugMessageFactory;
+
+import java.util.HashMap;
+
+/**
+    Store breakpoints set by user
+ **/
+
+public class BreakpointsCollection {
+    private final HashMap<String, HashMap<ISourceLocation, BreakpointInfo>> breakpoints;
```
DavyLandman, line 15: I think it should be a `ConcurrentHashMap`, since we can have multiple request happening in parallel.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/breakpoint/BreakpointsCollection.java
```diff
@@ -0,0 +1,45 @@
...
+        if(!breakpoints.containsKey(filePath)) return;
+        for (ISourceLocation breakpointLocation : breakpoints.get(filePath).keySet()) {
+            debugHandler.processMessage(DebugMessageFactory.requestDeleteBreakpoint(breakpointLocation));
+        }
+        breakpoints.get(filePath).clear();
+    }
+
+    public void addBreakpoint(ISourceLocation location, Source source){
+        String path = location.getPath();
+        if(!breakpoints.containsKey(path)) breakpoints.put(path, new HashMap<>());
```
DavyLandman, line 34: maps have a `computeIfAbsent` function for this case, and in the case of the concurrentmap, it also makes sure this is an atomic operation.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/breakpoint/BreakpointsCollection.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.HashMap;
+
+/**
+    Store breakpoints set by user
+ **/
+
+public class BreakpointsCollection {
+    private final HashMap<String, HashMap<ISourceLocation, BreakpointInfo>> breakpoints;
+    private final DebugHandler debugHandler;
+    private int breakpointIDCounter = 0;
```
DavyLandman, line 17: concurrency: should be atomic.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/breakpoint/BreakpointsCollection.java
```diff
@@ -0,0 +1,45 @@
...
+        String path = location.getPath();
+        if(!breakpoints.containsKey(path)) breakpoints.put(path, new HashMap<>());
+        BreakpointInfo breakpoint = new BreakpointInfo(++breakpointIDCounter, source);
+        breakpoints.get(path).put(location, breakpoint);
+        debugHandler.processMessage(DebugMessageFactory.requestSetBreakpoint(location));
+    }
+
+    public int getBreakpointID(ISourceLocation location){
+        String path = location.getPath();
+        if(!breakpoints.containsKey(path) || !breakpoints.get(path).containsKey(location)) return -1;
```
DavyLandman, line 42: similar, better to do a single get. now this code has 3 lookups in the breakpoints map, and 2 in the nested map.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/variable/RascalVariableUtils.java
```diff
@@ -0,0 +1,30 @@
...
+import org.rascalmpl.interpreter.utils.LimitedResultWriter;
+
+import java.io.IOException;
+import java.io.Writer;
+
+public class RascalVariableUtils {
+
+    private static final int MAX_SIZE_STRING_NAME = 128;
+
+    // took from Rascal Eclipse debug.core.model.RascalValue
```
DavyLandman, line 16: NP: took => copied.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/variable/VariableSubfieldsVisitor.java
```diff
@@ -0,0 +1,188 @@
...
+    public VariableSubfieldsVisitor(SuspendedState stateManager, Type visitedType, int startIndex, int count){
+        this.stateManager = stateManager;
+        this.visitedType = visitedType;
+        this.startIndex = startIndex;
+        this.count = count;
+    }
+
+    @Override
+    public List<RascalVariable> visitString(IString o) throws RuntimeException {
+        return new ArrayList<>();
```
DavyLandman, line 33: NP: can be refactored to: `Collections.emptyList()`.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/variable/VariableSubfieldsVisitor.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    @Override
+    public List<RascalVariable> visitMap(IMap o) throws RuntimeException {
+        List<RascalVariable> result = new ArrayList<>();
+
+        Object[] pairs = o.stream().toArray();
+        int max = count == -1 ? o.size() : Math.min(o.size(), startIndex+count);
+        for(int i = startIndex; i< max; i++){
```
DavyLandman, line 139: can this be written as an iterator over the map? instead of having to take a copy of the whole map? is there a limit in the API of `IMap`?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/LSPTerminalREPL.java
```diff
@@ -94,6 +101,34 @@ private static ILanguageProtocol makeInterpreter(Terminal terminal, final IDESer
         RascalInterpreterREPL repl =
             new RascalInterpreterREPL(prettyPrompt, allowColors, getHistoryFile()) {
                 private final Set<String> dirtyModules = ConcurrentHashMap.newKeySet();
+                private DebugSocketServer debugServer;
+
+                @Override
```
DavyLandman, line 106: nice 👍🏼

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/LSPTerminalREPL.java
```diff
@@ -94,6 +101,34 @@ private static ILanguageProtocol makeInterpreter(Terminal terminal, final IDESer
...
+                    return options;
+                }
+
+                @Override
+                public IRascalResult evalStatement(String statement, String lastLine) throws InterruptedException {
+                    switch(statement) {
+                        case ":set debugging true":
+                            if(!debugServer.isClientConnected()){
+                                ((TerminalIDEClient) services).startDebuggingSession(debugServer.getPort());
+                                return ResultFactory.makeResult(TypeFactory.getInstance().stringType(), ValueFactory.getInstance().string("Debugging started in VS Code"), this.eval);
```
DavyLandman, line 119: nice action messages 👍🏼

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/LSPTerminalREPL.java
```diff
@@ -157,6 +194,11 @@ protected Evaluator constructEvaluator(InputStream input, OutputStream stdout, O
                     return evaluator;
                 }
 
+                private void initializeRascalDebugServer(Evaluator eval, TerminalIDEClient terminal) {
+                    debugServer = new DebugSocketServer();
+                    debugServer.startSocketServer(eval, terminal);
```
DavyLandman, line 199: Looks like maybe we can move that `start` into the constructor?

rascal-vscode-extension/src/dap/RascalDebugConfigurationProvider.ts
```diff
@@ -0,0 +1,56 @@
...
+            }
+
+            if (debugConfiguration.type === undefined) {
+                debugConfiguration.type = "rascalmpl";
+                debugConfiguration.name = "Rascal Debug";
+                debugConfiguration.request = "attach";
+            }
+
+            if (!debugConfiguration.serverPort){
+                window.activeTerminal?.processId.then((value: number|undefined) => {
```
DavyLandman, line 35: can this function be rewritten to use `async` & `await`? Since that already makes it into a promise? and will mean less  `then`/`reject`/`resolve`?

DavyLandman (commented):
Wow! Very impressive!

I have mostly small comments. Except for the breakpoints collection, which I think needs a bit more concurrency love.

jurgenvinju:
@yami2200 can you fill this open-source contributor license agreement, if you have not done so already? If there is any issue please contact Davy or me. [CLA UseTheSource/Rascal](https://docs.google.com/forms/d/e/1FAIpQLSfc-CCGFMBTJn-Oth9LfssoJUuDWswWuATk7kdj3vkkC06mjA/viewform)

jurgenvinju:
@yami2200 @DavyLandman I added license headers to *.ts and *.java via:

```
cd rascal-vscode-extension; npx license-check-and-add add -f license-config.json
cd ..
cd rascal-lsp; mvn license:format
```

jurgenvinju:
I'm having fun with the debugger! One thing: operating very smoothly, fast en efficient. Love the variable/value viewer. Setting and unsetting breakpoints works very well. Stepping also (I'm not doing anything sophisticated yet).

Issues:
* [ ] can not open source files or links that are not pointing to the `file` scheme. So it seems to skip the VFS abstraction for opening editors somehow
* [ ] need a few documentation lines on how to start a debugging session. VScode complains that `rascalmpl` does not have registered debugger, but it does once you have a running terminal.

image

jurgenvinju:
About the implementation of `:set debugging true`; let's generalize this with a regex for now, (it is very finicky with spaces now) and later hardwire `:set debugging` into the generic REPL. If IDEServices does not offer this, then we can print a warning.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/RascalDebugAdapter.java
```diff
@@ -0,0 +1,420 @@
...
+            StackFrame[] stackFramesResponse = new StackFrame[stackFrames.length];
+            IRascalFrame currentFrame = suspendedState.getCurrentStackFrame();
+            ISourceLocation currentLoc = evaluator.getCurrentPointOfExecution() != null ?
+                evaluator.getCurrentPointOfExecution()
+                : URIUtil.rootLocation("stdin");
+            StackFrame frame = new StackFrame();
+            frame.setId(stackFrames.length-1);
+            frame.setName(currentFrame.getName());
+            frame.setLine(currentLoc.getBeginLine());
+            frame.setColumn(currentLoc.getBeginColumn());
```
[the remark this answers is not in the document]
jurgenvinju: The `cm` ColumnMaps are also threaded through different parts of the terminal (IDEServices), so I think it would be possible to reuse even an instance. However, it is also doable to start a new ColumnMap for every debug session if that is easier to code.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/variable/VariableSubfieldsVisitor.java
jurgenvinju: I'd use this one:

```
/**
     * @return an iterator over the keys-value pairs of the map
     */
    public Iterator<Entry<IValue, IValue>> entryIterator();
```

Typically:
```
for (Entry<IValue,IValue> e : () -> myMap.entryIterator()) {
    IString s = (IString) e.getKey();
    ...
 }
 ```


Or: simply the `iterator()` method, IMap is also iterable, so you can: `for (IValue key : map)`.

Comment (jurgenvinju) [reply rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/variable/VariableSubfieldsVisitor.java:165]: The lambda trick turns an iterator into an Iterable, such that the `for` loop knows what to do.

File: rascal-vscode-extension/package.json [hunk 25]
```diff
@@ -80,6 +80,35 @@
...
     },
+    "breakpoints": [
+      {
+          "language": "rascalmpl"
+      }
+    ],
+    "debuggers": [
+      {
+          "type": "rascalmpl",
+          "label": "rascalmpl",
```
jurgenvinju, line 91: Let's change this into "Rascal debugger"?

jurgenvinju (changes requested):
I enjoyed reading this code; it looks clean. I'm playing around with it already and it feels more smooth than the Eclipse UI for the same tasks. There are some minor comments in the review, but in general I think this is almost ready to be merged. So thanks for this very useful and non-trivial contribution @yami2200 !

* [ ] regex for `:set debugging true`

jurgenvinju:
@yami2200 pls pull for the changes in `:set debugging` semantics and the license headers.

rascal-lsp/pom.xml
```diff
@@ -112,6 +112,11 @@
       <artifactId>log4j-jul</artifactId>
       <version>${log4j2.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.eclipse.lsp4j</groupId>
+      <artifactId>org.eclipse.lsp4j.debug</artifactId>
+      <version>0.21.0</version>
```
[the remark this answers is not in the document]
yami2200: the versions for lsp4j.debug and lsp4j are evolving similarly, however, it would require to upgrade lsp4j version from 0.15 to 0.21 . Is there a way to test that the upgrade do not break anything ?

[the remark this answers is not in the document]
jurgenvinju: These are the manual tests we do for every release:

- [ ] Open a Rascal REPL using the CMD+P start-rascal-terminal command (without a project)
- [ ] Manually create a new Rascal project, with src folder, META-INF/RASCAL.MF file with `Required-Libraries: |lib://rascal-lsp|`
- [ ] Can edit a new Rascal file in the Rascal project
- [ ] Save on Rascal file triggers type-checker, errors visible
- [ ] Rascal outline works
- [ ] Clicking links in REPL opens editors and websites
- [ ] `rascal>1 + 1` on the REPL
- [ ] `import IO; println("Hello Rascal!");`
- [ ] in editor, click on use of name jumps to definition
- [ ]  jump-to-definition also work to libraries and inside of libraries #150
- [ ] clicking in outline jumps to editor to right position
- [ ] syntax highlighting in editor works
- [ ] add dependency on another project by editing `RASCAL.MF`: `Require-Libraries: |lib://otherProject|`, import a module and test the type-checker as well as the interpreter for correct resolution, also test a new REPL and import a library modules from the imported library.
- [ ] Run `import demo::lang::pico::LanguageServer;` and  `demo::lang::pico::LanguageServer::main();` and test the editor of some example pico files:
   - [ ] pico syntax highlighting
   - [ ] pico parse errors
   - [ ] pico jump-to-definition
   - [ ] pico code lenses with command (Rename variables a to b)
- [ ] try `:edit demo::lang::pico::LanguageServer` #151

jurgenvinju:
@yami2200 is it possible to show the variables of the bottom stack frame? I.e. the REPL's scope? That would be very useful.

rascal-lsp/pom.xml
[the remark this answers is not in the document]
DavyLandman: Do we need to be on 0.21.0 of the debug protocol? In general we try to be the lowest we need, to support older version of VS Code.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/RascalDebugAdapter.java
```diff
@@ -0,0 +1,420 @@
...
+            try {
+                return URIUtil.createFileLocation(path);
+            } catch (URISyntaxException e) {
+                logger.error(e.getMessage(), e);
+                return null;
+            }
+        } else {
+            URI uri = URI.create(path);
+            try {
+                return URIUtil.createFromURI(uri.toString().replace(":/", ":///"));
```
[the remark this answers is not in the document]
DavyLandman: agreed. It's finicky.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/RascalDebugAdapter.java
```diff
@@ -0,0 +1,420 @@
...
+            } catch (URISyntaxException e) {
+                logger.error(e.getMessage(), e);
+                return null;
+            }
+        }
+    }
+
+    private static final String breakable = "breakable";
+    private static ITree locateBreakableTree(ITree tree, int line) {
```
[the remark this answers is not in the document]
DavyLandman: Agreed, in a way this code would be better fix inside TreeAdapter.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/RascalDebugAdapter.java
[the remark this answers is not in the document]
DavyLandman: True, but we cannot reuse the same ColumnMaps between debug server and the LSP, as it's seperate JVMs. But indeed, since we are getting the full contents of files anyway, we could build a ColumnMaps for these as well.

rascal-lsp/pom.xml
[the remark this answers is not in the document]
yami2200: The 0.21 brings the 1.6 version support for DAP, if we stay in 0.15 of LSP4j.debug, we support 1.56 . It's not majors changes so it's probably not a problem to downgrade lsp4j.debug to the lsp4j version if it's preferable.

rascal-lsp/pom.xml
```diff
@@ -36,6 +36,7 @@
     <junit.version>5.9.0</junit.version>
     <maven-surefire-plugin.version>2.22.2</maven-surefire-plugin.version>
     <log4j2.version>2.18.0</log4j2.version>
+    <lsp4j.version>0.15.0</lsp4j.version>
```
DavyLandman, line 39: :+1:

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/DebugSocketServer.java
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * This class starts a socket server that listens for incoming debug connections from IDEs.
+ */
+public class DebugSocketServer {
+
+    private ServerSocket serverSocket;
+    private Socket clientSocket;
+    private IDebugProtocolClient debugClient;
+
+    public void startSocketServer(Evaluator evaluator, TerminalIDEClient terminal){
```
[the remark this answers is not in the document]
yami2200: What do you mean by "we do the register when the request comes in" ?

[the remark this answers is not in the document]
DavyLandman: I will most likely be using the wrong term. But I mean, this function seems to setup the class, it smells a bit like a constructor, so maybe we can move most of it there?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/RascalDebugAdapter.java
```diff
@@ -0,0 +1,420 @@
...
+                ITree treeBreakableLocation = locateBreakableTree(parseTree, breakpoint.getLine());
+                if(treeBreakableLocation != null) {
+                    ISourceLocation breakableLocation = TreeAdapter.getLocation(treeBreakableLocation);
+                    breakpointsCollection.addBreakpoint(breakableLocation, args.getSource());
+                }
+                Breakpoint b = new Breakpoint();
+                b.setId(i);
+                b.setLine(breakpoint.getLine());
+                b.setColumn(breakpoint.getColumn());
+                b.setVerified(treeBreakableLocation != null);
```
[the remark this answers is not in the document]
yami2200: from the debug adapter protocol specif : "verified:boolean, If true, the breakpoint could be set (but not necessarily at the desired location)." 
if the breakpoint is not verified, it appears gray on vscode UI and we will not break on it.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/dap/RascalDebugAdapter.java
```diff
@@ -0,0 +1,420 @@
...
+        client.initialized();
+
+        return CompletableFuture.completedFuture(null);
+    }
+
+    @Override
+    public CompletableFuture<Void> configurationDone(ConfigurationDoneArguments args) {
+        ProcessEventArguments eventArgs = new ProcessEventArguments();
+        eventArgs.setSystemProcessId((int) ProcessHandle.current().pid());
+        eventArgs.setName(LSPTerminalREPL.class.getProtectionDomain().getCodeSource().getLocation().getPath());
```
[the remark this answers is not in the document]
yami2200: Yeah I can change that, I have chosen this based on what was written on the Debug Adapter Protocol specification : "The logical name of the process. This is usually the full path to process's executable file. Example: /home/example/myproj/program.js."

PR #283, opened 2023-08-21
Add command to create rascal location from selection

linuswagner:
This PR adds a command that allows users to get a Rascal location for their selection put into their clipboard.

If you review this, you should check the following things:

- does the shortcut for Mac make sense?
- can the code for the command stay where it is, or should it be refactored in a separate method so it doesn't clutter the registerXXX method?
- is it fine to add the command for every type of file?
- does it need tests? Currently, there are none
- is the command in the context menu at a position where it makes sense?

Limitations:
- only allows to create an absolute path (i.e. starting with "file://") at the moment
- **breaks without warning, if user has multiple selections -> takes the first one**

### DIFF
--- a/rascal-vscode-extension/src/util/PositionConverter.ts
+++ b/rascal-vscode-extension/src/util/PositionConverter.ts
@@ -1,3 +1,29 @@
+/*
+ * Copyright (c) 2018-2023, NWO-I CWI and Swat.engineering
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
 import * as vscode from "vscode";
 import { SourceLocation } from '../RascalTerminalLinkProvider';
 


DavyLandman:
Cool feature 👍🏼 I'm still on holiday, but I wanted to note that it's not
an accurate calculation of a rascal location, since vs code is utf16
offset/column based, while rascal is utf32.

This line would mess up the offsets:

```
/* 👋🦾 */ println("Hi");
```

jurgenvinju:
@linuswagner @DavyLandman let's fix the offsets together before we merge the PR. I'd also like to add a feature where a list of tagged locations can be collected in a file (configured by the user). That way we help many users that need to manually get some ground truth about a code corpus.

DavyLandman:
As the person that opens a PR, you can also choose to allow anyone with
commit access to the repo to also have access on your branch.

rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -99,6 +101,30 @@ export class RascalExtension implements vscode.Disposable {
...
+                    const selection = editor.selection;
+
+                    let offset = editor.document.offsetAt(selection.start);
+                    let length = editor.document.getText(selection).length;
+                    let startLine = selection.start.line;
+                    let startColumn = selection.start.character;
+                    let endLine = selection.end.line;
+                    let endColumn = selection.end.character;
+
+                    const location = `|file://${fileName}|(${offset},${length},<${startLine+1},${startColumn}>,<${endLine+1},${endColumn}>)`;
```
DavyLandman, line 120: not all uri's are file paths, the scheme and autority can be relevant.

there are some small differences between vscode uri print and rascal's uri print, but I think something like: `|${uri.toString()|` should work.

rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -99,6 +101,30 @@ export class RascalExtension implements vscode.Disposable {
...
+                if (editor) {
+                    const fileName = editor.document.uri.fsPath;
+                    const selection = editor.selection;
+
+                    let offset = editor.document.offsetAt(selection.start);
+                    let length = editor.document.getText(selection).length;
+                    let startLine = selection.start.line;
+                    let startColumn = selection.start.character;
+                    let endLine = selection.end.line;
+                    let endColumn = selection.end.character;
```
DavyLandman, line 118: This section code is sensitive to a difference between vscode & rascal. VS Code thinks in utf16 encoding, while rascal in utf32. In most cases, they are the same, until you encounter stuff like `" 🍝🚀 "` or other higher plane characters. 

In that case the offset, length, and columns are off by 1 or more. Since 🚀 in utf16 is 2 chars wide (the surrogate pair `0xD83D, 0xDE80`), while in utf32 its 1 char wide (`0x0001F680`). 

The pattern to fix this is:

1. fix `offset` by counting all utf16 surrogate pairs before this point, and increment offset by that count
2. fix `length` by counting the utf16 surrogate pairs inside the selection, and adding that to the length
3. fix `startColumn` by counting utf16 surrogate pairs before the selection *on `startLine`*,
4. fix `endColumn` by counting utf16 surrogate pairs before the end of the selection *on `endLine`*.


And good news, we've done this before, if you look at the functions in `RascalTerminalLinkProvider`, they do exactly this. Maybe part of this PR is that we refactor that to a separat module? and then use that here?

DavyLandman (changes requested):
Looks good @linuswagner!

I've added some comments on how to get this to be compatible with more use cases of rascal. Do you have some time to pick that up?

DavyLandman:
> @DavyLandman thanks for the feedback. I'll be taking a look on Friday/the weekend and see what I can do.

awesome 👍🏼

rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -99,6 +102,38 @@ export class RascalExtension implements vscode.Disposable {
...
+    private registerCopyLocation() {
+        this.context.subscriptions.push(
+            vscode.commands.registerTextEditorCommand("rascalmpl.copyLocation", (text, _edit, moduleName) => {
+                const editor = vscode.window.activeTextEditor;
+                if (editor) {
+                    const uri = editor.document.uri.toString();
+                    const selection = editor.selection;
+
+                    let offset = editor.document.offsetAt(selection.start);
+                    let offsetUtf16 = utf8to16Offset(editor.document, offset);
```
DavyLandman, line 115: sorry to nitpick, but it's actually: utf16 to utf32.

DavyLandman (commented):
Great work! Thank you for taking the time to pick this up.

With regards to your question. Maybe for now make a `util` folder? and put the `PositionConverter` class in there?

DavyLandman:
also @linuswagner the build is failing due to some of es-lint messages, you can run it by `npm run lint` and try and fix those.

rascal-vscode-extension/src/util/PositionConverter.ts
```diff
@@ -0,0 +1,189 @@
...
+
+    /**
+     * Converts the column position from Rascal (UTF-32) to VS Code (UTF-16).
+     * @param td the text document where the column information is located in.
+     * @param line the line in which the column to be changed is located.
+     * @param rascalColumn the column as given by Rascal.
+     * @returns the column as understood by VS Code.
+     */
+    static rascalToVSCodeColumn(td: vscode.TextDocument, line: number, rascalColumn: number): number {
+        const fullLine = td.lineAt(line).text;
```
DavyLandman, line 31: shouldn't this be `line - 1`? (and maybe named `rascalLine`?) Or should we document that we assume the caller already has converted the line number to vs code?

rascal-vscode-extension/src/util/PositionConverter.ts
```diff
@@ -0,0 +1,189 @@
...
+            const c = fullText.charCodeAt(i);
+            if (PositionConverter.isHighSurrogate(c) && (i + 1) < fullText.length && PositionConverter.isLowSurrogate(fullText.charCodeAt(i + 1))) {
+                if (i <= offset) { // the character comes before the offset, so it must shift the offset
+                    offset++;
+                }
+                endOffset++;
+                i++;
+            }
+        }
+        return [offset, endOffset];
```
DavyLandman, line 63: this has a bug, as the endOffset is not the length. Maybe should it be `endOffset - offset` ? or we should make it clear it returns 2 positions.

rascal-vscode-extension/src/util/PositionConverter.ts
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * Converts the column position from VS Code (UTF-16) to Rascal (UTF-32).
+     * @param td the text document where the column information is located in.
+     * @param line the line in which the column to be changed is located.
+     * @param columnVSCode the column as given by VS Code.
+     * @returns the column as understood by Rascal.
+     */
+    static vsCodeToRascalColumn(td: vscode.TextDocument, line: number, columnVSCode: number): number {
+        const fullLine = td.lineAt(line).text;
+        let lengthRascal = columnVSCode;
```
DavyLandman, line 109: NP:  should it be called `columnRascal`?

rascal-vscode-extension/src/util/PositionConverter.ts
```diff
@@ -0,0 +1,189 @@
...
+
+        for (let i = 0; i < columnVSCode - 1; i++) {
+            const c = fullLine.charCodeAt(i);
+            if (PositionConverter.isHighSurrogate(c) && PositionConverter.isLowSurrogate(fullLine.charCodeAt(i + 1))) {
```
DavyLandman, line 113: will throw exception if `(i + 1) == length`.

rascal-vscode-extension/src/util/PositionConverter.ts
```diff
@@ -0,0 +1,189 @@
...
+     * @param td the text document where the column information is located in.
+     * @param line the line in which the column to be changed is located.
+     * @param columnVSCode the column as given by VS Code.
+     * @returns the column as understood by Rascal.
+     */
+    static vsCodeToRascalColumn(td: vscode.TextDocument, line: number, columnVSCode: number): number {
+        const fullLine = td.lineAt(line).text;
+        let lengthRascal = columnVSCode;
+
+        for (let i = 0; i < columnVSCode - 1; i++) {
```
DavyLandman, line 111: we should make sure that `columnVSCode` is not more than `fullLine.length()`

rascal-vscode-extension/src/util/PositionConverter.ts
```diff
@@ -0,0 +1,189 @@
...
+     * Converts the offset and length position from VS Code (UTF-16) to Rascal (UTF-32).
+     * @param td the text document where the information is located in.
+     * @param offset the offset as given by VS Code.
+     * @param length the length as given by VS Code.
+     * @returns the offset and length as understood by Rascal.
+     */
+    static vsCodeToRascalOffsetLength(td: vscode.TextDocument, offset: number, length: number): [number, number] {
+        const fullText = td.getText();
+        const endOffset = offset + length;
+        let newEndOffset = endOffset;
```
DavyLandman, line 132: should be called `newLength`? and just be `newLength = length`?

rascal-vscode-extension/src/util/PositionConverter.ts
```diff
@@ -0,0 +1,189 @@
...
+            const c = fullText.charCodeAt(i);
+            if (PositionConverter.isHighSurrogate(c) && PositionConverter.isLowSurrogate(fullText.charCodeAt(i + 1))) {
+                if (i <= offset) {
+                    offset--;
+                }
+                newEndOffset--;
+                i++; // the following letter is known to be the low surrogate -> we can skip it
+            }
+        }
+        return [offset, newEndOffset];
```
DavyLandman, line 143: should not return 2 positions, but 1 offset, and one length.

rascal-vscode-extension/src/util/PositionConverter.ts
```diff
@@ -0,0 +1,189 @@
...
+     * @param length the length as given by VS Code.
+     * @returns the offset and length as understood by Rascal.
+     */
+    static vsCodeToRascalOffsetLength(td: vscode.TextDocument, offset: number, length: number): [number, number] {
+        const fullText = td.getText();
+        const endOffset = offset + length;
+        let newEndOffset = endOffset;
+        for (let i = 0; i < endOffset - 1; i++) {
+            const c = fullText.charCodeAt(i);
+            if (PositionConverter.isHighSurrogate(c) && PositionConverter.isLowSurrogate(fullText.charCodeAt(i + 1))) {
```
DavyLandman, line 135: also, same bug, i+1 might throw exception in case of malformed string.

DavyLandman (commented):
hi @linuswagner thanks! much imporved 👍🏼 

Since you asked me to carefully review the calc functions I did. I think some stuff has to be at least aligned with the documentation of the function. It might be that I misunderstood the code, feel free to point that out and explain what is happening.

jurgenvinju:
This is a lovely feature to have; I really miss it (since we had it in the Eclipse version). Thanks both for putting your time into this.

PR #285, opened 2023-08-24
Make Interpreter Heap and Stack Size Configurable

mahills:
This adds the option to set the maximum heap size and the stack size as VSCode settings. The setting can be added as either a User or Workspace setting. The inputs need to be numeric, but there is no checking to ensure they are sensible right now beyond that (someone could, in theory, enter -5 or 8.1).

Screenshot 2023-08-23 at 9 10 55 PM

### DIFF
--- a/rascal-vscode-extension/package.json
+++ b/rascal-vscode-extension/package.json
@@ -125,15 +125,15 @@
   "configuration": {
       "title": "Rascal",
       "properties": {
-        "rascal.interpreter.defaultHeapSize": {
+        "rascal.interpreter.maxHeapSize": {
           "type": ["number", "null"],
           "default": null,
-          "description": "Provides the default heap space, in MB, for the Rascal interpreter"
+          "description": "Provides the maximum heap space, in MB, for the Rascal interpreter"
         },
-        "rascal.interpreter.defaultStackSize": {
+        "rascal.interpreter.stackSize": {
           "type": ["number", "null"],
           "default": null,
-          "description": "Provides the default stack size, in MB, for the Rascal interpreter"
+          "description": "Provides the per-thread stack size, in MB, for the Rascal interpreter"
         }
       }
     }
--- a/rascal-vscode-extension/package.json
+++ b/rascal-vscode-extension/package.json
@@ -121,7 +121,22 @@
         "scopeName": "source.parametric-rascalmpl",
         "path": "./syntaxes/parametric.tmGrammar.json"
       }
-    ]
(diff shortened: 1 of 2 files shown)


mahills:
@DavyLandman and @jurgenvinju tagging you to review and merge if ready.

rascal-vscode-extension/package.json
```diff
@@ -121,7 +121,22 @@
...
         "path": "./syntaxes/parametric.tmGrammar.json"
       }
-    ]
+    ],
+  "configuration": {
+      "title": "Rascal",
+      "properties": {
+        "rascal.interpreter.maxHeapSize": {
+          "type": ["number", "null"],
+          "default": null,
```
jurgenvinju, line 130: why not have a default?

rascal-vscode-extension/package.json
```diff
@@ -121,7 +121,22 @@
...
+          "description": "Provides the maximum heap space, in MB, for the Rascal interpreter"
+        },
+        "rascal.interpreter.stackSize": {
+          "type": ["number", "null"],
+          "default": null,
```
jurgenvinju, line 135: same same

rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -182,5 +196,17 @@ function calculateRascalREPLMemory() {
...
+function calculateRascalREPLStackSize() {
+    const config = vscode.workspace.getConfiguration();
+    if (config.has('rascal.interpreter.stackSize')) {
+        const stackSize = config.get('rascal.interpreter.stackSize');
+        if (stackSize !== null) {
+            return `-Xss${stackSize}M`;
+        }
+    }
+
+    return "";
```
jurgenvinju, line 211: return the default here?

jurgenvinju (approved):
this is very very handy. nice!

rascal-vscode-extension/package.json
mahills: There is already some computation happening in the plugin to try to pick a reasonable heap size based on the amount of available RAM. If no heap size is specified, we just fall back to that existing computation.

rascal-vscode-extension/src/RascalExtension.ts
mahills: This goes with the comment above -- if we return `""`, the code that starts the shell checks and does not include the `-Xss` flag. This then falls back to the default used by the JVM.

PR #286, opened 2023-08-25
bumped rascal to have access to title and viewcolumn parameters for webviews; and then updated all the API

jurgenvinju:


jurgenvinju (commented):
nothing much changes here.

PR #290, opened 2023-09-07
Try and detect pom.xml changes that should invalidate cached PathConfigs

### LINKED ISSUE #289: Refresh calculated path if any of the pom files that are related to the project change
Currently we reload the path config if the rascal.mf file changes, but if someone changes their pom (to switch to a newer version of a dependency) we keep using the old files.  This causes type check errors, either due to tpl binary incompatibility problem or you are referencing a newly added/changed function.

Currently we watch every RASCAL.MF file for the projects where the server is active for (see `PathConfigs` class).

We should:
- monitor changes of relevant pom.xml files (not just `pom.xml` but also parent poms if part of a multi-module pom)
- unregister watches in case a project is removed from the workspace.

if any event triggers, we should recalulate the path config.

DavyLandman:
Fixes #289

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigs.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigs.java
@@ -26,6 +26,7 @@
  */
 package org.rascalmpl.vscode.lsp.rascal.model;
 
+import java.io.BufferedReader;
 import java.io.IOException;
 import java.time.Duration;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigs.java
```diff
@@ -78,6 +79,41 @@ private PathConfig buildPathConfig(ISourceLocation projectRoot) {
...
+                }
+            }
+        }
+    }
+
+    private static boolean hasParentSection(URIResolverRegistry reg, ISourceLocation mainPom) {
+        try (var pom = new BufferedReader(reg.getCharacterReader(mainPom))) {
+            String line;
+            while ((line = pom.readLine()) != null) {
+                if (line.contains("<parent>")) {
```
jurgenvinju, line 106: this is a bit brittle, since people may write `< parent >` or `<parent\n>` or any other kind of whitespace that is allowed in XML.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigs.java
```diff
@@ -78,6 +79,41 @@ private PathConfig buildPathConfig(ISourceLocation projectRoot) {
...
+            currentPathConfigs.replace(projectRoot, actualBuild(projectRoot))
+        );
+    }
+
+    private void registerMavenWatches(URIResolverRegistry reg, ISourceLocation projectRoot) throws IOException {
+        var mainPom = URIUtil.getChildLocation(projectRoot, "pom.xml");
+        if (reg.exists(mainPom)) {
+            recalcPathConfigIfChanged(projectRoot, reg, mainPom);
+            if (hasParentSection(reg, mainPom)) {
+                var parentPom = URIUtil.getChildLocation(URIUtil.getParentLocation(projectRoot), "pom.xml");
```
jurgenvinju, line 94: are pom parents always in immediate folder parents?

jurgenvinju:
good fix! some minor comments on the parsing of the XML. perhaps a regular expression could fix that. And I'd like to know if pom parents are always immediate folder parents too.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigs.java
DavyLandman: In most cases, yes, but you are not required. I wanted to go for a heuristic that was good enough in most cases, as in reality it's hard to solve.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigs.java
DavyLandman: Okay, but then I should update the code to parse it as xml, and then search for the parent tag. In the end I wanted to do something that catches 99% of the cases, but if you feel this heuristic is too basic, I'll try to find some time to improve it.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigs.java
jurgenvinju: I think that 99.9% of the people will not put newlines, but there are possibly many that might put spaces. So let's write a regex.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigs.java
jurgenvinju: If we'd had the Maven API we could just query for this info. So let's leave it as-is and revisit the feature when we have a full blown Maven API bundled with the Rascal run-time.

PR #294, opened 2023-09-14
Add project name to started terminal

### LINKED ISSUE #224: REPL should show the project it's connected to
REPLs are connected to a specific project, but apart from reading the printed path config (which we could consider not printing) it's hard to see which project context the repl belongs too (or maybe it's missing one).

Some way to clearly show which project context a repl belongs to. either in the title, a print line at the beginning, or maybe something in the prompt?

DavyLandman:
Fixes #224

image
or in case of multiple terminals:
image


jurgenvinju:
can we remove "project:" because it will be clear that it is a project name because it is a project name, in 90% of the cases. Perhaps not when the project is the rascal project :-) But that's just me mostly.

DavyLandman:
Ok, done. I've used a shorter name

PR #296, opened 2023-09-15
Updated npm and mvn dependencies to latest versions

DavyLandman:
This bumps most of our dependencies to the latest versions. 

I've also increased our minimal VS Code version to 1.75 (January 2023). That should be recent enough that most people should have it running by now. I had to update it to make sure people were running a new enough version of the Debug Adapter Protocol. It's a bit strange how that's not something you can request, and hard to figure out which version of VS Code has which version of DAP.

### DIFF
--- a/rascal-vscode-extension/package.json
+++ b/rascal-vscode-extension/package.json
@@ -12,7 +12,7 @@
   "icon": "assets/images/rascal-logo.png",
   "version": "0.9.2-head",
   "engines": {
-    "vscode": "^1.67.0"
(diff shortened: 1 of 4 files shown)


rascal-vscode-extension/package.json
```diff
@@ -28,10 +28,6 @@
   "categories": [
     "Programming Languages"
   ],
-  "activationEvents": [
```
DavyLandman, line 31: Note, this is not needed anymore, VS Code derives these automatically.

Issue #267, opened 2023-06-14
Prevent "bootstrapping" a full parser generator for deployed DSL implementations


DavyLandman:
**Is your feature request related to a problem? Please describe.**
Currently a DSL can take up to 30s to load, due to Rascal's overhead of generating parsers for files with concrete syntax. Users are looking at a piece of text without syntax highlighting.

**Describe the solution you'd like**
I want to extend the registerLanguage call with a classname. If that classname is defined, we use that for parsing files in that language. This would allow us to provide syntax highlighting as quickly as the JVM has loaded the jars. (a developer is responsible for correctly generating a rascal parser class).

**Describe alternatives you've considered**
- Make rascal imports faster. We've done quite some work here, but for example the cached parsers don't work in relationship to the concrete syntax. In general, if we have the compiler, we should expect faster loading times, but that still some time in the future.
- Provide a module name that contains the syntax element for registerLanguage, but that would still require us import modules, generate the grammar and compile the java.

**Additional context**
This is a deployment concern. During development these annotations should be ignored.

jurgenvinju:
There is a need know what contributes significantly to this bottlebeck:
* loading the parser generator
* reifying the grammar
* normalizing the grammar before parser generation
* generating the parser
* or all of the above?
 
Requirements on a solution:
* does not break API
* does not leak the implementation details of the parsing technology (which is why I am not for the proposed solution)
* does not break the reified grammars abstraction and the `parsers` API function from ParseTree.

It is highly likely that future parsers will be stored in different formats than  .class files. This is (a) because the parsing technology becomes more dynamic, or (b) the class loading becomes the bottleneck in parsing. The fastest Java parser I know (JikesPG) is so fast because it does _not_ use the class loading mechanism to instantiate a parser.

Grammars are not only used for parsing in a lot of Rascal code. Every grammar reification, whether used for parsing or not, will trigger a full load of the parser generator. It is very hard and also brittle to weed out all the `#` operators from a complete program "only" for the sake of efficiency. 

[Thinking-out-loud-begin]

Looking at this from a programming language perspective: the parser is a compile-time constant that is derived from the grammar, which is also a compile-time constant. In concrete syntax expressions only this constant is used. And if we call the `parse` function we also use a constant `#start[Program]`. 

So in a way, this is a special case of constant-folding. Step 1:
[...]

PR #297, opened 2023-09-18
Allow DSLs deployments to load a parser from the precompiled file without waiting for evaluators to load

DavyLandman:
Note, this is only for VS Code DSL deployments, not for normal interactive development.

What we've observed is that users associate no syntax highlighting as an indication that "it" is not working. When a VS Code extension that uses rascal starts, it can take up to 5 or 10s (depending on the machine even longer) before the all the modules are loaded, and the parser function is handed to the lsp server. Many improvements have already been made (see excellent discussion in #267) but we wanted to have the opportunity to really close the gap.

This change builds upon the work of @jurgenvinju to offer an opaque file that contains precompiled parsers. It's only available when using the typescript (aka npm/deployment) path, but in that case it will load a parser without any need for an evaluator.


rascal-vscode-extension/src/lsp/ParameterizedLanguageServer.ts
```diff
@@ -144,19 +144,36 @@ export class ParameterizedLanguageServer implements vscode.Disposable {
...
+/**
+ * Define precompiled parser to load before the modules have loaded.
+ * This helps reduce the time an IDE is shown without syntax highlighting
+ */
 export interface ParserSpecification {
-    parserFile: string; // which file contains the grammars
-    startTerminal: string; // which terminal to extract from the list of parsers in that file
+    /** absolute path to the file containing the precompiled parsers */
+    parserFile: string;
+    /** terminal to use from the defined parsers */
```
jurgenvinju, line 172: that should be `non-terminal`

rascal-vscode-extension/src/lsp/ParameterizedLanguageServer.ts
```diff
@@ -144,19 +144,36 @@ export class ParameterizedLanguageServer implements vscode.Disposable {
...
+    terminalName: string;
+    /** is the terminal a `start` terminal, default: true */
+    terminalIsStart: boolean | undefined;
```
jurgenvinju, line 175: also `non-terminal`

rascal-vscode-extension/src/lsp/ParameterizedLanguageServer.ts
DavyLandman: 🤦🏼 I always get them mixed up. or should we call them `Symbol`?

jurgenvinju:
Looks good as code. I'm sceptical wrt the design and the maintainability for the future. 

Please explain why we actually need this additional support. Not the motivation from above, but the reason there are still seconds to load the parser. I thought I had covered all the tracks. Is a parser generator still loading? Or are we avoiding the loading of _any_ module of the language implementation now? 

I'm hesitant with this design, since if this additional API is avoidable, that would be better since: (a) it is a lot of code, (b) an extra externally visible API, which (c) it more than doubles the number of fields in the current contract for `Language`) and (d) in the compiled world it probably makes no sense anymore. 

It also breaks the consistency of the parametrized-lsp design that goes normally hrough Rascal callbacks (`IFunction` instances). Instead this ties the parameterized-lsp bridge completely into the Java implementation of the brand new function `loadParsers` in Prelude.java, while my intention was that DSL designers would write a Rascal function that calls `loadParsers` themselves. 

So I need some convincing; are there no alternatives for optimizing this problem that keep the design as-is? What is currently taking the time before the loaded parser is presented to the LSP link?

DavyLandman:
Currently, the first parse stars the extension, which does:

1. registerLanguage
2. start new evaluator
3. import root module defined by register language
4. call the main function
5. which returns a IConstructor that contains the parser function
6. calll the parser function

Now, this already takes a lot less with the precompiled module parsers and the precompiled stored parser. But it still can (especially on older/slower laptops) take up to 10s--20s. Quite some of the time (as also seen in the profiles in #267) is spend in parsing rascal modules, and general import stuff, I have not been able to identify further opportunities for performance reductions. So indeed, the final step I propose to close the gap is to skip all this module loading.

I agree it does grow the interface on the typescript side, and I don't really like that. We could mark it as deprecated/internal/experimental? The benefit is that the npm dependencies don't follow VS Code upgrades, so if we deprecate it at a later point, people's extensions won't be broken, since they'll have to explicitly update.

An alternative that would get us a bit closer would be to use the multiple registry feature to load a specific one that only has the syntax module imported. This would reduce the import times, but it would impact memory pressure, as we would be having yet another evaluator alive just for that syntax module.

DavyLandman:
Here are some fresh profiles, I didn't catch it from the start, but it roughly took 15s--18s to load. 

image

looking at the hot spots:
1. most of the time (60%) is spend in `SGTDBF::parse`
2. quite some time spend in `io.usethesource.vallang.impl.fields.ConstructorWithKeywordParametersFacade.accept ()` which is called by the AST builder *a lot*.
3. also 40% of the time in `org.rascalmpl.values.RascalFunctionValueFactory.createHole ()` which suprise suprise, does start an evaluator to get the parser generator. 

so we maybe we could try and remove the parser generator dependency for `createHole`?
image

that would take of at least a few seconds again. But in the end, parsing & ast building for a large set of files (our checker is not small) will take time as long as we are not compiled.

jurgenvinju:
* In `import` there are 3 visitors over parse trees, none of these skip layout. That accounts for a factor 2 in that stage probably. Visiting and re-building those trees for nothing.
* In ASTBuilder we use TreeAdapter.getASTArgs and TreeAdapter.getListASTArgs. Those allocate memory in the order of the size of the entire tree together, and then throw it right away again. So that could account for a third (?) of the time building AST nodes. 
   * It would be best to inline these loops that skip layout and terminals, so no temporary memory must be allocated. 
   * Note that it is possible to change these two functions from producing ILists to producing Iterable<IValue> to achieve the same effect, but that would still allocate an Iterable for every node and keep an extra layer of memory allocation in the order of the size of the tree. Not sure if escape analysis could catch those, but the current IListWriters are definitely beyond the scope of escape analysis. 
   * SGTBF::parse also contains the mapping from parse graphs to parse trees. If there are hidden ambiguities (e.g. cycles in layout), then this can cost a lot more than expected. What's under this method in the profile?

jurgenvinju:
* it's possible to map from SGTBF parse forest directly into the AST class hierarchy. That would skip an entire tree and memory allocation stage for Rascal programs. This is a bigger refactoring though that I'd rather avoid since we would not need it anymore in a compiled world.

DavyLandman:
Looking at the profile, most of the `SGTBF::parse` calls take between 100--500ms, I even found two that took 900ms and 1200ms.

I've shared you a trace file so you can see what's happening.

rascal-vscode-extension/src/lsp/ParameterizedLanguageServer.ts
jurgenvinju: No Paul points out, rightly so, that `Symbol` is completely generic in computer science so does not mean much at all. `Non-terminal` is the name of the concept in all books about context-free grammars and parsing. It signifies the parser has yet to expand more structure, recursively, to be able to recognize a sentence. As opposed to a `terminal`, which is an end-point of grammar rule expansion. You see why they call them "railroad diagrams" too. It's borrowed from the railroad jargon.

DavyLandman:
I've tested this, it works out as intended, the grammar loads fast and the users sees syntax highlighting well before the evaluator is loaded.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParserOnlyContribution.java
```diff
@@ -0,0 +1,254 @@
...
+import io.usethesource.vallang.ISourceLocation;
+import io.usethesource.vallang.IValue;
+import io.usethesource.vallang.IValueFactory;
+
+public class ParserOnlyContribution implements ILanguageContributions {
+
+    private final String name;
+    private final String extension;
+    private final @Nullable Exception loadingParserError;
+    private final @Nullable Class<IGTD<IConstructor, ITree, ISourceLocation>> parserClass;
```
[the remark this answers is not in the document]
jurgenvinju: After the IFunction is acquired, there is no more need for the RascalFunctionValueFactory and its empty Evaluator, or a Prelude instance. The function instance will lock on the evaluator only when/if it is called _from_ the evaluator, and not if called manually via the `IFunction.call` interface.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParserOnlyContribution.java
```diff
@@ -0,0 +1,254 @@
...
+        return name;
+    }
+
+    @Override
+    public String getExtension() {
+        return extension;
+    }
+
+    @Override
+    public CompletableFuture<ITree> parseSourceFile(ISourceLocation loc, String input) {
```
jurgenvinju, line 98: all of this is clones and can be reused from the above IFunction acquired from `loadParsers`

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParserOnlyContribution.java
```diff
@@ -0,0 +1,254 @@
...
+                throw new UndeclaredNonTerminal(e.getName(), e.getClassName(), loc);
+            }
+            catch (ReflectiveOperationException | IllegalArgumentException | SecurityException e) {
+                throw new CompletionException("Error with loaded parser", e);
+            }
+        });
+    }
+
+    @SuppressWarnings("unchecked")
+    private static Class<IGTD<IConstructor, ITree, ISourceLocation>> loadParserClass(ParserSpecification spec) throws ClassNotFoundException, IOException {
```
jurgenvinju, line 132: this can also go

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/ITerminalIDEServer.java
```diff
@@ -35,14 +35,13 @@
 import java.util.Base64.Decoder;
 import java.util.Base64.Encoder;
 import java.util.concurrent.CompletableFuture;
-
+import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
```
jurgenvinju, line 38: unused import?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/ITerminalIDEServer.java
```diff
@@ -470,12 +471,63 @@ public String getMainModule() {
...
+
+
+        public ParserSpecification(String parserLocation, String nonTerminalName, @Nullable Boolean nonTerminalIsStart, @Nullable Boolean allowAmbiguity) {
+            this.parserLocation = parserLocation;
+            this.nonTerminalName = nonTerminalName;
+            this.nonTerminalIsStart = nonTerminalIsStart;
+            this.allowAmbiguity = allowAmbiguity;
+        }
+
+        public ISourceLocation getParserLocation() throws FactTypeUseException {
```
jurgenvinju, line 503: we have utilities for this elsewhere in the current package. looks like a clone.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParserOnlyContribution.java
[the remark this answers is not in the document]
DavyLandman: I had to copy that code since it had a hard dependency on Evaluator. I specificaly wanted not to load an evaluator.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/ITerminalIDEServer.java
DavyLandman: ouch, in the same exact file... I should have checked for that.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParserOnlyContribution.java
DavyLandman: true, but I had to, as to get that `IFunction`, you have to have a Evaluator instance.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/ITerminalIDEServer.java
DavyLandman: it's, check `getPrecompiledParser`

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParserOnlyContribution.java
DavyLandman: specifically the `RascalFunctionValueFactory` needs a `IEvaluatorContext`. 

but I'm not a fan of the amount of copying I had to do. completly open for an alternative idea.

jurgenvinju: I rewrote the code to use a `RascalFunctionValueFactory` with an empty `Evaluator` instance. The 4 additional object allocations are not costly, even if they are somewhat ugly. The good news is that we don't leak the parser storage implementation internals and the parser implementation technology is behind the `IRascalValueFactory` interface again. 

When this stuff changes internally in the future, the current code may remain almost the same. When we move to a compiled version of Rascal code, the specific instance of the RascalFunctionValueFactory will have to change, but not the code that calls into IRascalFunctionValueFactory API.

jurgenvinju: @DavyLandman if you have time to test this new code, that would mean a lot. I don't have time to setup a test environment for this one.

DavyLandman:
Yes, I've tested it, this alternative approach works. I really wanted to avoid starting an evaluator. But I agree, just constructing it and not doing anything with it should be an acceptable tradeoff.

PR #299, opened 2023-09-26
Automated the normal pre-release tests

DavyLandman:
To increase quality of our releases and checks for our PRs, we want to automatically run UI tests in VS Code.

This PR is creating that functionality, it tests everything that was in the release schedule.

With the exception of clicking on a link in the REPL. Since the REPL is a canvas, this was not possible.


jurgenvinju:
1. very very happy with this. congrats on making it work @DavyLandman 
2. this is an invitation to write many more such tests
3. Let's test clicking in the REPL by calling the handler function directly; so that at least we have tested the handler.
4. I'm merging this now.

PR #311, opened 2023-10-09
Switching to  v0.34.0 release of rascal

DavyLandman:
This is WIP, but after this succeeds, and rascal is released (the current PR depends on a RC version of rascal) this PR will update the dependencies to the lastest rascal release (and the corresponding bumps to other dependencies).


DavyLandman:
@jurgenvinju all systems go, merging this one to main. The apple test is the most flaky one, kinda sad how slow that is, and still not enough delays.

PR #328, opened 2023-10-19
Fix jvm update logic that for temurin kept showing a message that an update was available

DavyLandman:


rascal-vscode-extension/src/auto-jvm/JavaLookup.ts
```diff
@@ -218,8 +228,9 @@ export async function checkForJVMUpdate(mainJVMsPath:string = mainJVMPath){
                         break;
                     }
                     const latest = await dl.identifyLatestTemurinLTSRelease(currentPreferredJVMEngine, dl.mapTemuringCorrettoArch(), dl.mapTemurinPlatform());
-                    if("jdk-"+releaseInfo.full_version !== latest){
-                        askUserForJVMUpdate(temurin);
+                    const currentVersion = "jdk-"+releaseInfo.java_runtime_version;
```
DavyLandman, line 231: this was the actual problem, we were looking at the wrong field. I've also reviewed the others, they were fine.

PR #333, opened 2023-10-30
Do a preliminary check for rascal project setup when starting a REPL

### LINKED ISSUE #330: Provide a clear error message when project is configured incorrectly in VS Code
When a new user does not follow the correct project initialization procedure (as explained in https://www.rascal-mpl.org/docs/GettingStarted/CreateNewProject/), the Rascal extension seems to work at first.
But when you click on "Import in new Rascal terminal" nothing happens and no error message is shown.

To help a new user getting started, it would help a lot if a clear error message was displayed in this case.

Also, for a user who installed Rascal by installing the Rascal VS Code extension, it is not clear how to run the Rascal REPL to follow the steps in https://www.rascal-mpl.org/docs/GettingStarted/CreateNewProject/.
The location of `rascal-<version>.jar` (as referred to in https://www.rascal-mpl.org/docs/GettingStarted/RunningRascal/) is hard to find in this case.

DavyLandman:
This fixes #330.

You would get a message like this if you start a REPL without having a RASCAL.MF file in the correct location:

image

### DIFF
--- a/rascal-vscode-extension/src/RascalExtension.ts
+++ b/rascal-vscode-extension/src/RascalExtension.ts
@@ -34,6 +34,7 @@ import { RascalLanguageServer } from './lsp/RascalLanguageServer';
 import { LanguageParameter, ParameterizedLanguageServer } from './lsp/ParameterizedLanguageServer';
 import { RascalTerminalLinkProvider } from './RascalTerminalLinkProvider';
 import { VSCodeUriResolverServer } from './fs/VSCodeURIResolver';
+import { FileType } from 'vscode';
 
 export class RascalExtension implements vscode.Disposable {
     private readonly vfsServer: VSCodeUriResolverServer;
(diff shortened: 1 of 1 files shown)


DavyLandman:
We can also make these errors non-modal:
image

but not sure if that might be too subtle.

(you can expand it for more details:
image
)

jurgenvinju:
BTW, the "Takeout" project also has the problem of a capital `T` in its name. Do we check for this too?

DavyLandman:
> BTW, the "Takeout" project also has the problem of a capital `T` in its name. Do we check for this too?

our rascal.mf validator does. but that was missing at the moment :)

DavyLandman:
It's a trade-off, should we try and make a full (in rascal) checker and fixer of project setups, or just detect most frequent errors, and point to the right way to construct a project?

I'm all in favor of making a nicer experience for our users, but we don't have to do that all at once.

jurgenvinju:
> It's a trade-off, should we try and make a full (in rascal) checker and fixer of project setups, or just detect most frequent errors, and point to the right way to construct a project?

Definitely we should not try and make a full rascal project fixer and checker. There is no bandwidth for that. We should keep doing what you've been doing: making essential improvements.

DavyLandman:
@jurgenvinju agreed.

I've added a button so you can open the relevant documentation on our site, just to help users on the way to do it. and we can think about a proper way to fix the project setups later on.

PR #337, opened 2023-11-08
Fix cross project automatic reloading

### LINKED ISSUE #336: Module reloader does not work across project boundaries
**Describe the bug**

Cross project dependencies work, but the module reloading feature doesn't see updates for modules from a different project.

**To Reproduce**

project A has module A. project B has module B that imports A. If you open a repl for B, it works nicely. but if you change A, and run something in the REPL, the automatic module reloading is not triggered. only when you explicitly re-import A, does the REPL see the changes. 

**Expected behavior**
The module reloader picks up the changes of the other project. Since it can see the changes after doing a re-import. My hunch is that the file watch is giving path changes for a different URI than we expect.

 - Context: VS Code latest extension on Windows
 - Rascal Version: 0.34.1

DavyLandman:
This PR fixes #336

I think we should also see if we can fix the underlying bug reported in usethesource/rascal#1884. But this fix would make sure our users are not hit by this any longer.


jurgenvinju:
ok good workaround. when we fix it in rascal, this code still works as well. so we can clean up independently

jurgenvinju (approved):

Issue #343, opened 2023-12-12
Allow multiple file extensions for the same language


DavyLandman:
**Is your feature request related to a problem? Please describe.**
Right now, we don't allow multiple file extensions for the same DSL. Both the LSP and VS Code do not have this restriction. 

LSP doesn't talk about extensions, and even discusses how a single language server could host multiple languages (it says we should use the `languageId` property of a document, not re-inspect the extension of the file).

VS Code allows you to configure multiple extensions for a single language, the extension property is always a list in package.json.

Apart from the disparity to LSP and VS Code, we have a customer where the same language has multiple extensions. There is a work around, but that does case quite a few more interpreters to be loaded, and other features of VS Code are negatively impacted by splitting all those extensions into different languages.

**Describe the solution you'd like**
`registerLanguage`/`Language` should be changed, to provide a `set[str] extension` instead of a `str extension` Parameter. This will involve some invasive changes in rascal-lsp project, and a few small ones in rascal as well. I think we should make an overload/function constructor to support the old single extension version, that just constructs the set when needed.

**Describe alternatives you've considered**
Defining every different extension with a different language would involve loading parallel evaluators (increasing memory pressure and loading time quite a bit) and it also hurts other VS Code features, which are normally connected/shared to the language identifier.

jurgenvinju:
Good feature! Since this would change the basic API in a breaking way (changing a `str` to a `set[str]`), let's add deprecated functions that simulate the old way to ease the transition for our users. These functions would construct a singleton set for them.

jurgenvinju:
Let's do exactly as you wrote. I misread and repeated your sentences :-)

PR #344, opened 2023-12-12
Feature: multiple extensions for single language

DavyLandman:
As discussed in #343 we want to better align with LSP & VS Code by allowing multiple extensions for a single language.

The biggest change was in our multiplexer, which used to work by extension alone, but now maps extensions to language, and then continues as before.

I've tested that the old DSLs still work without requiring changes.


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/ITerminalIDEServer.java
```diff
@@ -482,9 +484,30 @@ public String getMainModule() {
 
         @Override
         public String toString() {
-            return "LanguageParameter(pathConfig=" + pathConfig + ", name=" + name + ", extension=" + extension
+            return "LanguageParameter(pathConfig=" + pathConfig + ", name=" + name + ", extensions=" + extensions
```
rodinaarssen, line 487: The `toString` of an array is probably not what you want here

rascal-vscode-extension/src/lsp/ParameterizedLanguageServer.ts
```diff
@@ -152,8 +152,8 @@ export interface LanguageParameter {
     pathConfig: string
     /** name of the language */
     name: string;
-    /** extension for files in this language */
-    extension:string;
```
urbanfly, line 156: Consider leaving the original singular property and supporting it for backwards compatibility.

DavyLandman: yup, I was also thinking about that. that would be nicer, as we did the same for the rascal consumption of the API.

DavyLandman: The problem is that the new definition would still be an compile time error for consumers, as they would not have defined `extensions` in their object. and I didn't want to make that one optional.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParserOnlyContribution.java
```diff
@@ -29,6 +29,10 @@
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.Arrays;
```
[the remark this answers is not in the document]
DavyLandman: True, I spotted these, I have a commit incomming that cleaned up these.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/ITerminalIDEServer.java
DavyLandman: tnx, fixed.

jurgenvinju (approved):
necessary and efficiently solved. Do our tests cover all the cases we've changed here? And do we have a test for the multiple-extension situation?

DavyLandman:
Thanks. well, I updated the pico example to use the new constructor. And the test fail, but that's just UI test jank, since only the first DSL test fails.

In general there is not difference between how a language with 1 extension or more is handled now.

DavyLandman:
>Otherwise, this feature will break the next time someone does something "quickly". Like me :-)

that seems wise, I added a test for it.

rascal-vscode-extension/src/lsp/ParameterizedLanguageServer.ts
jurgenvinju: Interesting alternatives! I understand this discussion, but are we sure we have to be backward compatible on this side? In principle, our own server side (in Java) is the only client of this interface, and supplier of this data (as far as I know). And the more nice tricks we do here in TypeScript, the more nasty this object will look in Java. The java JSON-RPC bridge does something "special" for union types, but that's as far as my compliments go ;-)

urbanfly: I don't know how much this change will affect consumers of [this package](https://www.npmjs.com/package/@usethesource/rascal-vscode-dsl-lsp-server). Maybe it is fine to break the few (?) consumers. If that's true, no problem. It has been "considered" and rejected 😄

DavyLandman: I think the concern is that this LanguageParameter is also used by deployed languages. 

I'm happy breaking that API a bit, since we are in 0.x, and the compile error would be very present in node, and only happen if you explicitly upgrade that dependency.

At the moment, I know all the places where this is happening.

DavyLandman: Ok, we are argee :)

jurgenvinju: nice to talk about this stuff @urbanfly ! thanks

PR #370, opened 2024-03-27
Upgrade to latest LSP client implementation and improve typescript configuration

### LINKED ISSUE #348: Switch to ESM & node 18 & newer version of VS Code
latest vscode-language-client (https://github.com/usethesource/rascal-language-servers/pull/321) and chai requires both ESM and node 18. 

Looks like this is something we have to tackle soon.

DavyLandman:
This PR does a few things that we've been delaying:

- finish upgrade to node 18 (more and more of our dependencies stopped working on node 16, and VS Code was now running node 18 as well)
- simplify our typescript config
- upgrade to latest lsp library
- we could drop node-fetch dependency as java 18 has fetch build in.
- switch to esbuild for both faster compilation and less configuration (webpack required quite some extra config)

As a side effect this has to increase our minimal VS Code version to 1.82.0 (which was released in June 2023). Since most users are frequently updating VS Code, this should be fine.

This fixes #348 , we could not switch to ESM, as VS Code does not support that, but we could switch to a newer module system, just so we can use the new lsp library.


rascal-vscode-extension/src/auto-jvm/downloaders.ts
```diff
@@ -26,7 +26,7 @@
  */
 import * as os from 'os';
 import * as path from 'path';
-import fetch, { Response } from 'node-fetch';
```
jurgenvinju, line 29: Remove commented code

rascal-vscode-extension/src/auto-jvm/downloaders.ts
```diff
@@ -176,12 +176,23 @@ export async function identifyLatestTemurinLTSRelease(version: number, arch: Tem
         throw new Error("Adoptium returned no releases");
     }
     const rel =releases.versions[0];
+    if (!rel) {
+        throw new Error("Missing versions in api result" + rel);
```
jurgenvinju, line 180: Cryptic error for user. Include talk about the jdk

jurgenvinju:
Thanks for this upgrade!

rascal-vscode-extension/src/auto-jvm/downloaders.ts
DavyLandman: oops :) well spotted.

rascal-vscode-extension/src/auto-jvm/downloaders.ts
DavyLandman: good point, I've updated it.

PR #388, opened 2024-06-03
Report errors to the user

### LINKED ISSUE #384: Report all exceptions to LSP via `ResponseErrorException` instead of generic exception, which don't reach VS Code with any detail.s
Most of exceptions/errors that rascals throws are hidden in the output window, but LSP has an dedicated `error` field for this. If we translate them to this, VS Code can decide how/if to report errors.

Add something so that most of our futures get a special handler to wrap exceptions with `ResponseErrorException`.

Leave as is, with the downside that users are less updated in case of something goes wrong.

**Additional context**

DavyLandman:
Initially we crashed frequently and we wanted to hide this from the user. But now we are much more stable, so we can show all errors to the user. Note that VS Code and the language client implementation report errors differently for some calls

Fixes #384

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
@@ -57,6 +57,7 @@ import org.rascalmpl.vscode.lsp.IBaseTextDocumentService;
 import org.rascalmpl.vscode.lsp.LSPIDEServices;
 import org.rascalmpl.vscode.lsp.rascal.RascalLanguageServer;
 import org.rascalmpl.vscode.lsp.util.concurrent.InterruptibleFuture;
+import io.usethesource.vallang.IConstructor;
 import io.usethesource.vallang.ISourceLocation;
 import io.usethesource.vallang.IString;
 import io.usethesource.vallang.IValue;
@@ -115,23 +116,36 @@ public class EvaluatorUtil {
     private static String formatMessage(Throw e) {
         String message = "";
         IValue thrown = e.getException();
-        if (thrown instanceof IString) {
-            message = ((IString)thrown).getValue();
+        if (thrown instanceof IConstructor) {
+            IConstructor exp = (IConstructor)thrown;
+            if (exp.has("message")) {
+                message = asString(exp.get("message"));
+            }
+            else {
+                message = exp.getName();
+            }
         }
         else {
-            LimitedResultWriter res = new LimitedResultWriter(512);
-            try {
-                new StandardTextWriter(false).write(thrown, res);
-            }
-            catch (/*IO Limit*/ RuntimeException | IOException _ignored) {
-            }
-            message = res.toString();
+            message = asString(thrown);
         }
         return message + "\nat: " + e.getLocation();
     }
 
+    private static String asString(IValue v) {
+        if (v instanceof IString) {
+            return ((IString)v).getValue();
+        }
+        LimitedResultWriter res = new LimitedResultWriter(512);
+        try {
+            new StandardTextWriter(false).write(v, res);
+        }
+        catch (/*IO Limit*/ RuntimeException | IOException _ignored) {
+        }
+        return res.toString();
+    }
+
     private static String formatMessage(StaticError e) {
-        return "Static error in rascal code: " + e.getMessage() + "\nat: " + e.getLocation();
+        return "Static error: " + e.getMessage();
     }
 
     public static CompletableFuture<Evaluator> makeFutureEvaluator(ExecutorService exec, IBaseTextDocumentService docService, BaseWorkspaceService workspaceService, IBaseLanguageClient client, String label, PathConfig pcfg, boolean addRascalCore, final String... imports) {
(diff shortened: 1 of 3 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
```diff
@@ -80,34 +85,23 @@ public static <T> InterruptibleFuture<T> runEvaluator(String task, CompletableFu
...
             }
             catch (Throw e) {
                 logger.error("Internal error during {}\n{}: {}\n{}", task, e.getLocation(), e.getMessage(),
                         e.getTrace());
-                // logger.error("Full internal error: ", e);
-                if (throwFailure) {
-                    throw e;
-                }
-                return defaultResult;
+                throw new ResponseErrorException(new ResponseError(ResponseErrorCode.RequestFailed, formatMessage((Throw)e), null));
```
toinehartman, line 96: Sonar reports an unnecessary cast here

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
```diff
@@ -80,34 +85,23 @@ public static <T> InterruptibleFuture<T> runEvaluator(String task, CompletableFu
...
             }
             catch (StaticError e) {
                 logger.error("Static Rascal error in {}\n{}: {}", task, e.getLocation(), e.getMessage());
-                // logger.error("Full internal error: ", e);
-                if (throwFailure) {
-                    throw e;
-                }
-                return defaultResult;
-            } 
+                throw new ResponseErrorException(new ResponseError(ResponseErrorCode.InternalError, formatMessage((StaticError)e), null));
```
toinehartman, line 100: Sonar reports an unnecessary cast here

toinehartman (approved):
I am not sure where propagating *all* errors might influence the end-user experience, but as a developer, LGTM! Two minor nitpicks that came up in the Sonar analysis.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
DavyLandman: Thanks, after refactoring I forgot to check these messages.

DavyLandman:
> I am not sure where propagating all errors might influence the end-user experience

True, my reasoning: it's time we switch it around, we don't expect errors now, so if they exist, they should become visible such that users would report it.

toinehartman:
> we don't expect errors now

I misse that you mentioned this already in the PR description. Makes total sense!

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
```diff
@@ -118,6 +112,28 @@ public static <T> InterruptibleFuture<T> runEvaluator(String task, CompletableFu
...
 
+    private static String formatMessage(Throw e) {
+        String message = "";
+        IValue thrown = e.getException();
+        if (thrown instanceof IString) {
+            message = ((IString)thrown).getValue();
+        }
+        else {
+            LimitedResultWriter res = new LimitedResultWriter(512);
+            try {
```
jurgenvinju, line 123: This could be improved. The way the standard exceptions work from the `Exception` module in the
Stdlib is that must constructors of RuntimeException have a str field named "message"; it not then the name of the constructor is the message. Other fields contribute to the cause analysis. 

Proposal to translate this knowledge to a more elaborate error message over simply printing the constructor, since the end-user is the target audience rather than the rascal programmer. I would ignore the other causes for now to keep it short. The full error constructor can always be logged.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
```diff
@@ -118,6 +112,28 @@ public static <T> InterruptibleFuture<T> runEvaluator(String task, CompletableFu
...
+            }
+            catch (/*IO Limit*/ RuntimeException | IOException _ignored) {
+            }
+            message = res.toString();
+        }
+        return message + "\nat: " + e.getLocation();
+    }
+
+    private static String formatMessage(StaticError e) {
+        return "Static error in rascal code: " + e.getMessage() + "\nat: " + e.getLocation();
```
jurgenvinju, line 134: Don't have to mention the programming language. Also isn't the location already reflected in the metadata? Probably we can elide the location here because it is accessible via the UI.

jurgenvinju (changes requested):
This is very nice! Small incision with great effect

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
DavyLandman: Maybe true for the Static Errors indeed.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
DavyLandman: good points, I've rewritten that code, would you review again @jurgenvinju ?

Issue #382, opened 2024-05-29
Offer rename refactoring of Rascal identifiers


DavyLandman:
**Is your feature request related to a problem? Please describe.**
I want to have rename refactoring work for rascal. 

When I as a user trigger rename on:

- function
- variable
- module name

I want rascal to calculate the correct rename operation based on whats currently open in the workspace.

**Describe the solution you'd like**
LSP has a dedicated api for [rename](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_rename). To pace our development, I would like to have support for rename first, and to the generic refactorings in a later stage. Similarly, if needed, we can split the support for rascal vs the support for DSLs.

Ideally the rename command is executed in a similar way as `outline` is done, we have rascal code that is invoked by the rascal-lsp server.

**Describe alternatives you've considered**
- write the refactoring in Java: would be a bad idea, will be quite verbose
- create a generic refactoring framework that models any refactoring of LSP to rascal, such that we can leverage that: too much work as an initial addition. First get rename right, use that implementation to learn how to do the rest.

**Additional context**
Good news is that we already have [TextEdits](https://www.rascal-mpl.org/docs/Library/analysis/diff/edits/TextEdits/) in Rascal, which map cleanly on the edits that LSP needs.

jurgenvinju:
Textedits also cover file renamings, so that's cool if we change a module name

jurgenvinju:
It would be best if we can reuse information generated by the type-checker; for example the summaries, collect them for all modules in a project and then reverse the references. 

I would limit the current functionality to within a current project and push cross project uses/renamings to an enhancement issue. We don't even have cross-project incremental type-checking

toinehartman:
Thanks for the directions!
I set up all the boilerplate today and was just looking into (re)using type checker information and collecting across all modules. @DavyLandman and I will discuss tomorrow.

toinehartman:
# Feature milestones

**Milestone 1 [boilerplate]**
- [x] Testing framework
- [x] Support for renaming local variables

**Milestone 2 [locals]**
- [x] Nested function names
- [x] Formal parameters
- [x] Pattern variables

**Milestone 3 [globals]**
- [ ] Global variables
- [ ] Functions (private)
- [ ] Functions (public)
- [ ] Aliases

**Milestone 4 [modules]**
- [ ] Module names

**Milestone 5 [grammars]**
- [ ] Symbols
- [ ] Productions
[...]

PR #389, opened 2024-06-10
Rename refactoring support for Rascal (locals)

toinehartman:
Add rename support for Rascal itself (but not for DSLs). Can be triggered in VS Code using F2 or right-click->rename.

The renaming feature is intended to be released incrementally. This PR includes milestones 1&2 as described in #382.


rascal-lsp/src/main/rascal/lang/rascal/lsp/Rename.rsc
```diff
@@ -0,0 +1,275 @@
...
+
+void checkCausesDoubleDeclaration(set[Define] currentDefs, set[Define] newDefs) {
+    // Is newName already resolvable from a scope where <current-name> is currently declared?
+    throwIfNotEmpty(CausesDoubleDeclaration, {<nD.defined, cD> | Define nD <- newDefs
+                                                               , loc cD <- currentDefs.defined
+                                                               , isContainedIn(cD, nD.scope)
+                                             });
+}
+
+bool isImplicitDefinition(start[Module] _, rel[loc use, loc def] useDef, Define _: <_, _, _, variableId(), defined, _>) = defined in useDef.use; // use of name at implicit declaration loc
```
DavyLandman, line 104: nit: I think instead of `Define _: <...>` we can just say  `<...>`

rascal-lsp/src/main/rascal/lang/rascal/lsp/Rename.rsc
```diff
@@ -0,0 +1,275 @@
...
+
+bool isImplicitDefinition(start[Module] m, rel[loc, loc] _, Define _: <_, _, _, patternVariableId(), defined, _>) {
+    visit (m) {
```
DavyLandman, line 107: I get what is happening here, but it feels like there should be a more rascal way? especially the if part of it. like, why it's not part of the pattern match in `qualifiedName` or something?

rascal-lsp/src/main/rascal/lang/rascal/lsp/Rename.rsc
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    return false;
+}
+
+default bool isImplicitDefinition(start[Module] _, rel[loc, loc] _, Define _) = false;
+
+void checkCausesCapture(TModel tm, start[Module] m, set[Define] currentDefs, set[loc] currentUses, set[Define] newDefs) {
+    set[Define] newNameImplicitDefs = {def | Define def <- newDefs, isImplicitDefinition(m, tm.useDef, def)};
+    set[Define] newNameExplicitDefs = {def | Define def <- newDefs, !isImplicitDefinition(m, tm.useDef, def)};
```
DavyLandman, line 123: nit: these two seem like set difference:

```
allDefs = {def | Define def <- newDefs };
newNameImplicitDefs = {d | d <- allDefs, isImplicitDefinition(m, tm.useDef, d)};
newNameExplicitDefs = allDefs - newNameImplicitDefs ;
```

and ideally (also faster):

```
newNameImplicitDefs = findImplicitDefs(m, tm.useDef); // only a single visit over the tree instead of a visit over the tree for every define
newNameExplicitDefs = {def | Define def <- newDefs } - newNameImplicitDefs;
```

rascal-lsp/src/main/rascal/lang/rascal/lsp/Rename.rsc
```diff
@@ -0,0 +1,275 @@
...
+                                      , Define cD <- getDefines(tm, cU)
+                                      , isContainedIn(cU, nD.scope)
+                                      , isStrictlyContainedIn(nD.scope, cD.scope)
+        };
+
+    throwIfNotEmpty(CausesCapture, implicitDeclBecomesUseOfCurrentDecl + explicitDeclShadowsCurrentDecl);
+}
+
+void checkEditableDefinitions(set[Define] currentDefs) {
+    bool isEditableLocation(loc l) = l.scheme in {"file", "cwd", "home", "memory", "project"};
```
DavyLandman, line 145: It's not possible to make an accurate list, and also, if a file is readonly on disk, we still cannot write to it. Let's let VS Code worry about updating files or not, as it's a whole mess in the virtual file system if something is readonly or not.

rascal-lsp/src/main/rascal/lang/rascal/lsp/Rename.rsc
```diff
@@ -0,0 +1,275 @@
...
+}
+
+void checkLegalRename(TModel tm, start[Module] m, loc cursorLoc, set[Define] currentDefs, set[loc] currentUses, str newName) {
+    set[Define] newNameDefs = {def | def <- tm.defines, def.id == newName};
+    try {
+        checkEditableDefinitions(currentDefs);
+        checkLegalName(newName);
+        checkCausesDoubleDeclaration(currentDefs, newNameDefs);
+        checkCausesCapture(tm, m, currentDefs, currentUses, newNameDefs);
+    } catch IllegalRenameReason r: {
```
DavyLandman, line 156: I'm not sure I truly like this style of using exceptions for "regular" control flow. I get the beautify, but there are some constrains to this style:

1. it only allows for a single message
2. we have to be careful about other exceptions.

if every checker would return a set of `IllegalRenameReason` we can plus all the calls together, and if the result is non-empty, throw it is an illegal rename. It might also remove the need for `throwIfNotEmpty` as we can use the comprehension directly (with a small change of the signature of the `IllegalRenameReason` ADT I guess).

rascal-lsp/src/main/rascal/lang/rascal/lsp/Rename.rsc
```diff
@@ -0,0 +1,275 @@
...
+    }
+    checkUnsupported(m.src, currentDefs);
+}
+
+void checkUnsupported(loc moduleLoc, set[Define] defsToRename) {
+    Maybe[str] isUnsupportedDefinition(Define _: <scope, id, _, functionId(), _, _>) = just("Global function definitions might span multiple modules; unsupported for now.")
+        when moduleLoc == scope; // function is defined in module scope
+    default Maybe[str] isUnsupportedDefinition(Define _) = nothing();
+
+    throwIfNotEmpty(UnsupportedRename, {<def.defined, msg> | def <- defsToRename, just(msg) := isUnsupportedDefinition(def)});
```
DavyLandman, line 167: if we switch to returning a set, we can get rid of quite some maybe's as well.

rascal-lsp/src/main/rascal/lang/rascal/lsp/Rename.rsc
```diff
@@ -0,0 +1,275 @@
...
+    } catch IllegalRenameReason r: {
+        throw IllegalRename(cursorLoc, r);
+    }
+    checkUnsupported(m.src, currentDefs);
+}
+
+void checkUnsupported(loc moduleLoc, set[Define] defsToRename) {
+    Maybe[str] isUnsupportedDefinition(Define _: <scope, id, _, functionId(), _, _>) = just("Global function definitions might span multiple modules; unsupported for now.")
+        when moduleLoc == scope; // function is defined in module scope
+    default Maybe[str] isUnsupportedDefinition(Define _) = nothing();
```
DavyLandman, line 165: wouldn't it be easier to mark what we do support? instead of all the cases we don't? Like I'm not seeing module names or grammar names for example.

rascal-lsp/src/main/rascal/lang/rascal/tests/Rename.rsc
```diff
@@ -0,0 +1,410 @@
...
+ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+}
+module lang::rascal::tests::Rename
```
DavyLandman, line 27: 🎉 nice that we have tests! 🎉

rascal-vscode-extension/test-workspace/test-project/src/main/rascal/RenameExamples.rsc
```diff
@@ -0,0 +1,74 @@
+module RenameExamples
```
DavyLandman, line 1: do we still need this file as well? and if so, add a small comment on top of why.

rascal-vscode-extension/test-workspace/test-project/src/main/rascal/SingleModuleRenameTest.rsc
```diff
@@ -0,0 +1,31 @@
+module SingleModuleRenameTest
```
DavyLandman, line 1: similar for this file, do we need it? or can we move it to the `rascal-lsp/test` directory?

DavyLandman (changes requested):
All in all, I'm quite happy, looks good, this has been a significant amount of work. I think my main question is regarding the exceptions for the errors.

And it might be nice to have 1 integration test added to the UI test to make sure the rename also works (and keeps working).

rascal-lsp/src/main/rascal/lang/rascal/lsp/Rename.rsc
toinehartman: This is not purely a visit over the tree. In case of pattern variables, this syntactical analysis is needed, but for regular variables, it is not. However, I understand the point and can rewrite to something that visits the tree at most once.

rascal-lsp/src/main/rascal/lang/rascal/lsp/Rename.rsc
toinehartman: I am aware of that, but was trying to prevent messing up the arity or types of the tuples here.

DavyLandman: Fair, it's tricky sometimes.

rascal-lsp/src/main/rascal/lang/rascal/lsp/Rename.rsc
DavyLandman: Yeah, I like the thought, but in the end, I don't think we can get this water tight, without adding a `isWritable(loc)` function to `IO.rsc` (not even sure if we can do that properly)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
```diff
@@ -114,11 +120,13 @@ public class RascalTextDocumentService implements IBaseTextDocumentService, Lang
     private final ColumnMaps columns;
     private @MonotonicNonNull FileFacts facts;
     private @MonotonicNonNull BaseWorkspaceService workspaceService;
+    private final DocumentChanges docChanges;
 
     public RascalTextDocumentService(ExecutorService exec) {
         this.ownExecuter = exec;
         this.documents = new ConcurrentHashMap<>();
         this.columns = new ColumnMaps(this::getContents);
+        this.docChanges = new DocumentChanges(this);
```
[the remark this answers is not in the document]
toinehartman: That has been on my list for a while. Now done in f141a96f.

rascal-lsp/pom.xml
```diff
@@ -160,6 +174,13 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <version>${maven-surefire-plugin.version}</version>
+        <configuration>
```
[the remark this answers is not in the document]
toinehartman: Turns out not to be necessary. Removed.

rascal-lsp/src/main/rascal/lang/rascal/lsp/Rename.rsc
toinehartman: In geneal, `findDeclarationName` will take care of what we do and don't support. In this case however, we need some more analysis to see if we really support this definition.

toinehartman: Done!

toinehartman:
> All in all, I'm quite happy, looks good, this has been a significant amount of work. I think my main question is regarding the exceptions for the errors.
> 
> And it might be nice to have 1 integration test added to the UI test to make sure the rename also works (and keeps working).

Thanks. Processed all your remarks and added an integration test. Can you please have another look?

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
```diff
@@ -152,5 +153,21 @@ describe('IDE', function () {
         await driver.actions().doubleClick(mainItem!).perform();
         await driver.wait(async ()=> (await editor.getCoordinates())[0] === 5, Delays.normal, "Cursor should have moved to line that contains the println function");
     });
+
+    it ("rename works", async() => {
+        const editor = await ide.openModule(TestWorkspace.libFile);
+        await editor.moveCursor(7, 15);
+
+        await bench.executeCommand("Rename Symbol");
+        const renameBox = await driver.wait(async () => await bench.getEditorView().findElement(By.className("rename-input")), Delays.fast, "Rename box should appear");
```
DavyLandman, line 162: maybe we can do something that reuses the same functions that are present in `util.ts`?

DavyLandman (approved):
Looks good to me. We'll have to wait for the checker integration to finish, and then we can finish this PR.

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
toinehartman: Had a look, but I don't see anything related to finding arbitrary editor elements there. I can refactor once we write more related tests (e.g. cross-module and/or other refactorings).

rascal-lsp/src/main/rascal/lang/rascal/lsp/Rename.rsc
```diff
@@ -0,0 +1,272 @@
...
+    // Will this rename turn an implicit declaration of `newName` into a use of a current declaration?
+    rel[loc use, loc oldDef, loc newDef] implicitDeclBecomesUseOfCurrentDecl =
+        {<nD.defined, nD.defined, cD> | Define nD <- newNameImplicitDefs
+                                      , <loc cS, loc cD> <- currentDefs<scope, defined>
+                                      , isContainedIn(nD.defined, cS)
+        };
+
+    // Will this rename turn a use of <oldName> into a use of an existing declaration of <newName> (shadowing)?
+    rel[loc use, loc oldDef, loc newDef] explicitDeclShadowsCurrentDecl =
+        {<cU, cD.defined, nD.defined> | Define nD <- newNameExplicitDefs
```
jurgenvinju, line 128: if this is a efficiency bottleneck then rewriting a bit with the `o` composition operator may take off the edge. It typically avoids one more iteration through the target set, so takes off one complexity order. I think there are two hidden `o` operators here before two filters.

jurgenvinju (approved):
very nice. can't wait to see it in action :-)

rascal-lsp/src/main/rascal/lang/rascal/lsp/Rename.rsc
```diff
@@ -0,0 +1,275 @@
...
+    | DeclarationsNotEditable(set[loc] decls)
+    ;
+
+data RuntimeException
+    = IllegalRename(loc location, IllegalRenameReason reason)
+    | UnsupportedRename(set[tuple[loc location, str message]])
+    | UnexpectedFailure(str message)
+;
+
+private set[loc] getUses(TModel tm, loc def) {
```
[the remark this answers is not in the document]
toinehartman: Yeah, they are all at the bottom (and marked with comments). I am not sure if these will be exposed in the new typechecker. Will have a look once we get there.

rascal-lsp/src/main/rascal/lang/rascal/lsp/Rename.rsc
toinehartman: Fixed.

rascal-lsp/src/main/rascal/lang/rascal/lsp/Rename.rsc
toinehartman: I will think about that. Performance has not been my primary concern and since this should only consider local variables (otherwise an exception will be thrown before this code is even called), I am not too worried right now.

toinehartman: Actually went ahead and rewrote this. Much better, thanks!

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
DavyLandman: I meant that for example the `hasElement` function looks to do exactly what this line is doing.

toinehartman: Missed that completely. Fixed it.

toinehartman:
I pushed some more changes, addressing some remarks by @PaulKlint outside of this PR. Especially 5fe2162 should highly reduce the number of parse tree visits in real scenarios. Now, a module tree will be visited at most once.

rascal-lsp/src/main/rascal/lang/rascal/lsp/Rename.rsc
toinehartman: Different approach (not for this PR, but for cross-module features that we're working on next): reject the renaming if any related definition is outside of the current workspace?

toinehartman:
TODO:

- [ ] Keyword parameters (local functions only) (ht @sungshik)

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
```diff
@@ -155,5 +156,21 @@ describe('IDE', function () {
...
         await driver.wait(async ()=> (await editor.getCoordinates())[0] === 5, Delays.normal, "Cursor should have moved to line that contains the println function");
     });
+
+    it ("rename works", async() => {
+        const editor = await ide.openModule(TestWorkspace.libFile);
+        await editor.moveCursor(7, 15);
+
+        await bench.executeCommand("Rename Symbol");
+        const renameBox = await ide.hasElement(editor, By.className("rename-input"), Delays.fast, "Rename box should appear");
+        await renameBox.sendKeys(Key.BACK_SPACE, "i", Key.ENTER);
```
DavyLandman, line 166: can we not use `setText` on the input box?

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
```diff
@@ -155,5 +156,21 @@ describe('IDE', function () {
...
+        await driver.wait(async () => (await editor.isDirty()), Delays.extremelySlow, "Rename should have resulted in changes in the editor");
```
DavyLandman, line 167: nit : `async () => (await editor.isDirty()` => `() => editor.isDirty()`

DavyLandman (approved):
Looks good, 2 questions on the test

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
toinehartman: Fixed

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
toinehartman: As discussed in private, that's not straightforward here. Resolving.

PR #394, opened 2024-06-14
New rascal-core & rascal

DavyLandman:
This brings:

- new typechecker
- various bugfixes in rascal
- the new repl progress bar

to Rascal.

It's still in Draft as stuff doesn't work yet, I just got it to compile.


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/TerminalIDEServer.java
```diff
@@ -224,7 +226,7 @@ public void warning(WarningMessage param) {
 
     @Override
     public void registerLocations(RegisterLocationsParameters param) {
-        new BasicIDEServices(null).registerLocations(param.getScheme(), param.getAuthority(), param.getMapping());
+        new BasicIDEServices(new PrintWriter(PrintWriter.nullWriter()), new NullRascalMonitor()).registerLocations(param.getScheme(), param.getAuthority(), param.getMapping());
```
jurgenvinju, line 229: Perhaps we can just inline the default method here, because it skips all the monitor variables anyway:

```
default void registerLocations(IString scheme, IString auth, IMap map) {
	  URIResolverRegistry.getInstance().registerLogical(new LogicalMapResolver(scheme.getValue(), auth.getValue(), map));
	}

  default void unregisterLocations(IString scheme, IString auth) {
    URIResolverRegistry.getInstance().unregisterLogical(scheme.getValue(), auth.getValue());
	}
```

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
```diff
@@ -148,17 +149,16 @@ private static String formatMessage(StaticError e) {
...
-    public static CompletableFuture<Evaluator> makeFutureEvaluator(ExecutorService exec, IBaseTextDocumentService docService, BaseWorkspaceService workspaceService, IBaseLanguageClient client, String label, PathConfig pcfg, boolean addRascalCore, final String... imports) {
+    public static CompletableFuture<Evaluator> makeFutureEvaluator(ExecutorService exec, IBaseTextDocumentService docService, BaseWorkspaceService workspaceService, IBaseLanguageClient client, String label, IRascalMonitor monitor, PathConfig pcfg, boolean addRascalCore, final String... imports) {
         return CompletableFuture.supplyAsync(() -> {
             Logger customLog = LogManager.getLogger("Evaluator: " + label);
-            IRascalMonitor monitor = new LSPIDEServices(client, docService, workspaceService, customLog);
+            IDEServices services = new LSPIDEServices(client, docService, workspaceService, customLog, monitor);
             boolean jobSuccess = false;
             try {
-                monitor.jobStart("Loading " + label);
+                services.jobStart("Loading " + label);
```
jurgenvinju, line 158: This "Loading " job does not have steps. Better remove it; it will make more sense if we leave it to the internal evaluators "loading modules" job.

jurgenvinju: There is a `doImport` call in this method that can now take the entire array of `imports` as a parameter. So the loop can be dropped and the internal monitor can do each module as a `jobStep`

jurgenvinju (changes requested):
* we can drop the empty "Loading ... " job (see comment)
* use `doImport(string array or ...param)` instead of looping ourselves for better monitors.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
DavyLandman: I'm going to add job steps, as the reason why this monitor is the top-level one is that it's part of the title of this progress bar. So I want a pretty one, and not one that says which module it's loading (you can still see that in the details.)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/TerminalIDEServer.java
DavyLandman: that would cause a clone of that code, I would rather not.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
DavyLandman: Reason's I'm not doing that `doImport` call:

1. the current code allows me to print a more accurate error message of which root module went wrong
2. the `doImport` with the varargs is not in a release rascal jar yet, so we can always migrate to that later.

jurgenvinju: Just know that every doImport call will now produce its own progress bar; or we go back to the steps without starts situation.

jurgenvinju: Note that the exception will tell you which module went wrong anyhow.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/TerminalIDEServer.java
DavyLandman: ok, refactored that code a bit.

DavyLandman:
@jurgenvinju I've address all your comments. can you approve the PR?

jurgenvinju (approved):

PR #401, opened 2024-07-02
Trying to make sure java classes can be used in VS Code projects

DavyLandman:
Right now, functions with `@javaClass` annotations do not work, I think this should be enough.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
@@ -51,6 +51,7 @@ import org.rascalmpl.interpreter.utils.LimitedResultWriter;
 import org.rascalmpl.library.util.PathConfig;
 import org.rascalmpl.shell.ShellEvaluatorFactory;
 import org.rascalmpl.uri.URIUtil;
+import org.rascalmpl.uri.classloaders.SourceLocationClassLoader;
 import org.rascalmpl.vscode.lsp.BaseWorkspaceService;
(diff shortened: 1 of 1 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
```diff
@@ -163,12 +164,16 @@ public static CompletableFuture<Evaluator> makeFutureEvaluator(ExecutorService e
                 eval.getConfiguration().setRascalJavaClassPathProperty(System.getProperty("rascal.compilerClasspath"));
                 eval.addClassLoader(RascalLanguageServer.class.getClassLoader());
                 eval.addClassLoader(IValue.class.getClassLoader());
+                ClassLoader cl = new SourceLocationClassLoader(pcfg.getClassloaders(), ShellEvaluatorFactory.class.getClassLoader());
```
jurgenvinju, line 167: This requires an `append(pcfg.getBin())` to make sure classes in the current target folder are reachable as well.

jurgenvinju: So `pcfg.getClassLoaders().append(pcfg.getBin())`  I believe. 

On the lib:///'less branchof rascal this will become `pcfg.getLibsAndTarget()`

jurgenvinju (changes requested):
good catch! Just add the target to be sure.

DavyLandman:
Looking at some more code, this is also needed in the REPL, especially if you take into account the multiple project setup. where we have to also have the target folder of other projects in our classpath.

jurgenvinju (approved):

Issue #342, opened 2023-12-06
Make it easier to call rascal LSP commands by helping the user translate to rascal call


DavyLandman:
**Is your feature request related to a problem? Please describe.**
currently a user of a lsp command, has to (when not invoked as part of a code-lens) call `vscode.commands.executeCommand`, with `rascal-meta-command`  as target (in deployment mode the language name is suffixed to it), and then the next argument is a full rascal value in a javascript string.

The user has to take care to translate JS types to rascal literals. This can be error prone (quoting of strings/escaping of nested quotes & back slashes). 

**Describe the solution you'd like**

If we were to extend the command handler with a second mode that would take JS values and translates them as good as we can to a rascal value.

```ts
// original mode:
vscode.commands.executeCommand("rascal-meta-command", "calcX("A string \\\" with nested \\\\ stuff gets messy", |file:///x.txt|)");

// second mode (only triggered if more than 1 argument:
vscode.commands.executeCommand("rascal-meta-command", "calcX", "A string \" with nested \\ stuff gets messy", vscode.Uri.file("/x.txt"));
```

Which would:

- translate strings/bool/numbers to rascal literals
- translate vscode.Uri / Ranges / Positions to source locations (with correct translation of offsets/columns)
- translate arrays to lists, sets to sets
- the rest of the objects to JSON representation

urbanfly:
The 'original mode' example demonstrates the difficulty with escaping strings correctly - it is not escaping the quotes that surround the inner string.

```ts
// original mode:
vscode.commands.executeCommand("rascal-meta-command", "calcX(\"A string \\\" with nested \\\\ stuff gets messy\", |file:///x.txt|)");
```

PR #402, opened 2024-07-02
code actions for dsls

jurgenvinju:
This PR connects the `codeActions` LSP server service for the parametrized LSP server (so for DLSs).

* A `CodeActionContributor` is of type `list[Command] (list[Tree] focus)` and registered with `data LanguageService = codeActionContributor(CodeActionContributor actions)`. The "focus" tells us which parse trees are active on the given line and column of the current cursor, ordered with the most specific (smallest) trees first and ending always with the `start` tree.
 
Pico example, where we are looking for a cursor position over an `IdType` somewhere in a `start[Program]`, and we register the command `removeDecl` which can be interpreter later:
   
```rascal
list[Command] picoActions([*_, IdType x, *_, start[Program] program]) 
    = [removeDecl(program, x, title="remove <x>")];
```

* Normal `Command`s are used just like the lenses. They need to be evaluated by the commandExecutor contribution.
* Additionally, every `Message` can now have a keyword parameter: `data Message(list[Command] fixes=[])`. Every message that is registred with the LSP client (e.g. via a Summary), will lead to a contributed code action when the cursor  is at the `at` location of the info, warning or error message.
* `demo::lang::pico::LanguageServer` demonstrates these two different ways of contributing actions (see above). Below the Message alternative:

```rascal
s.messages += {<src, error("<id> is not defined", src, 
                                               fixes=[changeToFix(src, existing<0>, title="Change to <existing<0>>") | existing <- defs])
                             > 
                            | <src, id> <- uses, id notin defs<0>};
```

The `error` constructor now receives an additional `fixes` keyword field with a list of options on how to fix the broken variable reference. Each  generated command fixes the bug in its own way (with a different variable name to use).


jurgenvinju:
Will post some screenshots later. This is still a draft.

jurgenvinju:
Looking for suggestions on how to improve the asynchronous code and suggestions how to test the UI part of this. It's most LSP integration code. The algorithm for generating focus lists could be tested as a unit.

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
```diff
@@ -81,31 +104,73 @@ Summary picoSummarizer(loc l, start[Program] input, PicoSummarizerMode mode) {
...
+  | removeDecl(start[Program] program, IdType toBeRemoved)
+  ;
 
-rel[loc,Command] picoLenses(start[Program] input) = {<input@\loc, renameAtoB(input, title="Rename variables a to b.")>};
+@synopsis{Adds an example lense to the entire program.}
+rel[loc,Command] picoLenses(start[Program] input) 
+    = {<input@\loc, renameAtoB(input, title="Rename variables a to b.")>};
 
 
+@synopsis{Generates inlay hints that explain the type of each variable usage.}
```
DavyLandman, line 128: while I appreciate these synopsis, the amount of changes in this file makes it harder to review the actual changes versus the comment & layout changes. I would appreciate it to separate that into a different PR next time, where possible.

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
```diff
@@ -81,31 +104,73 @@ Summary picoSummarizer(loc l, start[Program] input, PicoSummarizerMode mode) {
...
 
+@synopsis{Command handler for the renameAtoB command}
 value picoCommands(renameAtoB(start[Program] input)) {
     applyDocumentsEdits(getAtoBEdits(input));
     return ("result": true);
 }
 
+@synopsis{Command handler for the changeToFix command}
+value picoCommands(changeToFix(loc src, str newName)) {
+    applyDocumentsEdits([changed(src.top, [replace(src, newName)])]);
```
DavyLandman, line 151: this does not seem compatible with VS Code & LSP, where we shouldn't be changing the files on disk, but should let LSP & VS Code handle the execution of the edits.

I know we allow if for lenses, but I don't like that now this is the default behavior, it does not align with LSP.

rascal-vscode-extension/.vscode/launch.json
```diff
@@ -49,6 +56,30 @@
 				"${workspaceFolder}/out/test/**/*.js"
 			],
 			"preLaunchTask": "npm: pretest"
-		}
+		},
+		{
```
DavyLandman, line 60: do we need to copy this over to this launch.json? it's already in rascal-lsp one? won't this cause duplicate entries in the list of launch configureations?

DavyLandman (changes requested):
I like the work and in a lot of places a good design. But I think we have to change how the edits are applied.

The LSP specification says:

> The code action request is sent from the client to the server to compute commands for a given text document and range. These commands are typically code fixes to either fix problems or to beautify/refactor code. The result of a [textDocument/codeAction](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction) request is an array of [Command](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#command) literals which are typically presented in the user interface. To ensure that a server is useful in many clients the commands specified in a code actions should be handled by the server and not by the client (see [workspace/executeCommand](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_executeCommand) and ServerCapabilities.executeCommandProvider). If the client supports providing edits with a code action then that mode should be used.

Not responding to a CodeAction with a list of changes gives a sub-optimal user experience, especially in relation to unsaved changes.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -500,6 +515,64 @@ public CompletableFuture<SemanticTokens> semanticTokensRange(SemanticTokensRange
...
+            // and which came back with the codeAction's list of relevant (in scope) diagnostics:
+            Stream<IValue> quickfixes = params.getContext().getDiagnostics()
+                .stream()
+                .map(Diagnostic::getData)
+                .filter(Objects::nonNull)
+                .filter(JsonPrimitive.class::isInstance)
+                .map(JsonPrimitive.class::cast)
+                .map(JsonPrimitive::getAsString)
+                .map(contribs::parseCommands)
+                .map(CompletableFuture::join)
```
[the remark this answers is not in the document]
sungshik: Maybe [`allOf`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#allOf-java.util.concurrent.CompletableFuture...-) can be used here?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
```diff
@@ -95,15 +95,33 @@ public static Diagnostic translateDiagnostic(IConstructor d, ColumnMaps cm) {
...
         result.setMessage(((IString) d.get("msg")).getValue());
-        result.setRange(Locations.toRange(getMessageLocation(d), cm));
+        Range range = Locations.toRange(getMessageLocation(d), cm);
+        result.setRange(range);
+
+        storeFixCommands(d, result, range);
         return result;
     }
 
+    private static void storeFixCommands(IConstructor d, Diagnostic result, Range range) {
```
sungshik, line 105: Nit: Maybe swap the order of this function and the one below (to keep the two versions of `translateDiagnostic` together)

sungshik: Actually, it seems these two versions of `translateDiagnostics` are fully identical clones, except for the type of `cm`. I realize it's not the point of this PR, but as the PR makes these clones bigger, I feel at least a bit justified to complain about it 😉

Concrete suggestion:

```java
public static Diagnostic translateDiagnostic(IConstructor d, ColumnMaps cm) {
  return translateDiagnostic(d, Locations.toRange(getMessageLocation(d), cm));
}

public static Diagnostic translateDiagnostic(IConstructor d, LineColumnOffsetMap cm) {
  return translateDiagnostic(d, Locations.toRange(getMessageLocation(d), cm));
}

private static Diagnostic translateDiagnostic(IConstructor d, Range r) {
  ... // content of current translateDiagnostic
}
```

(There might be better/conciser alternatives.)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/TreeSearch.java
```diff
@@ -0,0 +1,124 @@
...
+			}   
+            else {
+                // stop and return failure
+                return false;
+            }
+		}
+
+		if (TreeAdapter.isAmb(tree)) {
+            // pick any tree to make the best of it.
+			return computeFocusList(focus, (ITree) tree.getAlternatives().iterator().next(), line, column);
```
sungshik, line 93: If the tree has no children (maybe it's guaranteed not to be?), an exception is thrown here

rascal-lsp/src/main/rascal/demo/lang/pico/examples/ite.pico
```diff
@@ -0,0 +1,13 @@
...
+  input : natural,  
+  output : natural;
+  
+   input := 0;
+   output := 1;
+   if input then
+     output := 1
+   else
+     output := 2
+   fi
```
sungshik, line 12: Nit: Indentation is different vs. lines 3-4 (and also vs. `fac.pico`)

sungshik (commented):
Nice feature 🙂 Just a few minor comments from my side.

jurgenvinju:
@DavyLandman this merge commit was not totally transparant to me: 1e1ad3a7a754be1e4d15b5d7cea286e25c198421 
   * finicky details with the monitor and a list of imports
   * possible other things that are going on that you may have in active memory.

jurgenvinju:
@DavyLandman furthermore the concept of "CodeAction" is now mirrored from the LSP into util::LanguageServer, such that:
* code actions have optionally `edits` that are always executed first, 
* next tot their (optional) `command`. 
* Also UI details via the `CodeActionKind` are now mapped which gives interesting effects (prioritization, funny icons, previews, yes/no pop-ups, fix-all).

rascal-lsp/src/main/rascal/demo/lang/pico/examples/ite.pico
jurgenvinju: :-) fixed. these examples were taken from flybytes where they were taken from the asf-library project in the meta-environment.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
jurgenvinju: nice! done.

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
jurgenvinju: ah. because this new synopsis is not related to the current PR. I see. thanks

rascal-vscode-extension/.vscode/launch.json
jurgenvinju: I wanted to create a single button launch for the entire second level. That works sometimes, and sometimes it does not. I don't understand the difference.

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
```diff
@@ -81,31 +104,73 @@ Summary picoSummarizer(loc l, start[Program] input, PicoSummarizerMode mode) {
     return s;
 }
 
+@synopsis{Finds a declaration that the cursor is on and proposes to remove it.}
+list[Command] picoActions([*_, IdType x, *_, start[Program] program])
```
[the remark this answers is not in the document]
jurgenvinju: I have `list[Tree]` so we probably agree on this :-)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -500,6 +515,64 @@ public CompletableFuture<SemanticTokens> semanticTokensRange(SemanticTokensRange
...
+                    Stream.concat(quickfixes, actions)
+                    .map(IConstructor.class::cast)
+                    .map(cons -> constructorToCommand(contribs.getName(), cons))
+                    .map(cmd  -> Either.<Command,CodeAction>forLeft(cmd))
+                    .collect(Collectors.toList())
+                );
+        }
+        else {
+            logger.warn("ignoring command execution (no code actions configured for this document): {}", params.getTextDocument());
+            return CompletableFuture.completedFuture(null);
```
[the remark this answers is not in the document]
jurgenvinju: Mmmm yes indeed. 

Please explain a bit more. Why is it and _error_ not to have any code actions right now? Next time the engineer registers their new contributions, there might be. Another time when they remove those contributions again not. It's a normal situation during language engineering right? That's why I selected `warn` and returned a default.

I changed `null` to `Collections.emptyList()` accordingly. Much better default.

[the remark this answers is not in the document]
jurgenvinju: Ah yes. contributions can never be null. Right!

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
```diff
@@ -180,9 +180,9 @@ public static CompletableFuture<Evaluator> makeFutureEvaluator(ExecutorService e
                         eval.doImport(eval, i);
                     } catch (Exception e) {
                         logger.catching(e);
-                        logger.error("Failure to import, RascalResolver: {}", eval.getRascalResolver());
-                        throw new RuntimeException("Failure to import required module " + i, e);
-                    }
+                        logger.error("Failure to import, due to {}, RascalResolver: {}", e.getMessage(), eval.getRascalResolver());
+                        throw new RuntimeException("Failure to import required module " + i + " due to:" + e.getMessage(), e);
```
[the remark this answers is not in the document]
jurgenvinju: the merge conflict was big but easy to fix :-)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/TreeSearch.java
```diff
@@ -0,0 +1,124 @@
...
+
+import io.usethesource.vallang.IList;
+import io.usethesource.vallang.IListWriter;
+import io.usethesource.vallang.ISourceLocation;
+import io.usethesource.vallang.IValue;
+
+/**
+ * Utilities for finding sub-trees based on UTF-24 line/column indexing.
+ */
+public class TreeSearch {
```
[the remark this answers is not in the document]
jurgenvinju: Yes it prevents the need for some kind of reverse lookup (with or without cache) based on the assumption that a recent tree is present.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/TreeSearch.java
```diff
@@ -0,0 +1,124 @@
...
+    public static IList computeFocusList(ITree tree, int line, int column) {
+        var lw = IRascalValueFactory.getInstance().listWriter();
+        computeFocusList(lw, tree, line, column);
+        return lw.done();
+    }
+
+    private static boolean computeFocusList(IListWriter focus, ITree tree, int line, int column) {
+		ISourceLocation l = TreeAdapter.getLocation(tree);
+
+		if (l == null) {
```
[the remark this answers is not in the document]
jurgenvinju: it's ok for me now. Must be fixed somehow.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -500,6 +515,64 @@ public CompletableFuture<SemanticTokens> semanticTokensRange(SemanticTokensRange
...
+        final ILanguageContributions contribs = contributions(params.getTextDocument());
+        final var loc = Locations.toLoc(params.getTextDocument());
+        final var start = params.getRange().getStart();
+        final var startLine = start.getLine() + 1;
+        final var startColumn = columns.get(loc).translateInverseColumn(startLine, start.getCharacter(), false);
+        
+        if (contribs != null) {
+            // first we filter out the "fixes" that were optionally sent along with earlier diagnostics
+            // and which came back with the codeAction's list of relevant (in scope) diagnostics:
+            Stream<IValue> quickfixes = params.getContext().getDiagnostics()
```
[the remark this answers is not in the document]
jurgenvinju: done. I thought btw that the streams were already lazy, so all computation effectively was already on the back ground tasks that were iterating through the streams. The code outside the background task was only building the streams. But maybe I was misunderstanding this.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/InterpretedLanguageContributions.java
```diff
@@ -278,13 +284,26 @@ private static ISet loadContributions(Evaluator eval, LanguageParameter lang) {
         return (ISet) eval.eval(eval.getMonitor(), lang.getMainFunction() + "()", URIUtil.rootLocation("lsp"))
             .getValue();
     }
+ 
+    @Override
+    public CompletableFuture<IList> parseCommands(String command) {
```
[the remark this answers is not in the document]
jurgenvinju: I like this duplication because it avoids java generics. better have it clear which non-terminal is used. Also in the new code a `list[CodeAction]` is used so it makes the functions more different.

rascal-lsp/RELEASE-NOTES
```diff
@@ -0,0 +1 @@
+../rascal-vscode-extension/README.md
```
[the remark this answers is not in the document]
jurgenvinju: merged!

build.sh
```diff
@@ -7,7 +7,7 @@ extra_flags=''
 clean="clean"
 while getopts 'fd' flag; do
   case "${flag}" in
-    f) extra_flags='-Drascal.compile.skip' ;;
+    f) extra_flags='-Drascal.compile.skip -Drascal.tutor.skip' ;;
```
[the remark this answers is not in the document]
jurgenvinju: merged

jurgenvinju:
> I like the work and in a lot of places a good design. But I think we have to change how the edits are applied.
> 
> The LSP specification says:
> 
> > The code action request is sent from the client to the server to compute commands for a given text document and range. These commands are typically code fixes to either fix problems or to beautify/refactor code. The result of a [textDocument/codeAction](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction) request is an array of [Command](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#command) literals which are typically presented in the user interface. To ensure that a server is useful in many clients the commands specified in a code actions should be handled by the server and not by the client (see [workspace/executeCommand](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_executeCommand) and ServerCapabilities.executeCommandProvider). If the client supports providing edits with a code action then that mode should be used.
> 
> Not responding to a CodeAction with a list of changes gives a sub-optimal user experience, especially in relation to unsaved changes.

See above, I have taken your suggestion and mapped the entire CodeAction protocol.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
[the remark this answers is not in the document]
DavyLandman: Streams  are lazy, but their execution is done at collect time, so if the collect is not explicitly scheduled on the background, it will happen in the foreground. So in this specific case, it looks like all the calls scheduled here would indeed be executed inside the completable futre.

In general the pattern in this class is: do as much as you can inside of a CompletableFuture, just so that 

1. exceptions are properly translated to the server responds
2. you don't have to look at all your down stream calls if they might do a complex calculation that would block the server handler longer than needed.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/TreeSearch.java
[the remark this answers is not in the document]
DavyLandman: image
github still shows it as garbled. maybe enable `"editor.renderWhitespace": "all", ` temporarly to show the tabs & spaces mixed.

DavyLandman:
Thanks for taking the time, I'll allocate some time soon to review, both the tricky merge and new changes.

jurgenvinju:
I'm still fixing the merge. Even though I did an automatic merge lots of changes from `main` are not on the branch yet.

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
```diff
@@ -81,31 +104,73 @@ Summary picoSummarizer(loc l, start[Program] input, PicoSummarizerMode mode) {
...
 
+@synopsis{Command handler for the changeToFix command}
+value picoCommands(changeToFix(loc src, str newName)) {
+    applyDocumentsEdits([changed(src.top, [replace(src, newName)])]);
+    return ("result": true);
+}
+
+@synopsis{Command hamlder for the removeDecl command}
+value picoCommands(removeDecl(start[Program] program, IdType toBeRemoved)) {
+    applyDocumentsEdits([changed(program@\loc.top, [replace(toBeRemoved@\loc, "")])]);
```
DavyLandman, line 157: Can we rewrite one of these examples to not use a command, but have a edit directly defined on the action?

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
```diff
@@ -81,31 +104,73 @@ Summary picoSummarizer(loc l, start[Program] input, PicoSummarizerMode mode) {
...
+@description{
+Register the Pico language and the contributions that supply the IDE with features.
+
+((registerLanguage)) is called twice here:
+1. first for fast and cheap contributions
+2. asynchronously for the full monty that loads slower
+
+:::tip
+Run each contribution on an example in the terminal to test it first.
+Any feedback (errors and exceptions) is faster and more clearly printed in the terminal.
```
DavyLandman, line 171: 👍🏼

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
```diff
@@ -81,31 +104,73 @@ Summary picoSummarizer(loc l, start[Program] input, PicoSummarizerMode mode) {
...
+    return ("result": true);
+}
+
+@synopsis{The main function registers the Pico language with the IDE}
+@description{
+Register the Pico language and the contributions that supply the IDE with features.
+
+((registerLanguage)) is called twice here:
+1. first for fast and cheap contributions
+2. asynchronously for the full monty that loads slower
```
DavyLandman, line 167: not sure if it's more async than the other? just on a complelty different evaluator as to not impact the other?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/TreeSearch.java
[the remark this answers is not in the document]
DavyLandman: There is some relation ship to the `IRangeMap` caches that we can explore in a future PR.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/TreeSearch.java
```diff
@@ -0,0 +1,124 @@
...
+            return false;
+        }
+
+        if (line < loc.getBeginLine() || line > loc.getEndLine()) {
+            return false;
+        }
+
+        if (line == loc.getBeginLine()) {
+            // then make sure we are not before the start
+            return loc.getBeginColumn() <= column;
```
DavyLandman, line 38: what if `endLine == line`, in that case you also have to make sure that `endColumn` is after `column.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/TreeSearch.java
```diff
@@ -0,0 +1,124 @@
...
+            }
+		}
+
+		if (TreeAdapter.isAmb(tree) && !tree.getAlternatives().isEmpty()) {
+            // pick any tree to make the best of it.
+			return computeFocusList(focus, (ITree) tree.getAlternatives().iterator().next(), line, column);
+		}
+
+		if (TreeAdapter.isAppl(tree)) {
+			IList children = TreeAdapter.getASTArgs(tree); // this skips layout trees
```
DavyLandman, line 97: do we need to build a new list everytime? seems a bit expensive?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/TreeSearch.java
```diff
@@ -0,0 +1,124 @@
...
+
+			for (IValue child : children) {
+				ISourceLocation childLoc = TreeAdapter.getLocation((ITree) child);
+
+				if (childLoc == null) {
+					continue;
+				}
+
+				if (inside(childLoc, line, column)) {
```
DavyLandman, line 106: this logic seems the same as the start of this function, shouldn't we let recursion take care to only look at the interesting childs? and just only check the result of the `computeFocusList`?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
```diff
@@ -248,15 +248,22 @@ public static CompletableFuture<Evaluator> makeFutureEvaluator(ExecutorService e
...
+                
                 if (pcfg != null) {
                     for (IValue src : pcfg.getSrcs()) {
                         eval.addRascalSearchPath((ISourceLocation) src);
                     }
                 }
 
-                eval.doImport(services, imports);
-
+                for (String i : imports) {
```
DavyLandman, line 258: The merge went wrong here, this was the old code, that got removed in this commit: https://github.com/usethesource/rascal-language-servers/commit/dae2ab7eb871c71fe4dda0c668e17250b413f406

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -202,6 +218,7 @@ data LanguageService
     | definer(Definer definer)
     | referrer(Referrer reference)
     | implementer(Implementer implementer)
+    | codeActionContributor(CodeActionContributor actions)
```
DavyLandman, line 221: Somehow `codeActionContributor` seems like name that is different than the other names in this list, can we find something that matches the rest?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParserOnlyContribution.java
```diff
@@ -138,6 +138,11 @@ public InterruptibleFuture<ISet> lenses(ITree input) {
         return InterruptibleFuture.completedFuture(VF.bool(false));
     }
 
+    @Override
+    public CompletableFuture<IList> parseCodeActions(String commands) {
+        return CompletableFuture.completedFuture(null);
```
DavyLandman, line 143: don't return null, return `VF.list()`. In most places in `rascal-lsp` we try really hard to avoid `null`s.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -500,6 +577,62 @@ public CompletableFuture<SemanticTokens> semanticTokensRange(SemanticTokensRange
...
+                .map(Diagnostic::getData)
+                .filter(Objects::nonNull)
+                .filter(JsonPrimitive.class::isInstance)
+                .map(JsonPrimitive.class::cast)
+                .map(JsonPrimitive::getAsString)
+                // this is the "magic" resurrection of command terms from the JSON data field
+                .map(contribs::parseCodeActions)
+                // this serializes the stream of futures and accumulates their results as a flat list again
+                .reduce(emptyListFuture, (acc, next) -> acc.thenCombine(next, IList::concat)  
+                ).thenApply(IList::stream)
```
DavyLandman, line 606: nit: if we move the `)` to the previous line, it reads more clearly that it's just another part of the chain.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -500,6 +577,62 @@ public CompletableFuture<SemanticTokens> semanticTokensRange(SemanticTokensRange
...
+                .thenApply(IList::stream)
+            , () -> Stream.<IValue>empty())
+            ;
+
+        // final merging the two streams of commmands, and their conversion to LSP Command data-type
+        return codeActions.thenCombine(quickfixes, (actions, quicks) -> 
+                Stream.concat(quicks, actions)
+                .map(IConstructor.class::cast)
+                .map(cons -> constructorToCodeAction(contribs.getName(), cons))
+                .map(cmd  -> Either.<Command,CodeAction>forRight(cmd))
```
DavyLandman, line 625: nit: can this be `Either::forRight`?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -372,6 +386,69 @@ private CodeLens locCommandTupleToCodeLense(String languageName, IValue v) {
...
+        result.setKind(constructorToCodeActionKind(kind));
+
+        return result;
+    }
+
+    /**
+     * Translates `refactor(inline())` to `"refactor.inline"` and `empty()` to `""`, etc.
+     */
+    private String constructorToCodeActionKind(IConstructor kind) {
+        if (kind == null) {
```
DavyLandman, line 427: can it be null? if so, add `@Nullable` to the parameter.

DavyLandman (changes requested):
I like this a lot better 👍🏼 I've spotted some small bugs and points of clarifications.

rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
```diff
@@ -146,5 +147,21 @@ describe('DSL', function () {
         await lens!.click();
         await driver.wait(async () => (await editor.getTextAtLine(9)).trim() === "b := 2;", 20_000, "a variable should be changed to b");
     });
+
+    it("quick fix works"), async() => {
```
DavyLandman, line 151: ah, there is an error, the `)` should not be there, this disabled the test from running it should be `it("quick fix works", ...`

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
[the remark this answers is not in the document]
jurgenvinju: Ok thanks!

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/InterpretedLanguageContributions.java
```diff
@@ -206,13 +211,26 @@ private static ISet loadContributions(Evaluator eval, LanguageParameter lang) {
...
     }
+ 
+    @Override
+    public CompletableFuture<IList> parseCodeActions(String command) {
+        return store.thenApply(commandStore -> {
+            try {
+                var TF = TypeFactory.getInstance();
+                return (IList) new StandardTextReader().read(VF, commandStore, TF.listType(commandStore.lookupAbstractDataType("CodeAction")), new StringReader(command));
+            } catch (FactTypeUseException | IOException e) {
+                logger.catching(e);
```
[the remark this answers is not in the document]
jurgenvinju: I'm pretty sure these exceptions are impossible to happen. Since we write the data and then read it back in. But in any case it can't do harm to communicate then you the user.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -116,7 +118,21 @@ A documenter is called on-demand, when documentation is requested by the IDE use
...
+@synopsis{Function profile for retrieving code actions focused around the current cursor}
+@description{
+Next to the quickfix commands that may be attached to diagnostic ((Message))s, the LSP
+can produce refactoring and quickfix or visualization actions specific for what is near
+or under the current cursor.
+
+An action contributor is called on demand when a user presses a light-bulb or asks for quick-fixes.
+The implementor is asked to produce only actions that pertain what is under the current cursor.
+
+The parameter `inFocus` lists all the subtrees ordered from inside to outside that the current
```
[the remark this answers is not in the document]
jurgenvinju: Me too. The trees are much more "language level" than a cursor position. It relies heavily on list matching though, so that's a skill you need to have to use this properly.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -284,10 +301,108 @@ data DocumentSymbolTag
 
 data CompletionProposal = sourceProposal(str newText, str proposal=newText);
 
+@synopsis{Attach any command to a message for it to be exposed as a quick-fix code action automatically.}
+@description{
+The fixes you provide with a message will be hinted at by a light-bulb in the editor's margin.
```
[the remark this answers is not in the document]
jurgenvinju: Isn't that an "under the hood" thing? There is no relation on the Rascal level other than that they use the same datatype.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -284,10 +301,108 @@ data DocumentSymbolTag
...
+}
+@pitfalls{
+* the code for error messaging may become cluttered with code for fixes. It is advisable to only _collect_ information for the fix
+and store it in a ((Commmand)) constructor inside the ((CodeAction)), and not already execute the quickfix.
+}
+data Message(list[CodeAction] fixes = []);
+
+@synopsis{A Command is a parameter to a CommandExecutor function.}
+@description{
+Commands can be any closed term. Add any constructor you need to express the execution parameters
```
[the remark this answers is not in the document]
jurgenvinju: How about "closed term (a pure value without open variables or function/closure values embedded in it)"

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -284,10 +301,108 @@ data DocumentSymbolTag
...
+of a command.
+
+You write the ((CommandExecutor)) to interpret each kind of ((Command)) individually.
+A ((Command) constructor must have fields or keyword fields that hold the parameters of the
+to-be-executed command. 
+
+Commands are produced for delayed and optional execution by:
+* ((LensDetector)), where the will be executed if the lens is selected in the editor 
+* ((CodeActionContributor)), where they will appear in context-menus for quick-fix and refactoring
+* ((Message)), where they will appear in context-menus on lines with error or warning diagnostics
```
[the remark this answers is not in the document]
jurgenvinju: Nice! I'll add that here. Do we have a manual page on how you write TS code that does this?

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
jurgenvinju: I'll think about that. It's true this one is the odd one out.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
jurgenvinju: Constantly fighting with the layout with these chains. Thx.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
jurgenvinju: I'll try that. Shorter!

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
jurgenvinju: Yes this is Nullable.

rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
jurgenvinju: Thanks! Good catch

jurgenvinju:
I'll work on the feedback by Davy, which will change some important names so this is good to do now. If/when the tests start succeed I'll flag the PR as ready. Bringing quickfixes you rascal will be in another PR

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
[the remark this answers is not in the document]
DavyLandman: okay, yeah that's better.

DavyLandman:
:+1: great. also don't forget to merge from main branch, some updates have happened there that should help you out.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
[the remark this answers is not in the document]
DavyLandman: No, but #342 tracks the desire to make this easier to do.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
jurgenvinju: This will not work because java method references can not accept actual type parameters and at the same time infer the type of the surrounding context.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/TreeSearch.java
jurgenvinju: I'll have a look. There is also an AST children streamer somewhere. But I found in previous measurements that the stream was much slower than building the relatively short lists of a parse tree.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/TreeSearch.java
jurgenvinju: No it's different. We check one time extra that the literal is inside, but that is really a superfluous assert for documentation purposes. Once you arrive at the lexical level, you're there because that's the onl way you got there via line 106. The exception is of course if someone provides only a toplevel lexical tree, and this extra tests ensures sanity in such a case. 

We only recurse for context-free children which are `inside` on line 106.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/TreeSearch.java
jurgenvinju: that's exactly what is checked on line 41.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/TreeSearch.java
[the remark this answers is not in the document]
jurgenvinju: I don't understand where these tabs are coming from. Thanks!

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
jurgenvinju: For the user it is important to know that either call does not block the other, so "asynchronously" points to the relation between the first and the second call.

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
jurgenvinju: yes, before we merge we'll have examples of both.

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
```diff
@@ -1,5 +1,5 @@
 @license{
-Copyright (c) 2018-2023, NWO-I CWI and Swat.engineering
+Copyright (c) 2018-2024, NWO-I CWI and Swat.engineering
```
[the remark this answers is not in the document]
jurgenvinju: check!

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
jurgenvinju: inlayHinter has the same problem. 

If we follow `lenses` then:
* `inlayHinter` should be `inlays`
* `codeActionContributor` should be `actions` or `codeActions`

Any preference? I think `actions` is nice and short. That it's about code is "nogal wiedes".

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -284,10 +301,108 @@ data DocumentSymbolTag
...
+joining parent/children with periods.
+}
+data CodeActionKind
+    = empty()
+    | refactor(RefactorKind refactor = rewrite())
+    | quickfix()
+    | source(SourceActionKind source = organizeImports())
+    ;
+
+@synopsis{Used to prioritize menu options and choose relevant icons in the UI.}
```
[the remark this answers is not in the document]
jurgenvinju: yeah. it's like with the outlines. not so nice but let's keep as close to the LSP as possible.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -284,10 +301,108 @@ data DocumentSymbolTag
...
+        Command command          = noop(), 
+        str title                = command.title,
+        CodeActionKind kind      = quickfix()
+    );
+
+@synopsis{Kinds are used to prioritize menu options and choose relevant icons in the UI.}
+@description{
+This is an _open_ data type. The constructor names are used
+to compute the string values for the LSP by capitalizing and
+joining parent/children with periods.
```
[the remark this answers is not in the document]
jurgenvinju: yes

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -284,10 +301,108 @@ data DocumentSymbolTag
...
+* the `noop()` command will always be ignored.
+* _never_ add first-class functions or closures as a parameter or keyword field to a `Command`. The Command will
+be serialized, sent to the LSP client, and then sent back to the LSP server for execution. Functions can not be
+serialized, so that would lead to run-time errors.
+}
 data Command(str title="")
     = noop()
     ;
 
+data CodeAction
```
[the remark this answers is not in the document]
jurgenvinju: added lots of documentation

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -284,10 +301,108 @@ data DocumentSymbolTag
...
+}
+@examples{
+```rascal
+// here we invent a new command name `shouldBeInt` which is parametrized by a loc:
+data Command = shouldBeInt(loc src);
+
+// and we have the evaluator:
+value evaluator(shouldBeInt(loc src)) {
+    // overwrite that part of a file with `int`:
+    writeFile("int", src);
```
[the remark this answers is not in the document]
jurgenvinju: right; that's bad educationally as well. We'll come with something else.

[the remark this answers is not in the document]
jurgenvinju: I think I'll throw in a chart or a graph or something, to spark the imagination.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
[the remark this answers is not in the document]
jurgenvinju: I'd rather not motivate people to write more TS code here. Let them stir on it and think of a Rascal solution first.
And let's add a good tutor page on #342 so that people can find the right way to do it when they think of the need.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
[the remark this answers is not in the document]
jurgenvinju: O, I guess I haven't explained that very well. It does not work like that.

1. There are two sources of CodeActions on the Rascal level:
   * attached to diagnostic messages, and list that is relevant for the current error or warning message
   * provided by a rascal function that gets a focus lists
2. On the LSP level, the client asks the codeActionContributor API by providing a list of relevant diagnostics (with our previous actions already attached). This is when the above sources are merged.
3. The client picks a code actions
4. The edits are executed on the client side
5. The commands are send back to the server and our "CommandExecutor" is executed. This is the same executor that handles command events provided by `lenses`.

So it is the executor and not the codeActionContributor that handles the commands.

rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
```diff
@@ -146,5 +147,21 @@ describe('DSL', function () {
...
+        const editor = await ide.openModule(TestWorkspace.picoFile);
+        await editor.setTextAtLine(9, "az := 2;");
+        await editor.moveCursor(9,3);                   // it's where the undeclared variable `az` is
+        await ide.hasErrorSquiggly(editor, 15_000);   // just make sure there is indeed something to fix
+        const assist = await driver.wait(async() => editor.toggleContentAssist(true), 10_000, "there is should be a content assist menu at this spot.");
+
+        await assist!.getItems()
+            .then(items => items
```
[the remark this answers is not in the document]
jurgenvinju: maybe you can show me how; I need that data flow of the items to go well. That that work with more awaits?

runUItests.sh
```diff
@@ -0,0 +1,26 @@
...
+set -x;
+
+cd rascal-vscode-extension
+
+UITESTS=/tmp/vscode-uitests
+
+# cleanup to avoid contamination with previous runs
+
+rm -rf out || true
+rm -rf node_modules || true
```
DavyLandman, line 17: this is very nuclear, not needed, and will slow down your test cycle.

runUItests.sh
```diff
@@ -0,0 +1,26 @@
...
+rm -rf $UITESTS || true
```
DavyLandman, line 18: quite nuclear, will cause you to redownload vs code & chromedriver everytime. a bit of a heavy thing.


I think we should have these behind an optional flag or something (like `-c` to clean everything) or otherwise a separate script.

rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
```diff
@@ -155,12 +155,18 @@ describe('DSL', function () {
         await ide.hasErrorSquiggly(editor, 15_000);   // just make sure there is indeed something to fix
         const assist = await driver.wait(async() => editor.toggleContentAssist(true), 10_000, "there should be a content assist menu at this spot.");
 
-        await assist!.getItems()
-            .then(items => items
+        if (assist) {
+            while (!await assist.isLoaded());
```
DavyLandman, line 159: better do `driver.wait` with a timeout else the test might be stuck forever.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/InterpretedLanguageContributions.java
```diff
@@ -370,8 +403,26 @@ public CompletableFuture<SummaryConfig> getOndemandSummaryConfig() {
...
+        return InterruptibleFuture.flatten(parseCommand(command).thenCombine(
+            commandExecutor,
+            (cons, func) -> {
+                
+                if (func == null) {
+                    logger.info("Command is being executed without a registered command executor; for language " + name);
+                    throw new RuntimeException("No command executor registered for " + name);
+                }
+
+                return EvaluatorUtil.<@Nullable IValue>runEvaluator(
```
DavyLandman, line 416: is there a specific reason this is now a multiline call? I think it doesn't match the style of the other functions in this class?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/InterpretedLanguageContributions.java
```diff
@@ -370,8 +403,26 @@ public CompletableFuture<SummaryConfig> getOndemandSummaryConfig() {
...
         logger.trace("Full command: {}", command);
-        return InterruptibleFuture.flatten(parseCommand(command).thenCombine(commandExecutor,
-            (cons, func) -> EvaluatorUtil.<@Nullable IValue>runEvaluator("executeCommand", eval, ev -> func.call(cons), null, exec, true, client)
+
+        return InterruptibleFuture.flatten(parseCommand(command).thenCombine(
+            commandExecutor,
+            (cons, func) -> {
+                
+                if (func == null) {
+                    logger.info("Command is being executed without a registered command executor; for language " + name);
```
DavyLandman, line 412: nit: in log4j, it's preffered to use `{}`templates instead of string concat. 

apart from that, I think this would warrant an `error` instead of `info`.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -500,6 +579,62 @@ public CompletableFuture<SemanticTokens> semanticTokensRange(SemanticTokensRange
...
+                .thenApply(Versioned::get) 
+                .thenCompose(tree -> computeCodeActions(contribs, loc, startLine, startColumn, tree))
+                .thenApply(IList::stream)
+            , () -> Stream.<IValue>empty())
+            ;
+
+        // final merging the two streams of commmands, and their conversion to LSP Command data-type
+        return codeActions.thenCombine(quickfixes, (actions, quicks) -> 
+                Stream.concat(quicks, actions)
+                .map(IConstructor.class::cast)
```
DavyLandman, line 625: nit: since `.map` (and the others) continues on the result of `Stream.concat` I think it should be indented another 4 spaces

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -500,6 +579,62 @@ public CompletableFuture<SemanticTokens> semanticTokensRange(SemanticTokensRange
...
+                .map(cons -> constructorToCodeAction(contribs.getName(), cons))
+                .map(cmd  -> Either.<Command,CodeAction>forRight(cmd))
+                .collect(Collectors.toList())
+            );
+    }
+
+    private CompletableFuture<IList> computeCodeActions(final ILanguageContributions contribs, final ISourceLocation loc,
+            final int startLine, final int startColumn, ITree tree) {
+        IList focus = (IList) TreeSearch.computeFocusList(tree, startLine, startColumn);
```
DavyLandman, line 634: why do we need to cast to `IList`?

nit: I would have moved the invocation inline.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/TreeSearch.java
DavyLandman: I think there still is a bug. Since we never get to line 43 due to the return on this line. If the loc ends before the column (but is on the same line) it's not inside of it, but this clause would say, as long as it begins before column we're good.

rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
```diff
@@ -146,5 +147,26 @@ describe('DSL', function () {
...
+        const editor = await ide.openModule(TestWorkspace.picoFile);
+        await editor.setTextAtLine(9, "az := 2;");
+        await editor.moveCursor(9,3);                   // it's where the undeclared variable `az` is
+        await ide.hasErrorSquiggly(editor, Delays.verySlow);   // just make sure there is indeed something to fix
+        const editorView = new EditorView();
+
+        // find an editor action button by title
+        const button = await editorView.getAction("Change to a");
+
+        if (button) {
```
DavyLandman, line 161: how about instead of return true/false, we use driver.wait on the `getAction` call, so that it will retry it untill the action exists, and then if it doesn't appear in time, we have an error that says: hey we couldn't find the `Change to a` action.

rascal-vscode-extension/src/test/vscode-suite/utils.ts
```diff
@@ -269,11 +269,11 @@ export class IDEOperations {
 
 async function showRascalOutput(bbp: BottomBarPanel, channel: string) {
     const outputView = await bbp.openOutputView();
-    await outputView.selectChannel(`${channel} Language Server`);
+    await outputView.selectChannel(channel);
     return outputView;
 }
 
-export function printRascalOutputOnFailure(channel: 'Language Parametric Rascal' | 'Rascal MPL') {
+export function printRascalOutputOnFailure(channel: 'Language Parametric Rascal Language Server' | 'Rascal MPL') {
```
DavyLandman, line 276: I think the merge went wrong here. See https://github.com/usethesource/rascal-language-servers/commit/897c33d2d7c486a093d0c18e0752c1428c5cc982

as line 272 should not be changed, or otherwise the `Rascal MPL` also has to be postfixed with `Language Server`.

runUItests.sh
```diff
@@ -0,0 +1,24 @@
...
+cd rascal-vscode-extension
+
+UITESTS=/tmp/vscode-uitests
+
+# cleanup to avoid contamination with previous runs
+
+rm -rf $UITESTS || true
+
+# compiling the TS code as well as the test TS code at least once is required before execution
+npm ci
```
DavyLandman, line 19: same, put this behind a flag.

DavyLandman (commented):
Some small comments after reviewing the merge result.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/InterpretedLanguageContributions.java
jurgenvinju: Nothing crashes but it just does not have an implementation yet. That would be normal during DSL development, so I'm going for a warning.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/InterpretedLanguageContributions.java
jurgenvinju: I had a lot of benefit from putting breakpoints on the individual lines, so I'm leaving it at that for now.

rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
jurgenvinju: sure. I think you have this "in your fingers"? Can you refactor pls.

DavyLandman: done, see eb6fb747

runUItests.sh
jurgenvinju: I removed it for now and put it in comments.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/TreeSearch.java
jurgenvinju: ok I'll have a look. thanks

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/TreeSearch.java
```diff
@@ -0,0 +1,126 @@
...
+     * begin line/column and end line/column scope. The first and the last lines are
+     * where the column information is used.
+     *
+     * @param loc     location with line/column information that acts as a range
+     * @param line    line to check if its within the range
+     * @param column  column to see if we are within the range
+     * @return true iff the given `loc` range spans around the line/column position, inclusively.
+     */
+    private static boolean inside(ISourceLocation loc, int line, int column) {
+        line += 1; // LSP line offsets are 0-based, while Rascal tree locations are 1-based.
```
jurgenvinju, line 28: This should be done outside of this class. There we can handle the mapping from LSP line and columns to Rascal lines and columns. This class should be only in the Rascal world (now lines are mapped but columns are still not.)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/TreeSearch.java
jurgenvinju: you're right! Fixed it.

rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
jurgenvinju: nice!

jurgenvinju: 1) DSL
       quick fix works:
     TimeoutError: We should get the change to a code action
Wait timed out after 40221ms
      at /home/runner/actions-runner/_work/rascal-language-servers/rascal-language-servers/rascal-vscode-extension/node_modules/selenium-webdriver/lib/webdriver.js:915:22
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

jurgenvinju: This is still the case; right after the rewrite and continuing into the other fixes I've made.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/impl/TreeSearch.java
jurgenvinju: I just removed the fix here. In the surrounding API there was already a correction for 1-based to 0-based. So it is strange that this worked, and it's strange that it still works.

rascal-vscode-extension/src/lsp/ParameterizedLanguageServer.ts
```diff
@@ -41,7 +41,7 @@ export class ParameterizedLanguageServer implements vscode.Disposable {
         private readonly absoluteJarPath: string,
         private readonly deployMode = true,
         private readonly languageId = 'parametric-rascalmpl',
-        private readonly title = 'Language Parametric Rascal Language Server',
```
DavyLandman, line 44: This should be reverted. Or is there a reason this is happening @jurgenvinju ?

jurgenvinju: Apparantly this has been reverted and the code in the comment is out of date.

DavyLandman: No, there is still a change in this line, that I don't think should be here.

DavyLandman: I've removed this change, lets see if the test sill succeed.

DavyLandman (approved):
I think this code is ready for merging. @jurgenvinju when all the test has passed, you do the honors 🎉

jurgenvinju:
Done. This was a very nice collab @DavyLandman

PR #407, opened 2024-07-17
Generate semantic tokens to indicate absence of highlighting

sungshik:
This PR extends the semantic highlighter with generation of tokens to explicitly indicate absence of highlighting. To avoid generating an excessive number of such tokens, the tokenizer on-the-fly identifies maximal consecutive regions of "absence of highlighting" (separated by newlines) and generates only a single token per region.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
@@ -102,25 +102,57 @@ public class SemanticTokenizer implements ISemanticTokens {
 
     private static class TokenList {
         private final List<Integer> theList = new ArrayList<>(500);
-        private int previousLine = 0;
-        private int previousStart = 0;
+        private int previousLineAbsolute = 0;
+        private int previousStartAbsolute = 0;
 
         public List<Integer> getTheList() {
             return Collections.unmodifiableList(theList);
         }
 
-        public void addToken(int startLine, int startColumn, int length, @Nullable String category) {
-            int tokenCategory = TokenTypes.tokenTypeForName(category);
+        public void addToken(int lineAbsolute, int startAbsolute, int length, @Nullable String category) {
+            int type = TokenTypes.tokenTypeForName(category);
 
-            // https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/#textDocument_semanticTokens
-            theList.add(startLine - previousLine);
-            theList.add(startLine == previousLine ? startColumn - previousStart : startColumn);
-            theList.add(length);
-            theList.add(tokenCategory);
-            theList.add(0); // no support for modifiers yet
+            // If the token-to-add occurs right after the previous token, and if
+            // it has the same category, then the previous token can just be
+            // made longer (instead of adding a new token)
+            boolean canGrowPreviousToken =
+                !theList.isEmpty() &&
+                lineAbsolute == previousLineAbsolute &&
+                startAbsolute == previousStartAbsolute + previous(TokenField.LENGTH) &&
+                type == previous(TokenField.TYPE);
 
-            previousLine = startLine;
-            previousStart = startColumn;
+            if (canGrowPreviousToken) {
+                growPreviousToken(length);
+            } else {
+                // https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/#textDocument_semanticTokens
+                theList.add(lineAbsolute - previousLineAbsolute);
+                theList.add(lineAbsolute == previousLineAbsolute ? startAbsolute - previousStartAbsolute : startAbsolute);
+                theList.add(length);
+                theList.add(type);
+                theList.add(0); // no support for modifiers yet
+                previousLineAbsolute = lineAbsolute;
+                previousStartAbsolute = startAbsolute;
+            }
+        }
+
+        private enum TokenField {
+            // The order of these values is significant (consistent with the
+            // order of integers in `theList`)
+            LINE, START, LENGTH, TYPE, MODIFIER
+        }
+
+        private int previousIndexOf(TokenField field) {
+            return theList.size() - (5 - field.ordinal());
+        }
+
+        private int previous(TokenField field) {
+            return theList.get(previousIndexOf(field));
+        }
+
+        private void growPreviousToken(int newLength) {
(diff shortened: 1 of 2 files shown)


rascal-vscode-extension/package.json
```diff
@@ -147,6 +147,9 @@
         "scopes": {
           "ambiguity": [
             "invalid"
+          ],
+          "uncategorized": [
+            "meta.embedded"
```
sungshik, line 152: VS Code primarily supports `meta.embedded` for TextMate grammars with embedded languages, but as stated in their [documentation](https://code.visualstudio.com/api/language-extensions/syntax-highlight-guide#injection-grammars) (subsection "token types and embedded languages"), it seems to serve our use case, too:

> you can use a `meta.embedded.*` scope to reset VS Code's marking of tokens

I couldn't find a more suitable scope, neither [here](https://macromates.com/manual/en/language_grammars), nor [here](https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#semantic-token-provider), nor [there](https://stackoverflow.com/questions/76574973/what-textmate-scope-is-used-for-normal-code-text-in-vs-code)...

sungshik (commented):
Just a few additional comments

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -102,35 +102,73 @@ public SemanticTokensCapabilities capabilities() {
...
+        }
+
+        private int previous(TokenField field) {
+            return theList.get(previousIndexOf(field));
+        }
+
+        private void growPreviousToken(int newLength) {
+            int i = previousIndexOf(TokenField.LENGTH);
+            int oldLength = theList.remove(i);
+            theList.add(i, oldLength + newLength);
```
DavyLandman, line 155: As this code gets called a lot, I think it would be faster to not shuffle around the list.

```
theList.set(i, theList.get(i) + newLenght);
```

would be faster I expect.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -102,35 +102,73 @@ public SemanticTokensCapabilities capabilities() {
...
-                previousStart = startColumn;
+                previousLineAbsolute = lineAbsolute;
+                previousStartAbsolute = startAbsolute;
             }
         }
+
+        private enum TokenField {
+            // The order of these values is significant (consistent with the
+            // order of integers in `theList`)
+            LINE, START, LENGTH, TYPE, MODIFIER
```
DavyLandman, line 141: nice way to encode offset in the array list. you almost want to think about java records, or an intermediate class, but I guess for perf reasons we keep this code as "terse" as this is.

rascal-vscode-extension/package.json
DavyLandman: Can we also take a look in the source code of the vs code rendering which it uses as fallback/default?

since embedded can also be used for cases like html inside of typescript. (jsx)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -102,35 +102,73 @@ public SemanticTokensCapabilities capabilities() {
 
     private static class TokenList {
         private final List<Integer> theList = new ArrayList<>(500);
-        private int previousLine = 0;
-        private int previousStart = 0;
+        private int previousLineAbsolute = 0;
```
DavyLandman, line 105: as the statemachine around previous & current is getting bigger, I'm starting to wonder if we should do something where we collect the "current" token, and only flush when the category changes (or there is a gap). Such that we only write to `theList` once we've found a new token (or are at the end of the tree).

DavyLandman (approved):
This looks fine, my 2 main questions:

1. Maybe we can rewrite `TokenList` a bit more to better track the state
2. see if there is a different token we could use to mark an empty category.

rascal-vscode-extension/package.json
sungshik: Each theme defines general colors and token colors. For instance ([Dark](https://github.com/microsoft/vscode/blob/main/extensions/theme-defaults/themes/dark_vs.json)):

```
{
	"$schema": "vscode://schemas/color-theme",
	"name": "Dark (Visual Studio)",
	"colors": {
		"checkbox.border": "#6B6B6B",
		"editor.background": "#1E1E1E",
		"editor.foreground": "#D4D4D4",
		...
	},
	"tokenColors": [
    ...
		{
			"scope": "comment",
			"settings": {
				"foreground": "#6A9955"
			}
		},
		...
	]
}
```

In this theme, the only standard token colors that have the same `foreground` property as the general `editor.foreground` property are:

  - `meta.embedded`
  - `meta.template.expressions`
  - `keyword.operator`

Of these, combined with the quote in my previous comment, I think `meta.embedded` is the most suitable for us.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
sungshik: Yes, that would be a more direct way of doing it. I haven't investigated in detail, but the logic around the usage of `TokenList` (particularly in the `collect...` methods) makes this a bit hard to do. I think we should do improve this as part of the nested `syntax` non-terminal thing (which probably requires changes to the logic in the `collect...` methods anyway).

rascal-vscode-extension/package.json
DavyLandman: `editor.foreground` is not a valid scope I'm guessing?

and there is no -1 token we can pass in?

DavyLandman: I've just found:

```
const enum SemanticTokensProviderStylingConstants {
	NO_STYLING = 0b01111111111111111111111111111111
}
```

in the source code of VS, which is triggered if we send an token index that is not in the list of registered tokens. Have you tried what happens if we send an invalid token number? looking at the code, it might just be ignored in that case.

Otherwise, agreed, lets use `meta.embedded`.

sungshik: I can't seem to get `""` to work. Briefly discussed and decided to keep `meta.embedded` for now.

PR #408, opened 2024-07-18
Report an error and allow user to quickly open a GH issue

### LINKED ISSUE #406: Propogate crash of typechecker as annotation on first line of the module
Right now, when the typechecker crashes, the user gets no warning or message, they have to look in output window to figure out what is up. 

If the typechecker crashed, add an error to the module on which the typechecker was called on.

keep it as in, this would however make it harder for users to debug why their code is not showing errors.

DavyLandman:
This improves UX by reporting a crash inside one of the rascal callbacks of the rascal-lsp server to a user.

Before it would crash, dump a log to the output window, and only for some of the errors, report anything to the end user.

This PR shows an error:

image

and then allows the user to directly open a issue:

image

(not that if the error is a rascal error, you should see a a rascal stack trace).

This fixes #406


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -28,7 +28,6 @@
 
 import static org.rascalmpl.vscode.lsp.util.EvaluatorUtil.makeFutureEvaluator;
 import static org.rascalmpl.vscode.lsp.util.EvaluatorUtil.runEvaluator;
-
```
toinehartman, line 31: Did those disappear due to automatic import optimization? I think we should reduce changes in order/layout as much as possible here. Perhaps we align our editor configs or sync settings.

toinehartman (approved):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
DavyLandman: This happens due to optimize imports that I have in my muscle memory to drop unused imports.

toinehartman: Understood. But I also use optimize imports, and mine does not drop empty lines, for example. I think we should make sure to align settings as much as possible to prevent messy diffs.

DavyLandman: I assumed it was a result of some kind of merge conflict, or someone manually added a newline. I have no specific config, except that I've installed the editor config plugin. That helps normalize these kinds of things on a per repo basis.

PR #409, opened 2024-07-19
Dynamically add/update categories to productions of Rascal literals

sungshik:
This PR adds a semantic tokenizer feature (intended to be temporary) to dynamically add/update categories to productions in Rascal's own grammar with no category (e.g., number literals) or with a legacy category (e.g., string literals).

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
@@ -457,11 +457,10 @@ public class SemanticTokenizer implements ISemanticTokens {
                 }
             }
 
-            // If this semantic tokenizer is working with Rascal's own grammar,
-            // then hot-patch the parse tree by adding categories to otherwise
-            // uncategorized literals. These categories should eventually be
-            // incorporated directly in the grammar. Additional background:
-            // https://github.com/SWAT-engineering/rascal-textmate/pull/6.
+            // Apply a patch to the parse tree to dynamically add categories. In
+            // case of the rascal server, the patch adds categories to otherwise
+            // uncategorized literals. In case of parametric servers, the patch
+            // is empty and does nothing.
             category = patch.apply(prod, category);
 
             // now we go down in the tree to find more tokens and to advance the counters
@@ -542,7 +541,9 @@ public class SemanticTokenizer implements ISemanticTokens {
 
     // The idea behind the patch is to dynamically map productions in Rascal's
     // own grammar with no category (e.g, number literals) to current categories
-    // (i.e., semantic token types).
+    // (i.e., semantic token types). These categories should eventually be
+    // incorporated directly in the grammar. Additional background:
+    // https://github.com/SWAT-engineering/rascal-textmate/pull/6.
     private interface CategoryPatch extends BiFunction<IConstructor, String, String> {};
 
     private static class DefaultCategoryPatch implements CategoryPatch {
@@ -558,8 +559,8 @@ public class SemanticTokenizer implements ISemanticTokens {
         // past, are cached for fast lookups in the future. There are separate
         // caches for productions that *do* need to be mapped to a new category
         // and those that *do not* need this.
-        private Cache<IConstructor, String>  doPatch = Caches.build();
-        private Cache<IConstructor, Boolean> doNotPatch = Caches.build();
+        private Cache<IConstructor, String>  doPatch = buildCache();
+        private Cache<IConstructor, Boolean> doNotPatch = buildCache();
 
         @SuppressWarnings("java:S131") // Switches without defaults are intended in this method
         @Override
@@ -582,11 +583,11 @@ public class SemanticTokenizer implements ISemanticTokens {
                 case "integer":
                 case "real":
                 case "rational":
-                    return Caches.putThenGet(doPatch, prod, SemanticTokenTypes.Number);
+                    return doPatch.get(prod, p -> SemanticTokenTypes.Number);
                 case "location":
-                    return Caches.putThenGet(doPatch, prod, SemanticTokenTypes.String);
+                    return doPatch.get(prod, p -> SemanticTokenTypes.String);
                 case "regExp":
-                    return Caches.putThenGet(doPatch, prod, SemanticTokenTypes.Regexp);
+                    return doPatch.get(prod, p -> SemanticTokenTypes.Regexp);
                 }
             }
 
@@ -598,23 +599,17 @@ public class SemanticTokenizer implements ISemanticTokens {
             return SymbolAdapter.isLabel(def) &&
                 SymbolAdapter.getName(SymbolAdapter.getLabeledSymbol(def)).equals("Literal");
         }
-    }
 
-    private static class Caches {
-
-        public static <K, V> Cache<K, V> build() {
-            // The usage of weak keys and values ensures: (1) identity equality
-            // is used to compare keys; (2) entries are automatically evicted
-            // and garbage-collected when they are no longer strongly reachable.
+        private static <V> Cache<IConstructor, V> buildCache() {
+            // The usage of weak keys ensures: (1) identity equality is used to
+            // compare keys; (2) entries are automatically evicted and
+            // garbage-collected when they are no longer strongly reachable.
+            //
+            // Note: Parsers have static fields that contain references to the
+            // IConstructors, so using weak keys is safe.
             return Caffeine.newBuilder()
                 .weakKeys()
-                .weakValues()
                 .build();
-        }
-
-        private static <K, V> V putThenGet(Cache<K, V> cache, K k, V v) {
-            cache.put(k, v);
-            return v;
         }
     }
 }
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
@@ -539,16 +539,18 @@ public class SemanticTokenizer implements ISemanticTokens {
         }
     }
 
+    // The idea behind the patch is to dynamically map productions in Rascal's
+    // own grammar with no category, or with a legacy category (e.g., string
+    // literals), to current categories (i.e., semantic token types).
     private static class RascalPatch {
 
-        // The idea behind the patch is to dynamically map category-less
-        // productions to categories. As an optimization, productions that have
-        // been mapped before are cached in an identity map/set for fast
-        // lookups.
+        // As an optimization, productions that have already been mapped in the
+        // past, are cached in an identity map/set for fast lookups in the
+        // future.
         Map<IConstructor, String> positiveCache = new IdentityHashMap<>();
         Set<IConstructor> negativeCache = Collections.newSetFromMap(new IdentityHashMap<>());
 
-        @SuppressWarnings("java:S131") // Switches without defaults are intended here
+        @SuppressWarnings("java:S131") // Switches without defaults are intended in this method
         public String apply(IConstructor prod, String defaultCategory) {
 
             // Check the caches
@@ -561,6 +563,7 @@ public class SemanticTokenizer implements ISemanticTokens {
 
             // Apply the patch (positively)
             var def = ProductionAdapter.getDefined(prod);
+
             if (isLabeledLiteral(def)) {
                 switch (SymbolAdapter.getLabel(def)) {
                 case "integer":
@@ -573,6 +576,7 @@ public class SemanticTokenizer implements ISemanticTokens {
                     return putIntoPositiveCache(prod, SemanticTokenTypes.Regexp);
                 }
             }
+
             if (SymbolAdapter.isLex(def)) {
                 switch (SymbolAdapter.getName(def)) {
                 case "StringConstant":
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
@@ -585,6 +585,7 @@ public class SemanticTokenizer implements ISemanticTokens {
                 case "PreStringChars":
                 case "MidStringChars":
                 case "PostStringChars":
+                case "Char":
                     return putIntoPositiveCache(prod, SemanticTokenTypes.String);
                 }
             }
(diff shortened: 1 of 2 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -514,4 +538,71 @@ else if (Character.isSupplementaryCodePoint(currentChar)) {
...
+            if (SymbolAdapter.isLex(def)) {
+                switch (SymbolAdapter.getName(def)) {
+                case "StringConstant":
+                case "CaseInsensitiveStringConstant":
+                case "PreStringChars":
+                case "MidStringChars":
+                case "PostStringChars":
+                case "Char":
+                    return putIntoPositiveCache(prod, SemanticTokenTypes.String);
+                }
```
sungshik, line 590: Now that we are hot-patching all string-related productions here (which have `Constant` as category in the current grammar; see also [here](https://github.com/SWAT-engineering/rascal-textmate/pull/6#issuecomment-2228180407)), the following line might be changed too:

https://github.com/usethesource/rascal-language-servers/blob/615ee4c050c215280bcb170526173fb754dddd3c/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java#L191

Alternatively, we could decide not to hot-patch the string-related productions for performance (as the mapping of `Constant` already gives the same result).

sungshik (commented):
Just an additional comment

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -439,6 +454,15 @@ private void collectAppl(ITree arg, @Nullable String currentCategory) {
                 }
             }
 
+            // If this semantic tokenizer is working with Rascal's own grammar,
+            // then hot-patch that grammar by adding categories to otherwise
```
DavyLandman, line 458: nit: we're hotpatching the tree.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -58,10 +61,20 @@ public class SemanticTokenizer implements ISemanticTokens {
...
     private static final Logger logger = LogManager.getLogger(SemanticTokenizer.class);
 
+    private final @Nullable RascalPatch patch;
+
+    public SemanticTokenizer() {
+        this(false);
+    }
+
+    public SemanticTokenizer(boolean rascal) {
+        this.patch = rascal ? new RascalPatch() : null;
```
DavyLandman, line 71: I would prefer an an identity one `(t) -> t`, to avoid downstream null checks?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -514,4 +538,71 @@ else if (Character.isSupplementaryCodePoint(currentChar)) {
...
+    // The idea behind the patch is to dynamically map productions in Rascal's
+    // own grammar with no category (e.g, number literals), or with a legacy
+    // category (e.g., string literals), to current categories (i.e., semantic
+    // token types).
+    private static class RascalPatch {
+
+        // As an optimization, productions that have already been mapped in the
+        // past, are cached in an identity map/set for fast lookups in the
+        // future.
+        Map<IConstructor, String> positiveCache = new IdentityHashMap<>();
```
DavyLandman, line 551: Good idea, but we should use caffeine, and configure it to also clear the map. It's not a direct problem as we're only doing it for rascal productions, and they won't grow. But if we end up using for DSLs at some point, this will become a memory leak, as old productions (from a previous version of the grammar) will never be cleared.

We used to have cases where people would run out of memory after having eclipse open for a few weeks because of a cache like this.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -514,4 +538,71 @@ else if (Character.isSupplementaryCodePoint(currentChar)) {
             }
         }
     }
+
+    // The idea behind the patch is to dynamically map productions in Rascal's
+    // own grammar with no category (e.g, number literals), or with a legacy
+    // category (e.g., string literals), to current categories (i.e., semantic
+    // token types).
+    private static class RascalPatch {
```
DavyLandman, line 546: This whole class should be made thread-safe. As the semantic tokenizer can be called from multiple threads, in parallel.

Caffeine will also help with that, without having to spray around locks everywhere.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -514,4 +538,71 @@ else if (Character.isSupplementaryCodePoint(currentChar)) {
...
+        // past, are cached in an identity map/set for fast lookups in the
+        // future.
+        Map<IConstructor, String> positiveCache = new IdentityHashMap<>();
+        Set<IConstructor> negativeCache = Collections.newSetFromMap(new IdentityHashMap<>());
+
+        @SuppressWarnings("java:S131") // Switches without defaults are intended in this method
+        public String apply(IConstructor prod, String defaultCategory) {
+
+            // Check the caches
+            if (positiveCache.containsKey(prod)) {
```
DavyLandman, line 558: better to do  get, and if result is not null, return it. avoids double lookup in the map.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -514,4 +538,71 @@ else if (Character.isSupplementaryCodePoint(currentChar)) {
...
+    // own grammar with no category (e.g, number literals), or with a legacy
+    // category (e.g., string literals), to current categories (i.e., semantic
+    // token types).
+    private static class RascalPatch {
+
+        // As an optimization, productions that have already been mapped in the
+        // past, are cached in an identity map/set for fast lookups in the
+        // future.
+        Map<IConstructor, String> positiveCache = new IdentityHashMap<>();
+        Set<IConstructor> negativeCache = Collections.newSetFromMap(new IdentityHashMap<>());
```
DavyLandman, line 552: Can we add a comment on the negative cache? Positive and negative are so generic, that for me they either need a small comment, or a sharper name.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -514,4 +538,71 @@ else if (Character.isSupplementaryCodePoint(currentChar)) {
...
+                case "rational":
+                    return putIntoPositiveCache(prod, SemanticTokenTypes.Number);
+                case "location":
+                    return putIntoPositiveCache(prod, SemanticTokenTypes.String);
+                case "regExp":
+                    return putIntoPositiveCache(prod, SemanticTokenTypes.Regexp);
+                }
+            }
+
+            if (SymbolAdapter.isLex(def)) {
```
DavyLandman, line 581: is this not an else if?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
DavyLandman: lets do that! less code is better :+1:

DavyLandman (changes requested):
Mostly looks good, except for some concerns about the concurrency.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
sungshik: Fixed.

However, right now, I configured it so that cache entries are auto-evicted and garbage-collected as soon as the keys are no longer reachable. For Rascal's grammar, when the RascalServer runs for weeks, I expect this will never happen (because the grammar objects presumably are kept in memory the whole time). I could add other eviction policies based on time and/or number of cache entries, but I wonder if this is worth the payoff: given that Rascar's own parser is used all the time (when writing Rascal), many/most of the cache entries would very quickly be re-inserted I guess.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
sungshik: Renamed the caches

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
sungshik: Fixed

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
sungshik: It could have been, yes, but it's gone now anyway (next comment)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
sungshik: Fixed

sungshik (commented):
Thank you for the comments! They're all fixed now.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -514,4 +539,82 @@ else if (Character.isSupplementaryCodePoint(currentChar)) {
...
+
+    private static class Caches {
+
+        public static <K, V> Cache<K, V> build() {
+            // The usage of weak keys and values ensures: (1) identity equality
+            // is used to compare keys; (2) entries are automatically evicted
+            // and garbage-collected when they are no longer strongly reachable.
+            return Caffeine.newBuilder()
+                .weakKeys()
+                .weakValues()
```
DavyLandman, line 611: we don't need the overhead of weak values, as they'll be interned strings anyway.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -514,4 +539,82 @@ else if (Character.isSupplementaryCodePoint(currentChar)) {
...
+                .build();
+        }
+
+        private static <K, V> V putThenGet(Cache<K, V> cache, K k, V v) {
```
DavyLandman, line 615: in a way, the `Cache::get` function of caffeine does this already. at the cost of a mapping function.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -514,4 +539,82 @@ else if (Character.isSupplementaryCodePoint(currentChar)) {
...
+        private boolean isLabeledLiteral(IConstructor def) {
+            return SymbolAdapter.isLabel(def) &&
+                SymbolAdapter.getName(SymbolAdapter.getLabeledSymbol(def)).equals("Literal");
+        }
+    }
+
+    private static class Caches {
+
+        public static <K, V> Cache<K, V> build() {
+            // The usage of weak keys and values ensures: (1) identity equality
```
DavyLandman, line 606: while I like weak keys, I'm not sure if it's a safe enough assumption that somebody else will keep around a handle to the constructors in the grammar.

ok, I've checked, and indeed the parsers have a static field that contain the constructor, so that's safe and a smart choice. maybe document it in the comments of this cache.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -514,4 +539,82 @@ else if (Character.isSupplementaryCodePoint(currentChar)) {
...
+
+        private boolean isLabeledLiteral(IConstructor def) {
+            return SymbolAdapter.isLabel(def) &&
+                SymbolAdapter.getName(SymbolAdapter.getLabeledSymbol(def)).equals("Literal");
+        }
+    }
+
+    private static class Caches {
+
+        public static <K, V> Cache<K, V> build() {
```
DavyLandman, line 605: nit: does the key need to be generic? we're only using it for a very specific case I think?

DavyLandman (approved):
Some minor suggestions, where we can simplify the code a bit more, but looks good 👍🏼

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -439,6 +457,13 @@ private void collectAppl(ITree arg, @Nullable String currentCategory) {
                 }
             }
 
+            // If this semantic tokenizer is working with Rascal's own grammar,
```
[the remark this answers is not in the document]
sungshik: Fixed:

  - Moved the final two sentences down
  - Rephrased the first two sentences

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
sungshik: True, it doesn't have to be. But the cost of genericity was really low in this case 😉

I've changed it though, because the comment about the safe usage of weak keys (below) breaks the genericity and is specific to keys being constructors

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
sungshik: Ah yes, good catch

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
sungshik: Added a comment

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
sungshik: Fixed. (The cost of the mapping function should be negligible in this case.)

DavyLandman: Agreed

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
DavyLandman: ok, good :+1:

Issue #410, opened 2024-07-19
Error on save for `rsc` files outside project


sungshik:
Context:

I'm running second level and opened a few test `rsc` files outside a proper project folder (no folder is opened in Explorer). When I save, I get the exception below (`src` and `target` indeed do not exist).

Exception thrown:
```
|lib://rascal-core/lang/rascalcore/check/Checker.rsc|:260,14: [
  error(
    "PathConfig `srcs`: |file:///c:/Users/sung-/Desktop/src| does not exist",
    |file:///c:/Users/sung-/Desktop/src|),
  error(
    "PathConfig `resources`: IO(\"Unsupported scheme \\\'target\\\'\")",
    |target:///|)
]
```
Stacktrace:
```
	at rascalTModelForLocs(|lib://rascal-core/lang/rascalcore/check/Checker.rsc|(10440,27,<259,26>,<261,5>))
	at check(|lib://rascal-core/lang/rascalcore/check/Checker.rsc|(22694,64,<545,9>,<545,73>))
	at $shell$(|main://$shell$|)

```

DavyLandman:
True, this is a variant of #341, vs code plugin makes some assumptions about project setup. The `RASCAL.MF` checker already checks some, but if thats not there, we should maybe do something better there.

We should:

1. Warn the user about an unsupported project setup
2. Think about an better error to report in the checker.

DavyLandman:
Closing this one, as the detector of project setup should give a better warning for his.

Issue #341, opened 2023-12-04
Project does not load if Rascal project is not the root of the workspace in VS Code


linuswagner:
I'm unsure if this belongs here or is better off in the documentation of Rascal.

If you add a folder to VS Code that contains one or multiple Rascal projects (i.e. folders again) and try to launch code from a file inside these using `Import in new Rascal terminal`, this throws you an
`Could not import module SOME_MODULE_NAME: can not find in search path` error in the REPL.

It's unclear to me if this is something, we should document somewhere ("Make sure to load the Rascal project directly and add other folders to the workspace") or fix somewhere in the plugin. In any case, it is hard to understand if you don't know how Rascal/Java defines the search path for the JVM

DavyLandman:
true, we have the assumption that any folder is the root of a project. so 

I think we should indeed document that. and add some better warnings around that if we detect it.

so you have to add every subfolder as a "root" folder to your workspace.

DavyLandman:
Closing this, as we now have warnings/validators for this. Thanks for the inspiration @linuswagner

PR #411, opened 2024-07-19
Detect and warn about an incorrect project setup

DavyLandman:
This module now checks for some common errors:

- rascal file opened that is not part of any of the projects active in the workspace
- rascal file without a pom.xml in the right position
- rascal file without a rascal.mf in the right place
- rascal file in a path not defined in the rascal.mf sources key.

Some screenshots:

image
image

Btw, yes, we could do this in rascal with some validation, but that would involve another extension on the LSP protocol. Until we've grown the checker to deal with this properly, we can help the user make less mistakes.

Ideally we would also add some quick fixes, but that's not always so simple to figure out.

This prevents cases like #410 and #341, or at least leaves the user less confused.


jurgenvinju (approved):
Very useful stuff! Indeed we could think about replacing it by rascal code later; in the mean time our users will be thankful and I think we too 🤓

PR #419, opened 2024-08-05
Update license 2024

toinehartman:
Update the license header template, and apply it to all headers.


DavyLandman (changes requested):
I seem to remember @jurgenvinju had good reasons not to do this.

So if you want to propose this, we should understand what it means to change the year of files that haven't been touched in 2024.

DavyLandman (approved):

PR #427, opened 2024-08-20
Run CI on more PRs

toinehartman:
- Run build on all PRs
- Run linter on all PRs.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -7,7 +7,7 @@ on:
       - 'v[0-9]+.*'
   pull_request:
     branches:
-      - main
+      - '**'
 
 permissions:
   actions: read
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -11,7 +11,7 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
     branches:
-      - main
+      - '**'
 
 
 permissions:


toinehartman:
Linting is indeed skipped for this PR
afbeelding

.github/workflows/build.yaml
```diff
@@ -7,7 +7,7 @@ on:
       - 'v[0-9]+.*'
   pull_request:
     branches:
-      - main
+      - '**'
```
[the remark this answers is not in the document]
toinehartman: A, both should be 2 stars, for branches with a slash in them. Will fix.

.github/workflows/lint.yaml
```diff
@@ -11,7 +11,10 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
     branches:
-      - main
+      - '*'
+    paths:
+      - '**.{js,ts,java}'
```
DavyLandman, line 16: let's also do .xml and json? since we might change a lint config/dependency, and we want to know it breaks something (as is for example the case in #422 )

toinehartman: And `yml` itself probably?

DavyLandman: maybe, so uhm, all files then?

DavyLandman: I think one of our hinters also check license headers in rascal? or is that during mvn build?

toinehartman: That's `maven`. But still, I'll remove the file condition here, as it might cause the linter not to be ran unintentionally.

DavyLandman (approved):

PR #437, opened 2024-08-30
Add rascal standard library for REPLs without a project

### LINKED ISSUE #433: Generic Rascal Terminal (without a project) does not have standard library in srcs configured anymore.
**Describe the bug**

see subject

**To Reproduce**

Use the CMD+P "Create Rascal Terminal" command and use/import any standard library module.

jurgenvinju:
the no-project evaluator now gets the std lib on its path again

fixes #433

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/LSPTerminalREPL.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/LSPTerminalREPL.java
@@ -165,9 +165,10 @@ public class LSPTerminalREPL extends BaseREPL {
                             pcfg = PathConfig.fromSourceProjectRascalManifest(projectDir, RascalConfigMode.INTERPETER);
                         }
                         else {
-                            // TODO: see if we need to do something special for RascalConfigMode flag for this default path config
                             pcfg = new PathConfig();
+                            pcfg.addSourceLoc(URIUtil.rootLocation("std"));
                         }
+                        
                         evaluator.getErrorPrinter().println("Rascal Version: " + RascalManifest.getRascalVersionNumber());
                         evaluator.getErrorPrinter().println("Rascal-lsp Version: " + getRascalLspVersion());
                         new StandardTextWriter(true).write(pcfg.asConstructor(), evaluator.getErrorPrinter());


jurgenvinju:
the release notes symlink is unintended. I committed that before my holidays and did not push.

rascal-lsp/RELEASE-NOTES
```diff
@@ -0,0 +1 @@
+../rascal-vscode-extension/README.md
```
DavyLandman, line 1: that is not needed, `rascal-lsp/RELEASE-NOTES.md` already exists.

jurgenvinju:
it's a necessary link for the tutor to generate documentation for the rascal-lsp library.

rascal-lsp/RELEASE-NOTES
jurgenvinju: don't we have a duplicate then? or is this a different kind of release notes?

DavyLandman: yes, as that document describes it's duplicate, primarily because the versions of vs code and rascal-lsp don't align. And the release steps have been updated to include writing this duplicate entries.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/LSPTerminalREPL.java
```diff
@@ -165,9 +165,10 @@ protected Evaluator constructEvaluator(InputStream input, OutputStream stdout, O
                             pcfg = PathConfig.fromSourceProjectRascalManifest(projectDir, RascalConfigMode.INTERPETER);
                         }
                         else {
-                            // TODO: see if we need to do something special for RascalConfigMode flag for this default path config
                             pcfg = new PathConfig();
+                            pcfg.addSourceLoc(URIUtil.rootLocation("std"));
```
DavyLandman, line 169: This looks good to me, the TODO was accurate ;)

rascal-lsp/RELEASE-NOTES
jurgenvinju: for me rascal-lsp/RELEASE-NOTES.md is a new file. How did I get myself into this situation?

PR #448, opened 2024-09-05
Fix flaky detection of source path caused by VS Code flipping the case of drive letters sometimes

### LINKED ISSUE #442: For some modules, you get a warning about an incorrect sourc path, even though other siblings are fine.
**Describe the bug**
image

even thouh `salix::canvas::Demo` is fine.

**To Reproduce**

open `salix-contrib` and go to `salix/canvas/Heart.rsc`

DavyLandman:
Fixes #442

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -7,10 +7,10 @@ extra_flags=''
 clean="clean"
 while getopts 'fd' flag; do
   case "${flag}" in
-    f) extra_flags='-Drascal.compile.skip -Drascal.tutor.skip' ;;
+    f) extra_flags='-Drascal.compile.skip -Drascal.tutor.skip -DskipTests' ;;
     d) clean='' ;;
     *) printf "incorrect param, valid params:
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -7,10 +7,10 @@ extra_flags=''
 clean="clean"
 while getopts 'fd' flag; do
   case "${flag}" in
-    f) extra_flags='-Drascal.compile.skip -Drascal.tutor.skip' ;;
+    f) extra_flags='-Drascal.compile.skip -Drascal.tutor.skip -DskipTests' ;;
```
toinehartman, line 10: Are these changes committed intentionally?

toinehartman (approved):
LGTM. Thanks for this fix! I think I ran into this a couple of times and it's pretty annoying.

build.sh
DavyLandman: yes, but primarily since I didn't want to have a separate PR just for that addition.

PR #449, opened 2024-09-05
Reduce flakyness of the rename test by retrying the command to send the tename

DavyLandman:
After inspect a few of the crashes, it seemed to always fail at the `sendKeys` part. so lets just retry that whole thing on failure, and lets hope the mac tests don't fail again.


toinehartman (approved):
Although I would have preferred a more deterministic solution, I think this is a very nice way to get rid of the annoying failing test.

PR #452, opened 2024-09-19
wrap command return type

jurgenvinju:
The current CommandExecutor definition is `value (Command _command)`, so it can return anything
including `true` and `1`.

However, if we do that the JSON-RPC bridge breaks. The LSP4J method to implement is:
` public CompletableFuture<Object> executeCommand(ExecuteCommandParams params) {`

But `Object` must not be primitive or become primitive.

Therefore the current code in Rascal language servers (repeated for every language times every command) wraps every result in a map like so:
`return ("result" : true)`

This PR changes that such that we can write:
`return true;`

And for backward compatibility it wraps the result before sending it over the bridge:
`return vf.map().put(vf.string("result"), result);`

This way the Rascal type-checker can completely check the validity of the CommandExecutor again,
and the JSON-RPC bridge can not accidentally fail anymore.

If there exists code that wraps CommandExecutor results differently, e.g. like so:
`return [result];`

then this PR breaks those language servers. Therefore it would be good for @davylandman to have a close
look at this before we merge it.


jurgenvinju:
Also after this PR, if we don't replace `return ("result": true)` by `return true;` then the code also breaks on the client side that processes command results.

jurgenvinju:
There is a big TODO in this code that would break the client code on the TypeScript side even more. So let's push that into the future. It is also not strictly necessary.

DavyLandman:
I think something went wrong with this PR, it has a lot of commits in here? I think from the code-actions branch?

jurgenvinju:
Yes I branched off the wrong branch. Let's wait and merge the other one to
main first; then I'll rebase these changes.

Jurgen Vinju
  http://jurgen.vinju.org
  CWI SWAT
  TU Eindhoven
  Swat.engineering BV

jurgenvinju:
rewrote this on another PR.

PR #453, opened 2024-09-24
Update rascal.tmGrammar.json with generated grammar (commit 6106209 o…

sungshik:
Original: https://github.com/SWAT-engineering/rascal-textmate/blob/5eda17f209ae6b65924ddf27a77261b49c33880d/vscode-extension/syntaxes/rascal.tmLanguage.json

Changes wrt previous version:

  - Add support for multi-line tags
  - Remove support for multi-line strings -- This is to avoid file-long misclassification of non-string tokens as string tokens, which could result in a bit of a jarring user experience (e.g., introduce a parse error while writing some code on line 635; lose semantic highlighting; TextMate tokenizer misclassifies `>` on line 11 as string interpolation; the whole screen around line 635 suddenly turns orange)

### DIFF
--- a/rascal-vscode-extension/syntaxes/rascal.tmGrammar.json
+++ b/rascal-vscode-extension/syntaxes/rascal.tmGrammar.json
@@ -42,17 +42,8 @@
         }
       }
     },
-    "/inner/single/concretepart.bq": {
-      "match": "(\\\\\\`)",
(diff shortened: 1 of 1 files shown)


DavyLandman:
Could you in the description of the PR add a small list of changes/improvements to the current textmate?

PR #454, opened 2024-09-24
Remove lib schema and use new path config construction

jurgenvinju:
* [x] all uses of lib scheme replaced by PathConfig.resolveProjectInClasspath
* [x] any damage from  fixes in Rascal that have not been tested in rascal-lsp yet
* [x] fixes in the rascal project triggered by these tests
* [ ] make sure errors during configuration are registered as diagnostics at the right time and places
* [x] testing
* [ ] Remove dependence on project scheme for LSP interactions.


rascal-lsp/pom.xml
jurgenvinju, line 1: Why is enableStandardLibrary set to false again here?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/LSPTerminalREPL.java
```diff
@@ -104,25 +104,21 @@ protected IDEServices buildIDEService(PrintWriter err, IRascalMonitor monitor, T
     @Override
     protected Evaluator buildEvaluator(Reader input, PrintWriter stdout, PrintWriter stderr, IDEServices services) {
         var evaluator = super.buildEvaluator(input, stdout, stderr, services);
-        evaluator.addRascalSearchPath(URIUtil.correctLocation("lib", "rascal-lsp", ""));
-
-        URIResolverRegistry reg = URIResolverRegistry.getInstance();
+        try {
+            ISourceLocation projectDir = PathConfig.inferProjectRoot(URIUtil.createFileLocation(System.getProperty("user.dir")));
```
jurgenvinju, line 108: Add a comment that this user.dir business is due to the fact that we start the terminal process with the project folder as user dir.

Complexer code can be replaced by root location("cwd")

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/LSPTerminalREPL.java
```diff
@@ -104,25 +104,21 @@ protected IDEServices buildIDEService(PrintWriter err, IRascalMonitor monitor, T
...
+            reg.registerLogical(new ProjectURIResolver(services::resolveProjectLocation));
+            reg.registerLogical(new TargetURIResolver(services::resolveProjectLocation));
 
-        reg.registerLogical(new ProjectURIResolver(services::resolveProjectLocation));
-        reg.registerLogical(new TargetURIResolver(services::resolveProjectLocation));
+            debugServer = new DebugSocketServer(evaluator, (TerminalIDEClient) services);
 
-        debugServer = new DebugSocketServer(evaluator, (TerminalIDEClient) services);
+            var lspJar = PathConfig.resolveProjectOnClasspath("rascal-lsp");
+            evaluator.addRascalSearchPath(lspJar);
```
jurgenvinju, line 117: Duplicate on line 134

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/LSPTerminalREPL.java
```diff
@@ -133,26 +129,31 @@ protected Evaluator buildEvaluator(Reader input, PrintWriter stdout, PrintWriter
...
 
-            ClassLoader cl = new SourceLocationClassLoader(
-                pcfg.getClassloaders()
-                    .append(URIUtil.correctLocation("lib", "rascal",""))
-                    .append(URIUtil.correctLocation("lib", "rascal-lsp",""))
-                    .append(URIUtil.correctLocation("target", projectName, "")),
-                ClassLoader.getSystemClassLoader()
-            );
-
+            ClassLoader cl = new SourceLocationClassLoader(pcfg.getLibsAndTarget(), ClassLoader.getSystemClassLoader());
```
jurgenvinju, line 151: Pcfg.messages must be printed or registered as diagnostics or both.

rascal-lsp/src/main/rascal/lang/rascal/tests/rename/TestUtils.rsc
```diff
@@ -203,7 +203,7 @@ bool testRename(str stmtsStr, int cursorAtOldNameOccurrence = 0, str oldName = "
 public PathConfig getTestPathConfig(loc testDir) {
     return pathConfig(
         bin=testDir + "bin",
-        libs=[|lib://rascal|],
+        libs=[|std://rascal|],
```
jurgenvinju, line 206: That's not right. Std is only a source path for the interpreter. If we need the class files of the Rascal project here, for the java part of the stdlib, then use the functions to get a loc for the current Rascal project. I forget their name.

jurgenvinju: Srcs must contain std:// since the merge because it is not added by itself. PathConfig adds it and all other code that constructs pathConfig must do it too.

jurgenvinju (commented):
Some very small but decisive changes required. I think I added functions to util::Reflective to turn project names into absolute jar file paths already...  But if not then these have to be added.

jurgenvinju:
We have to get the pcfg.messages out in the open. I think for the terminal they should be printed using Messages.write. For the LSP evaluator I think the log feature should be used...

rascal-lsp/src/main/rascal/lang/rascal/tests/rename/TestUtils.rsc
jurgenvinju: I think it's not needed to add rascal.jar to the libs as the interpreter already runs in that jar and so the lib classes are loaded from there.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/LSPTerminalREPL.java
DavyLandman: Except that on `cwd` we cannot go to the parent? I was wondering the same, but I wanted to keep the code as close as possible to the original logic.

Fix that at a later point, if we have all this working.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/LSPTerminalREPL.java
DavyLandman: cleaned up.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/LSPTerminalREPL.java
DavyLandman: They are, a few lines higher (see `services.registerDiagnostics(pcfg.getMessages());`).

I've now also added a call to `Messages.write`

rascal-lsp/src/main/rascal/lang/rascal/tests/rename/TestUtils.rsc
DavyLandman: I think this is the result of an automatic search & replace

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/LSPTerminalREPL.java
jurgenvinju: Cwd is logical and will be resolved to a file uri before the parent is found. The code that checks the folder name to be equal to the project name takes care of this in createPathConfig

rascal-lsp/src/main/rascal/lang/rascal/tests/rename/TestUtils.rsc
jurgenvinju: Yes it certainly looks like that. Let's replace by the gerCurrentRascl for now

PR #457, opened 2024-09-27
Improving the UI test runtime

DavyLandman:
The UI test were taking too long to run, locally and on CI.

I dove in and started measure what was taking long, and bit by bit removing generic sleeps and replacing them by `driver.wait` calls. It also involved tuning memory and splitting of non UI tests to a separate stage.

Before (30min total time):
image

Now (13min total time):

image

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -75,7 +75,7 @@ jobs:
         working-directory: ./rascal-vscode-extension
         env:
           DELAY_FACTOR: 15
-          _JAVA_OPTIONS: '-Xmx5G'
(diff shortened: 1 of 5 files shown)


jurgenvinju:
Much appreciated, especially since my laptop can't be used for anything else during the wait!

jurgenvinju:
Just one more wait in macosx outline test to make longer.

DavyLandman:
> Just one more wait in macosx outline test to make longer.

It was already waiting 70s which should be enough for clicking on an outline item, what it needed was re-clicking, somehow we were too fast.

PR #464, opened 2024-10-03
Trying to improve a final issue with the UI tests

DavyLandman:
Sometimes on slow machines, it wouldn't work, and this is due to `findElement` throwing an exception that `driver.wait` doesn't swallow.

### DIFF
--- a/rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
+++ b/rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
@@ -166,7 +166,7 @@ describe('DSL', function () {
 
         // finds an open menu with the right item in it (Change to a) and then select
         // the parent that handles UI events like click() and sendKeys()
-        const menuContainer = await ide.hasElement(editor, By.xpath("//div[contains(@class, 'focused') and contains(@class, 'action')]/span[contains(text(), 'Change to a')]//ancestor::*[contains(@class, 'monaco-list')]"), Delays.verySlow,"The Change to a option should be available and focussed by default");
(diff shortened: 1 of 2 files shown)


rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
```diff
@@ -166,7 +166,7 @@ describe('DSL', function () {
 
         // finds an open menu with the right item in it (Change to a) and then select
         // the parent that handles UI events like click() and sendKeys()
-        const menuContainer = await ide.hasElement(editor, By.xpath("//div[contains(@class, 'focused') and contains(@class, 'action')]/span[contains(text(), 'Change to a')]//ancestor::*[contains(@class, 'monaco-list')]"), Delays.normal,"The Change to a option should be available and focussed by default");
+        const menuContainer = await ide.hasElement(editor, By.xpath("//div[contains(@class, 'focused') and contains(@class, 'action')]/span[contains(text(), 'Change to a')]//ancestor::*[contains(@class, 'monaco-list')]"), Delays.verySlow,"The Change to a option should be available and focussed by default");
```
toinehartman, line 169: Do we still need this longer delay, even with the improved "find element" implementation below?

toinehartman (approved):
Thanks for looking into this! LGTM.

rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
DavyLandman: I hope not, but it can't hurt. It will only take longer to crash in case it doesn't work. While if it's there, it will succeed within a second.

PR #469, opened 2024-10-04
Configuration of Rascal terminal names

sungshik:
### Context

This PR makes a small improvement to the VS Code extension that would make my life marginally easier 😉 I figured this might be useful to other users as well, so here we go.

### Problem

I have many modules that I need to import in separate terminals for testing purposes. (The modules share some conflicting definitions, so I can't import them in one terminal.) So:

 1. Open the modules:
    image

 2. Click this button in each of the modules:
    image

 3. Try to find a terminal for a specific module in `O(n)` time (terminals may appear out-of-order in this list):
    image

#### Abstract

Instead of giving a terminal the name of the project, give it the name of the module for which it was opened. (I realize this can also be done manually after a terminal is created, but that's a hassle with this many terminals.)

#### Concrete

This PR adds a setting to configure how Rascal terminal names should be formatted:

image

The default, `Project root`, gives the same behavior as before this PR. So, by default, nothing changes for users. Only when a user explicitly configures it to be `Module (qualified)` or `Module (unqualified)`, the name of the module for which a terminal was opened is inserted.

With `Module (qualified)`:

image

With `Module (unqualified)`:

image

### DIFF
--- a/rascal-vscode-extension/src/RascalExtension.ts
+++ b/rascal-vscode-extension/src/RascalExtension.ts
@@ -152,17 +152,20 @@ export class RascalExtension implements vscode.Disposable {
             const config = vscode.workspace.getConfiguration();
             const originFormat = config.get('rascal.terminal.name.originFormat');
             switch (originFormat) {
-                case 'Project root':
+                case 'Project root': {
                     const projectRoot = vscode.workspace.getWorkspaceFolder(uri);
                     if (projectRoot && projectRoot.name) {
                         return projectRoot.name;
                     }
                     break;
-                case 'Module':
+                }
+                case 'Module': {
                     const sep = uri.path.lastIndexOf("/");
                     if (sep !== -1 && uri.path.endsWith(".rsc")) {
                         return uri.path.substring(sep + 1, uri.path.length - 4);
                     }
+                    break;
+                }
                 default:
                     console.log(`Unknown origin format: ${originFormat}`);
             }
--- a/rascal-vscode-extension/src/RascalExtension.ts
+++ b/rascal-vscode-extension/src/RascalExtension.ts
@@ -136,7 +136,7 @@ export class RascalExtension implements vscode.Disposable {
                     shellPath: await getJavaExecutable(),
                     shellArgs: this.buildShellArgs(compilationPath, serverConfig, ...extraArgs),
                     isTransient: false, // right now we don't support transient terminals yet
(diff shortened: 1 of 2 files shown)


rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -148,6 +147,32 @@ export class RascalExtension implements vscode.Disposable {
...
+            switch (originFormat) {
+                case 'Project root': {
+                    const projectRoot = vscode.workspace.getWorkspaceFolder(uri);
+                    if (projectRoot && projectRoot.name) {
+                        return projectRoot.name;
+                    }
+                    break;
+                }
+                case 'Module': {
+                    const sep = uri.path.lastIndexOf("/");
```
DavyLandman, line 163: I would prefer to show the fully qualitfied name of the module, not just the last part. I think we have this calculated already?

DavyLandman:
Okay, for the people amongst us with repl-itus, I can see how this would help.

rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -148,6 +147,49 @@ export class RascalExtension implements vscode.Disposable {
...
+        return 'no project or module';
+    }
+
+    private qualifiedName: RegExp = (() => {
+        const name1 = '(?:[A-Z_a-z][0-9A-Z_a-z]*)';
+        const name2 = '(?:\\\\[A-Z_a-z][\\-0-9A-Z_a-z]*)';
+        const name = `(?:${name1}|${name2})`;
+        const qualifiedName = `(?:(?:${name}::)*(${name}))`;
+        return new RegExp(`^${qualifiedName}$`);
+    })(); // Build the regex only once
```
DavyLandman, line 191: If I remember correctly, a JavaScript regex cannot be reused.

But if you care for performance, a regex literal is a lot faster. I think since the jitter can decide to reuse the complication result.

sungshik: > If I remember correctly, a JavaScript regex cannot be reused.

Only when the `global`/`sticky` flag is set ([docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test))

DavyLandman: aah right :)

PR #473, opened 2024-10-11
A consistent naming scheme for the LanguageService constructors , with backward compatibility functions.

jurgenvinju:
This renames allmost all language services towards consistency.  

All verbs or all nouns did not work. 

Instead these are all proper adjectives to the word "service". E.g. "parsing service" and "references service" and "build service". _This is also the scheme of the original LSP services._ The new scheme is also consistent with the names in `Summary` of the cached output of the respective lazy services.

This changes:
* `parser` to `parsing` .. service
* `builder` to `build` .. service
* etc.

TODO's:
* [x] change service names to LSP 3.17 names, one-to-one
* [x] make new pico demo based on new names
* [x] add backward compatibility layer
* [x] fix the field and method and class names correspondingly in Java code
* [x] also apply renamings in Rascal code written for Rascal (e.g. Outline.rsc to DocumentSymbols.rsc)
* [x] test test test
* [x] check all documentation for the old names
* [x] inline the aliases for function profiles of the parameters of the LanguageService constructors (but leave the definitions for backward compatibility. Motivation: the aliases are opaque; you always have to look them up before you can define a language service.
* [x] merge with main
* [x] rename demo modules to make the new API the first choice.


rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -297,17 +298,57 @@ data LanguageService
...
+@synopsis{Construct a `parsing` LanguageService}
+LanguageService parser(Parser parser) = parsing(parser);
+
+@deprecated{Backward compatible with `analysis`}
+@synopsis{Construct a `analysis` LanguageService}
+LanguageService analyzer(Summarizer summarizer) = analysis(summarizer);
+
+@deprecated{Backward compatible with `build`}
+@synopsis{Construct a `build` LanguageService}
+LanguageService builder(Summarizer summarizer) = builder(summarizer);
```
DavyLandman, line 322: not renamed to build yet.

jurgenvinju: thx!

rascal-lsp/src/main/rascal/demo/lang/pico/NewLanguageServer.rsc
```diff
@@ -0,0 +1,187 @@
...
+ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+}
+module demo::lang::pico::NewLanguageServer
```
sungshik, line 27: Nit: Would it be possible to rename the old version to `...::LanguageServerDeprecatedNames` (or something along those line) and call the new version just `...::LanguageServer`?

rascal-lsp/src/main/rascal/demo/lang/pico/NewLanguageServer.rsc
```diff
@@ -0,0 +1,187 @@
...
+* You can run each contribution on an example in the terminal to test it first.
+Any feedback (errors and exceptions) is faster and more clearly printed in the terminal.
+}
+void main() {
+    registerLanguage(
+        language(
+            pathConfig(),
+            "Pico",
+            {"pico", "pico-new"},
+            "demo::lang::pico::LanguageServer",
```
sungshik, line 174:
```suggestion
            "demo::lang::pico::NewLanguageServer",
```

rascal-lsp/src/main/rascal/demo/lang/pico/NewLanguageServer.rsc
```diff
@@ -0,0 +1,187 @@
...
+            "picoLanguageContributor"
+        )
+    );
+    registerLanguage(
+        language(
+            pathConfig(),
+            "Pico",
+            {"pico", "pico-new"},
+            "demo::lang::pico::LanguageServer",
```
sungshik, line 183:
```suggestion
            "demo::lang::pico::NewLanguageServer",
```

rascal-lsp/src/main/rascal/demo/lang/pico/NewLanguageServer.rsc
```diff
@@ -0,0 +1,187 @@
...
+
+@synopsis{Provides each contribution (IDE feature) as a callback element of the set of LanguageServices.}
+set[LanguageService] picoLanguageContributor() = {
+    parsing(parser(#start[Program])),
+    documentSymbol(picoOutliner),
+    codeLense(picoLenses),
+    execution(picoCommands),
+    inlayHint(picoHinter),
+    definition(lookupDef),
+    codeAction(picoActions)
```
sungshik, line 43: I imagine Rascal users may look at this module for inspiration when they want to implement their own language server (to the extent of just copy-pasting the whole thing and find-replacing `pico` with `myLanguage`). If so, then I suppose the code in this module should reflect what we believe is good coding style for language servers. Long story short, maybe we should rename the arguments to be more consistent with the names of the constructors.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/model/RascalADTs.java
```diff
@@ -33,18 +33,18 @@ public class RascalADTs {
...
-        public static final String REFERRER = "referrer";
-        public static final String IMPLEMENTER = "implementer";
-        public static final String CODE_ACTION_CONTRIBUTOR = "actions";
+        public static final String INLAY_HINTER = "inlayHint";
+        public static final String COMMAND_EXECUTOR = "execution";
+        public static final String DOCUMENTER = "hover";
+        public static final String DEFINER = "definition";
+        public static final String REFERRER = "reference";
+        public static final String IMPLEMENTER = "implementation";
+        public static final String CODE_ACTION_CONTRIBUTOR = "codeAction";
```
sungshik, line 47: Should we rename the constants, too? I realize it's painful and invasive (iirc, there are also various methods that have these names), and for the Rascal *user* it doesn't matter. But for the Rascal *contributor* (especially new ones), it's nicer to work with a single style of names. I worry this may keep haunting us (not terribly, but still) if we don't unify as best as we can now. (I'd be happy to more actively contribute here btw, if you need the help.)

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -125,204 +109,132 @@ of the user.
...
-    | actions(CodeActionContributor actions)
+    | documentSymbol(list[DocumentSymbol] (Tree _input) labeler)
+    | codeLense(lrel[loc src, Command lens] (Tree _input) detector)
+    | inlayHint(list[InlayHint] (Tree _input) hinter)
+    | execution(value (Command _command) executor)
+    | hover(set[str] (Focus _focus) documentor)
+    | definition(set[loc] (Focus _focus) linker)
+    | reference(set[loc] (Focus _focus) linker)
+    | implementation(set[loc] (Focus _focus) linker)
+    | codeAction(list[CodeAction] (Focus _focus) actions)
```
sungshik, line 197: Nit: Also rename arguments to be consistent with the contructor? For instance, `linker` as a term doesn't seem to occur anywhere else in this module (not in the docs), so it's not so clear which abstraction/model it belongs to. I'd have a mild preference for `definitionService`, `definitionCallback`, `fDefinition`, `definitionFn`, or something along those lines.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -125,204 +109,132 @@ of the user.
...
-    | referrer(FocusReferrer reference)
-    | implementer(FocusImplementer implementer)
-    | actions(CodeActionContributor actions)
+    | documentSymbol(list[DocumentSymbol] (Tree _input) labeler)
+    | codeLense(lrel[loc src, Command lens] (Tree _input) detector)
+    | inlayHint(list[InlayHint] (Tree _input) hinter)
+    | execution(value (Command _command) executor)
+    | hover(set[str] (Focus _focus) documentor)
+    | definition(set[loc] (Focus _focus) linker)
+    | reference(set[loc] (Focus _focus) linker)
```
sungshik, line 195: Nit: The corresponding request in LSP is `textDocument/references` (plural)

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -125,204 +109,132 @@ of the user.
...
-    | lenses(OrderedLensDetector detector)
-    | inlayHinter(InlayHinter hinter)
-    | executor(CommandExecutor executor)
-    | documenter(FocusDocumenter documenter)
-    | definer(FocusDefiner definer)
-    | referrer(FocusReferrer reference)
-    | implementer(FocusImplementer implementer)
-    | actions(CodeActionContributor actions)
+    | documentSymbol(list[DocumentSymbol] (Tree _input) labeler)
+    | codeLense(lrel[loc src, Command lens] (Tree _input) detector)
```
sungshik, line 190: Nit: The corresponding request in LSP is `textDocument/codeLens` (without "e")

sungshik (commented):
Great to make this more consistent! I added a few comments, but essentially, my two remaining questions are these:

  - Should we also rename arguments to make them consistent with constructors? (e.g., the argument of `hover` is now `documentor`)
  - Should we also rename *everything* (well...) in the Java code?

I don't have strong opinions on these questions; just want to raise them for consideration. See the comments for details.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -125,204 +109,132 @@ of the user.
...
-}
-@pitfalls{
-* should be extremely fast in order to provide interactive access.
-* careful use of `@memo` may help to cache dependencies, but this is tricky!
-}
-@deprecated{The ((FocusDocumenter)) has replaced this type.}
+@deprecated{Only in use in deprecated functions}
 alias Documenter = set[str] (loc _origin, Tree _fullTree, Tree _lexicalAtCursor);
 
-@synopsis{Function profile for documentation contributions to a language server}
```
DavyLandman, line 157: is this great documentation now lost? is there no place to keep these around?  

(same for all the other details that was present in the aliases below)

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -125,204 +109,132 @@ of the user.
...
-* ((lenses)) discovers places to add "lenses" (little views embedded in the editor on a separate line) and connects commands to execute to each lense
-* ((inlayHinter)) discovers plances to add "inlays" (little views embedded in the editor on the same line). Unlike ((lenses)) inlays do not offer command execution.
-* ((executor)) executes the commands registered by ((lenses)) and ((inlayHinter))s.
+   * a ((hover)) service is a fast and location specific version of the `documentation` relation in a ((Summary)).
+   * a ((definition)) service is a fast and location specific version of the `definitions` relation in a ((Summary)).
+   * a ((reference)) service is a fast and location specific version of the `references` relation in a ((Summary)).
+   * an ((implementation)) service is a fast and location specific version of the `implementations` relation in a ((Summary)).
+* The ((documentSymbol)) service maps a source file to a pretty hierarchy for visualization in the "outline" view and "symbol search" features.
+* The ((lenses)) service discovers places to add "lenses" (little views embedded in the editor on a separate line) and connects commands to execute to each lense
```
DavyLandman, line 169: it's `codeLens` not `lenses`.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
DavyLandman: Not a nit I think, it should be spelled `codeLens`. [Lense is not an English word](https://www.yourdictionary.com/articles/lense-vs-lens-difference).

DavyLandman (commented):
I think we're getting close to a new set of names, but still require some tweaking.

rascal-lsp/src/main/rascal/demo/lang/pico/NewLanguageServer.rsc
jurgenvinju: Such a renaming would not be backward compatible.

jurgenvinju: Otherwise the tests explode and also take too long imho

jurgenvinju: We really need the backwards compatibility layer to be tested at every push still; and that layer utilizes the new API so we have full test coverage with the current situation.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/model/RascalADTs.java
jurgenvinju: Using java rename this shouldn't be too hard. Let's do it.

rascal-lsp/src/main/rascal/demo/lang/pico/NewLanguageServer.rsc
jurgenvinju: Agreed, we should rename them and use some of the old words in the synopsis and descriptions. E.g. documentSymbol should mention the word outline. 

This stuff is also still inspiration for people who are porting rascal eclipse plugins.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
jurgenvinju: Some of it has been moved to the LanguageServer documentation where it belongs. That can be extended more; I'm thinking of a table with a row per service.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
jurgenvinju: Good catch!

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
jurgenvinju: I've been having trouble with these names since forever. They are inconsequential because we never actually need the tags to get the function out. 

I propose we drop all those names. The constructor name documents the intent as it is for 100%.

Also we can't use a single name like "callback" because different types must have different names.

jurgenvinju: In general we need a Rascal coding standard for naming callback functions/closures.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/model/RascalADTs.java
jurgenvinju: @sungshik I welcome your help here. I'll be fixing stuff **/LanguageServer.rsc mostly.

rascal-lsp/src/main/rascal/demo/lang/pico/NewLanguageServer.rsc
DavyLandman: > Such a renaming would not be backward compatible.

We can update the test to use the old files, but if anyone finds the module (like in a link from somewhere) it would be nice if it's the newest description of how you should write down the file. 

Keeping around an outdated example might confuse more people, especially if it's not clearly marked as outdated.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
jurgenvinju: ok I dropped the names

rascal-lsp/src/main/rascal/demo/lang/pico/NewLanguageServer.rsc
jurgenvinju: I agree, but backwards compatibility is more important right now. We can clean up after we clean up the deprecated API. Just wait one release. This module is marked @deprecated with a link to the new module right at the top.

jurgenvinju: Don't forget that there are users of the old specific module, people have based their course material on this stuff.

DavyLandman: Can you elaborate on the backwards compatibility of a demo module? My understanding is that should/can import this module in their rascal code? I've always seen this as documentation, that we happen to also use in our UI tests.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/model/RascalADTs.java
jurgenvinju: pls pull though. there was some unfinished business in the constructor names

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
jurgenvinju: fixed! well spotted

jurgenvinju:
Converted back to draft while finishing the latest improvements and testing more.

rascal-lsp/src/main/rascal/demo/lang/pico/NewLanguageServer.rsc
jurgenvinju: We've asked all course owners at four universities; they've no problem with the proposed changes. This is the last thing to do before we merge (and the tests work)

jurgenvinju: done

PR #475, opened 2024-10-15
Debounce the parser

### LINKED ISSUE #358: Races in TextDocumentState
**Describe the bug**

The fields in [TextDocumentState](https://github.com/usethesource/rascal-language-servers/blob/3051c6b95157a563f6fba1517127a580ed7a68a9/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java) are volatile but not race-free. For instance, in method `update`:

https://github.com/usethesource/rascal-language-servers/blob/3051c6b95157a563f6fba1517127a580ed7a68a9/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java#L63-L67

When two threads execute this method concurrently (e.g., as part of two firings of `didChange`), the following race can happen:

```java
currentContent = text1; // by thread 1
currentContent = text2; // by thread 2
currentTree = newContents(text2); // by thread 2
currentTree = newContents(text1); // by thread 1
// at this point, currentContent and currentTree are inconsistent
```

Depending on usage patterns, the reads of these fields (via the getters) might need synchronization as well.

sungshik:
This PR adds optional debouncing to calls to get the parse tree from a `TextDocumentState`. Specifically, there are now two functions to get a parse tree:

  - `getTreeAsync()` returns a future for the parse tree of the current version of the document (current == the moment `getTreeAsync` is called), and the parser is immediately started. This is the existing behavior (i.e., the API is backward-compatible).
  - `getTreeAsync(delay)` returns a future for the parse tree of the current version of the document *or a future version*, but the parser is started only after a delay. Upon completion, the future provides the parse tree of the version of the document that was current when the parser was started.

The documentation in class `TextDocumentState` provides further details.

This PR also fixes #358.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java
@@ -200,7 +200,7 @@ class Debouncer<T> {
     // A debouncer is implemented using a *delayed executor* as `scheduler`. The
     // idea is to *periodically* check the state of the underlying resource
     // provider. More precisely, each time when the resource is requested,
-    // immediately check if case 1 or case 3 (above) are applicable. If so,
+    // immediately check if case 2 or case 3 (above) are applicable. If so,
     // return. If not, schedule a *delayed future* to retry the request, to be
     // completed after a small `period` (e.g., 50 milliseconds).
     //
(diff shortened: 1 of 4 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java
```diff
@@ -49,75 +59,253 @@
...
+            .thenCompose(Function.identity());
+
+        // Compute a new stamp
+        var delayRemaining = Math.max(oldStamp, delay);
+        var newStamp = delayRemaining - period;
+
+        // If the underlying resource provider is initialized, then return the
+        // future to get the resource
+        var future = getIfInitialized.get();
+        if (future != null && scheduled.weakCompareAndSet(oldRef, null, oldStamp, 0)) {
```
PieterOlivier, line 276: From the documentation of `AtomicStampedReference.weakCompareAndSet`:
"May fail spuriously and does not provide ordering guarantees, so is only rarely an appropriate alternative to compareAndSet."
So why use the weak variant? The "May fail spuriously" part scares me ;-)

sungshik: Ah, good catch! I misremembered the docs. It should be "strong" `compareAndSet` indeed.

PR #478, opened 2024-10-16
API and implementation of codefixes and code actions for Rascal itself

jurgenvinju:
* [x] lang::rascal::lsp::Actions module contains:
   * [x] Command definitions
   * [x] rascalCodeActions functions to detect code actions from a focus
   * [x] rascalExecutor for executing commands
* [x] factor common code for code actions from ParameterizedLSP in common class
* [x] introduce command execution to Rascal document service
* [x] introduce codeAction LSP message
   * [x] merge in quickfixes from Diagnostics
   * [x] call Actions module for more actions
* [x] testing manually 
* [x] adding test code
* [x] think of a way to expose `Command` and `CodeAction` definitions to rascal-core, such that the checker can annotate messages with quickfixes
* [x] try to remove more clones by factoring actions streaming code with function parameters.

https://github.com/user-attachments/assets/77d1a5a2-5109-4bee-a27b-63656d6b81dd


jurgenvinju:
@DavyLandman @toinehartman this is still a draft but if you are interested and perhaps have feedback... that would be nice.

I'm trying to prepare us for the inevitable removed of `Tree@\loc`, and some quickfixes and code actions in the editor to help our users and ourselves with this would be really useful. The DSL part of this follow the exact same design on purpose.

jurgenvinju:
* Typically the type-checker would attach CodeActions to error messages, for quick-fix purposes
* And for more general tools/refactorings/visualizations:
   * visualize import/extend graph
   * visualize call graph
   * convert curly body function to expression function
   * add parameter to all overloads
   * convert switch to overloaded function
   * ... etc ... the possibilities are endless :-)

rascal-lsp/src/main/rascal/lang/rascal/lsp/Actions.rsc
```diff
@@ -0,0 +1,106 @@
...
+    grammar = [trim("<i>") | i <- h.imports, i is \syntax];
+
+    newHeader = "<for (i <- sort(extends)) {><i>
+                '<}>
+                '<for (i <- sort(imports)) {><i>
+                '<}>
+                '<for (i <- grammar) {><i>
+                '<}>";
+
+    applyDocumentsEdits([changed(h@\loc.top, [replace(h.imports@\loc, newHeader)])]);
```
toinehartman, line 104: Why does this directly call `applyDocumentEdits` while the top-level rewrite on line 79 returns an `action(..., kind = refactor())`?

toinehartman (commented):
Nice improvement! I like how this allows to write transformations and other code actions relatively easily. I am curious how you'll tackle contribution code actions from the type checker!

rascal-lsp/src/main/rascal/lang/rascal/lsp/Actions.rsc
jurgenvinju: It's a demo of the possibilities. There is no semantic difference other than a different time of synchronization. The user does not see a difference. Both ways are applied in the editor.

jurgenvinju: However, there are reasons for choosing a command over edits, typically if the amount of edits is very large or expensive to compute, then the action edits will block the creation of the actions menu. In that case it is always better to register a command that will compute the edits later.

jurgenvinju: other reasons are if you need other side-effects next to the edits, then a command is useful. If you pick both, say an action with both edits and a command, then the edits will always be executed before the command is executed.

jurgenvinju:
> Nice improvement! I like how this allows to write transformations and other code actions relatively easily. I am curious how you'll tackle contribution code actions from the type checker!

I was planning on learning that from you @toinehartman

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -100,6 +107,8 @@ public RascalLanguageServices(RascalTextDocumentService docService, BaseWorkspac
         documentSymbolEvaluator = makeFutureEvaluator(exec, docService, workspaceService, client, "Rascal document symbols", monitor, null, false, "lang::rascal::lsp::DocumentSymbols");
         semanticEvaluator = makeFutureEvaluator(exec, docService, workspaceService, client, "Rascal semantics", monitor, null, true, "lang::rascalcore::check::Summary", "lang::rascal::lsp::refactor::Rename");
         compilerEvaluator = makeFutureEvaluator(exec, docService, workspaceService, client, "Rascal compiler", monitor, null, true, "lang::rascalcore::check::Checker");
+        actionEvaluator = makeFutureEvaluator(exec, docService, workspaceService, client, "Rascal actions", monitor, null, true, "lang::rascal::lsp::Actions");
```
DavyLandman, line 110: I'm not a fan of yet another evalutor getting added, this creates a bigger memory pressure. It also slows down initial startup, as they are all loaded in parallel when VS Code starts. 

Can't this one go into the `semanticEvaluator`? I mean, it's a balance, since `Summary` and especially `Rename` can also take quite a while, so it would mean the actions waiting in the queue.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -250,6 +261,49 @@ public List<CodeLensSuggestion> locateCodeLenses(ITree tree) {
...
+        ), exec);
+    }
+
+    private CompletableFuture<IConstructor> parseCommand(String command) {
+        return actionStore.thenApply(commandStore -> {
+            try {
+                return (IConstructor) new StandardTextReader().read(VF, commandStore, commandStore.lookupAbstractDataType("Command"), new StringReader(command));
+            }
+            catch (FactTypeUseException | IOException e) {
+                logger.catching(e);
```
DavyLandman, line 301: whe here a `logger.catching` but not in the `parseCodeActions`?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -281,6 +334,13 @@ public String getCommandName() {
...
             return shortName;
         }
+    }
 
+    public InterruptibleFuture<IList> codeActions(IList focus, PathConfig pcfg) {
+        return runEvaluator("Rascal codeActions", actionEvaluator, eval -> {
+            Map<String,IValue> kws = Map.of("pcfg", pcfg.asConstructor());
+            return (IList) eval.call("rascalCodeActions", "lang::rascal::lsp::Actions", kws, focus);
+        },
+        null, exec, false, client);
```
DavyLandman, line 344: Can we instead of `null`  return an empty list?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -250,6 +261,49 @@ public List<CodeLensSuggestion> locateCodeLenses(ITree tree) {
...
+    public InterruptibleFuture<IValue> executeCommand(String command) {
+        logger.debug("executeCommand({}...) (full command value in TRACE level)", () -> command.substring(0, Math.min(10, command.length())));
+        logger.trace("Full command: {}", command);
+
+        return InterruptibleFuture.flatten(parseCommand(command).thenApply(cons ->
+            EvaluatorUtil.<IValue>runEvaluator(
+                "executeCommand",
+                actionEvaluator,
+                ev -> ev.call("evaluateRascalCommand", cons),
+                null,
```
DavyLandman, line 287: can we return some empty `IValue`? `Command::noop` I'm guessing? instead of null?

rascal-lsp/src/main/rascal/lang/rascal/lsp/Actions.rsc
```diff
@@ -0,0 +1,156 @@
...
+        result += [action(command=visualImportGraphCommand(pcfg), title="Visualize project import graph")]
+               +  [action(command=sortImportsAndExtends(h), title="Sort imports and extends")]
+               ;
+    }
+
+    return result;
+}
+
+@synopsis{Add a license header if there isn't one.}
+list[CodeAction] addLicenseAction(start[Module] \module, PathConfig pcfg) {
```
DavyLandman, line 73: not everyone wants to put a license header in their source code. Some licenses allow for only a license in the root of the directory. Can a user disable these in some way?

rascal-lsp/src/main/rascal/lang/rascal/lsp/Actions.rsc
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    return "";
+}
+
+private DocumentEdit makeLicenseEdit(loc \module, str license)
+    = changed(\module.top, [replace(\module.top(0, 0), license)]);
+
+@synopsis{Rewrite immediate return to expression.}
+list[CodeAction] toplevelCodeActions(Toplevel t:
```
DavyLandman, line 113: nice one 👍🏼

rascal-lsp/src/main/rascal/lang/rascal/lsp/Actions.rsc
```diff
@@ -0,0 +1,156 @@
...
+        result += addLicenseAction(top, pcfg);
+    }
+
+    if ([*_, Toplevel t, *_] := focus) {
+        result += toplevelCodeActions(t);
+    }
+
+    if ([*_, Header h, *_] := focus) {
+        result += [action(command=visualImportGraphCommand(pcfg), title="Visualize project import graph")]
+               +  [action(command=sortImportsAndExtends(h), title="Sort imports and extends")]
```
DavyLandman, line 65: we should only offer this if the header is more than 1 import?

DavyLandman (changes requested):
I like how most of it looks 👍🏼 great reuse of the existing code-actions added the previous PR. My only concern is with the extra evaluator.

rascal-lsp/src/main/rascal/lang/rascal/lsp/Actions.rsc
jurgenvinju: It shows all modules in the entire project always, so it is not that specific to the current module.

rascal-lsp/src/main/rascal/lang/rascal/lsp/Actions.rsc
jurgenvinju: You can simply not choose to activate quickfix? If my memory does not fail me, the fact that licenses can only be in the root of the folder is less depending on the license and more on the locality and the scope of the license. American law is quite different from European law, and anglosaxon law from continental European law. 

In any case, the prudent lawyer would advice you to put it everywhere if you (a) don't want to be liable for the contents if someone copies an entire file without the project contents and (b) want the copier of the file to be liable to the conditions made in the license. 

In that sense this action is useful to anybody who wishes to be on the safe side.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
jurgenvinju: I'll have a look. We can return `("result":false)` for backward compatibility.

jurgenvinju: But in reality this is dead since that's what `evaluateRascalCommand` does already.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
jurgenvinju: check! Although this can never happen since `rascalCodeActions` returns the empty list by default.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
jurgenvinju: right, because the failure would have occurred there already earlier. So we fail hard there as soon as possible, and softer here.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
jurgenvinju: I think that would be acceptible. Currently my machine is only a bit sooner to present the actions than to present the summary. And you are right there will be a large overlap between summary code and actions code eventually anyway, so leading to memory duplication that is unwarranted.

rascal-lsp/src/main/rascal/lang/rascal/lsp/Actions.rsc
jurgenvinju: I've restricted the action to only appear if the cursor is on the QualifiedName part of the header (the module name). That way it is always within reach but not annoyingly so.

jurgenvinju:
I think this is ready if you do to. Note that this does not test and quickfixes attached to messages from the checker, yet. That's because the checker does not produce them yet. I plan to add a test here anway, as soon as one of those quickfixes starts popping up by the checker. But that's after merging IMHO.

jurgenvinju:
> I like how most of it looks 👍🏼 great reuse of the existing code-actions added the previous PR. My only concern is with the extra evaluator.

I fixed the evaluator but the PR still thinks this is open? Help, where do I click?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
```diff
@@ -415,6 +428,50 @@ public CompletableFuture<List<? extends CodeLens>> codeLens(CodeLensParams param
...
+    @Override
+    public CompletableFuture<List<Either<Command, CodeAction>>> codeAction(CodeActionParams params) {
+        logger.debug("codeAction: {}", params);
+
+        final var loc = Locations.toLoc(params.getTextDocument());
+        final var start = params.getRange().getStart();
+        // convert to Rascal 1-based line
+        final var startLine = start.getLine() + 1;
+        // convert to Rascal UTF-32 column width
+        final var startColumn = columns.get(loc).translateInverseColumn(start.getLine(), start.getCharacter(), false);
```
DavyLandman, line 440: I'm positive I've seen this code before, can we reduce the duplication? especially as it's easy to mess this up.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
```diff
@@ -415,6 +428,50 @@ public CompletableFuture<List<? extends CodeLens>> codeLens(CodeLensParams param
...
+                .thenApply(IList::stream)
+            , () -> Stream.<IValue>empty())
+            ;
+
+        // final merging the two streams of commmands, and their conversion to LSP Command data-type
+        return CodeActions.mergeAndConvertCodeActions(this, "", BaseWorkspaceService.RASCAL_LANGUAGE, quickfixes, codeActions);
+    }
+
+    private CompletableFuture<IList> computeCodeActions(final int startLine, final int startColumn, ITree tree, PathConfig pcfg) {
+        IList focus = TreeSearch.computeFocusList(tree, startLine, startColumn);
```
DavyLandman, line 464: please move this calculation of the focus list into the completeable future, to move as much of the calculation to async task.

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
```diff
@@ -187,5 +187,28 @@ describe('IDE', function () {
...
+        await editor.moveCursor(1,8); // in the module name
+
+        try {
+            // trigger the code actions
+            const inputarea = await editor.findElement(By.className('inputarea'));
+            await inputarea.sendKeys(Key.chord(TextEditor.ctlKey, "."));
+
+            // finds an open menu with the right item in it (Change to a) and then select
+            // the parent that handles UI events like click() and sendKeys()
+            const menuContainer = await ide.hasElement(editor, By.xpath("//div[contains(@class, 'focused') and contains(@class, 'action')]/span[contains(text(), 'Add missing license header')]//ancestor::*[contains(@class, 'monaco-list')]"), Delays.normal, "Add-license action should be available and focused");
```
DavyLandman, line 202: we should refactor this into a function into the IDE class, with the text as an argument.

DavyLandman (approved):
Some small remarks after a second review. I've changed my vote to approve, such that if you've went through them, and dealt with them to your discretion, you can merge.

toinehartman (approved):

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
jurgenvinju: ok will do

jurgenvinju:
@toinehartman right. Maybe we could make a Position to Position method like: `Position fromLSPpositionToRascalPosition(Position p)` . It is fairly critical code with weird constants and everything.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
jurgenvinju: this code was already called from the asynhronous task on lin 452. but I created yet another async wrapper anyway. Don't know if that helps.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseWorkspaceService.java
```diff
@@ -108,11 +110,12 @@ public void didChangeWorkspaceFolders(DidChangeWorkspaceFoldersParams params) {
 
     @Override
     public CompletableFuture<Object> executeCommand(ExecuteCommandParams params) {
-        if (params.getCommand().startsWith(RASCAL_META_COMMAND)) {
+        if (params.getCommand().startsWith(RASCAL_META_COMMAND) || params.getCommand().startsWith(RASCAL_COMMAND)) {
```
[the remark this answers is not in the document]
jurgenvinju: Yes we are, because otherwise the LSP would break. In fact we were matching the wrong name here (and als passing the wrong name) and things got jumbled up between the two LSP servers.

jurgenvinju:
@toinehartman @DavyLandman we now have single-point-of-change for translations from LSP startLine/column to Rascal's lines and columns with `Locations::toRascalPosition` applied everywhere. I've search for `+ 1`.

toinehartman:
Thanks @jurgenvinju! Nice work. Looking forward to trying out the initial refactorings *and* use of this framework to add others.

PR #482, opened 2024-10-17
Rename Java identifiers to be consistent with LSP

sungshik:
**Note:** All tests actually succeed, but SonarCloud is unhappy about "new" code duplication in `LanguageContributionsMultiplexer` and `InterpretedLanguageContributions`. This duplication was already there, but because the lines have changed due to renaming, SonarCloud is triggered.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ILanguageContributions.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ILanguageContributions.java
@@ -46,49 +46,50 @@ import io.usethesource.vallang.type.TypeStore;
 public interface ILanguageContributions {
     public String getName();
 
-    public CompletableFuture<ITree> parseSourceFile(ISourceLocation loc, String input);
-    public InterruptibleFuture<IList> outline(ITree input);
-    public InterruptibleFuture<IConstructor> analyze(ISourceLocation loc, ITree input);
-    public InterruptibleFuture<IConstructor> build(ISourceLocation loc, ITree input);
-    public InterruptibleFuture<IList> lenses(ITree input);
-    public InterruptibleFuture<@Nullable IValue> executeCommand(String command);
-    public CompletableFuture<IList> parseCodeActions(String command);
-    public InterruptibleFuture<IList> inlayHint(@Nullable ITree input);
-    public InterruptibleFuture<ISet> documentation(IList focus);
-    public InterruptibleFuture<ISet> definitions(IList focus);
-    public InterruptibleFuture<ISet> references(IList focus);
-    public InterruptibleFuture<ISet> implementations(IList focus);
-    public InterruptibleFuture<IList> codeActions(IList focus);
+    public CompletableFuture<ITree>              runParsingService(ISourceLocation loc, String input);
+    public InterruptibleFuture<IConstructor>     runAnalysisService(ISourceLocation loc, ITree input);
+    public InterruptibleFuture<IConstructor>     runBuildService(ISourceLocation loc, ITree input);
+    public InterruptibleFuture<IList>            runDocumentSymbolService(ITree input);
+    public InterruptibleFuture<IList>            runCodeLensService(ITree input);
+    public InterruptibleFuture<IList>            runInlayHintService(@Nullable ITree input);
+    public InterruptibleFuture<@Nullable IValue> runExecutionService(String command);
+    public InterruptibleFuture<ISet>             runHoverService(IList focus);
+    public InterruptibleFuture<ISet>             runDefinitionService(IList focus);
+    public InterruptibleFuture<ISet>             runReferencesService(IList focus);
+    public InterruptibleFuture<ISet>             runImplementationService(IList focus);
+    public InterruptibleFuture<IList>            runCodeActionService(IList focus);
 
-    public CompletableFuture<Boolean> hasAnalyzer();
-    public CompletableFuture<Boolean> hasBuilder();
-    public CompletableFuture<Boolean> hasOutliner();
-    public CompletableFuture<Boolean> hasLensDetector();
-    public CompletableFuture<Boolean> hasInlayHinter();
-    public CompletableFuture<Boolean> hasCommandExecutor();
-    public CompletableFuture<Boolean> hasDocumenter();
-    public CompletableFuture<Boolean> hasDefiner();
-    public CompletableFuture<Boolean> hasReferrer();
-    public CompletableFuture<Boolean> hasImplementer();
-    public CompletableFuture<Boolean> hasCodeActionsContributor();
+    public CompletableFuture<IList> parseCodeActions(String command);
+
+    public CompletableFuture<Boolean> hasAnalysisService();
+    public CompletableFuture<Boolean> hasBuildService();
+    public CompletableFuture<Boolean> hasDocumentSymbolService();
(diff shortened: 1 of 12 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -689,14 +688,14 @@ public CompletableFuture<List<? extends Location>> references(ReferenceParams pa
...
-            lookup(ParametricSummary::documentation, params.getTextDocument(), params.getPosition())
+            lookup(ParametricSummary::hovers, params.getTextDocument(), params.getPosition())
             .thenApply(Hover::new)
             , () -> null);
     }
 
     @Override
     public CompletableFuture<List<FoldingRange>> foldingRange(FoldingRangeRequestParams params) {
-        logger.debug("textDocument/foldingRange: {}", params.getTextDocument());
+        logger.debug("Folding range: {}", params.getTextDocument());
```
sungshik, line 698: Nit: I've changed this to be consistent with other `debug` calls (above)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Outline.java
sungshik, line 1: I wanted to rename this class to `DocumentSymbols`, but then decided not to, to preserve git history.

jurgenvinju: I'm ok with a rename. Easier to grep for in the future when we're looking for the feature.

jurgenvinju:
Pls merge but consider renaming Outline.java

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Outline.java
DavyLandman: Agreed, git can follow renames if they happen in a commit that doesn't do too much other stuff.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ILanguageContributions.java
```diff
@@ -46,49 +46,50 @@
 public interface ILanguageContributions {
     public String getName();
 
-    public CompletableFuture<ITree> parseSourceFile(ISourceLocation loc, String input);
-    public InterruptibleFuture<IList> outline(ITree input);
-    public InterruptibleFuture<IConstructor> analyze(ISourceLocation loc, ITree input);
-    public InterruptibleFuture<IConstructor> build(ISourceLocation loc, ITree input);
-    public InterruptibleFuture<IList> lenses(ITree input);
-    public InterruptibleFuture<@Nullable IValue> executeCommand(String command);
+    public CompletableFuture<ITree>              runParsingService(ISourceLocation loc, String input);
```
DavyLandman, line 49: Do they all need to be postfixed with service? If so, how about renaming the class?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Outline.java
sungshik: Ok, done

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ILanguageContributions.java
sungshik: > Do they all need to be postfixed with service?

I appended `...Service` to align the names with `util::LanguageServer`. For instance, the `definition` constructor in the ADT is defined as follows:

https://github.com/usethesource/rascal-language-servers/blob/eae06cf098e620deceeef4f90e3be10db2095257/rascal-lsp/src/main/rascal/util/LanguageServer.rsc#L255

So, `definitionService` is a function of type `set[loc] (Focus _focus)`, which is then eventually run in an implementation of this method:

https://github.com/usethesource/rascal-language-servers/blob/eae06cf098e620deceeef4f90e3be10db2095257/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ILanguageContributions.java#L57

Consistency aside, alternative names `runDefinition` or just `definition` would make less sense semantically to me, but this is a matter of taste probably.

> If so, how about renaming the interface?

Do you have anything in particular in mind? `ILanguageServices`?

DavyLandman: I would say, if the interface is named `ILanguageServices` than `.runDefinition` make sense, and doesn't have to repeat `service` everytime.

Similar to how the `lsp4j` & the lsp terms work.

In general, I have a preference for not repeating a word on every method, even the `run` prefix makes me feel a overly verbose.

sungshik: Just to clarify, the rationale for these field names is:
  - Each field contains an asynchronous computation to `run...`* a function
  - Each of these functions is called `...Service` in the Rascal ADT

I understand why you mention the `lsp4j` interface, and it's related, but I feel the situations aren't completely comparable.

Anyway, the current names are optimized for descriptiveness and consistency, but not for size. I think if we optimize for size (remove the `run...Service` prefix and suffix), we do lose a bit of descriptiveness and consistency (even when the interface is called `ILanguageServices`). Not terribly so, but a bit nevertheless. Of course, shorter names have advantages, too, so I'm not necessarily against it; just thinking out loud.

(*) Actually, maybe `call...` would have been even better, but that's not the point you're trying to make 😉

DavyLandman: Yes, agreed, the lsp4j is not the exact same, but it was an example of how the names of the lsp actions are used as method names instead of wrapping them with prefixes/suffixes.

I think going from the (for example) original `outline` to `runDocumentSymbolService` is quite a verbose increment. Or different less-extreme example `codeLens`  to ` runCodeLensService`. So I appreciate the idea to align the names with the new names in the rascal file, but this is also changing the whole naming style.

Maybe something like `requestDocumentSymbol` is a compromise?

sungshik: Cool, ok, I think we basically agree on the principles 🙂 I feel the compromise is more "everybody-lose" than "everybody-win" in this situation 😉, though, so I think the short names would be better then (just `documentSymbol`).

jurgenvinju: Agreed. Also I had to add the field names only because of a bug inn the interpreter. The constructor names are exactly equal to the LSP names without "Service" postfix. I had to add those to distinguish from the constructor names. 

Hopefully we can remove those double names in the future. Nobody uses them anyway.

jurgenvinju:
Shall we merge this?

DavyLandman:
I think the consensus is that @sungshik will revert some of the name changes into more compact variant. After that, yes, let's merge.

sungshik:
Renamed to shorter names, CI succeeds (modulo SonarCloud's duplication mistake), ready to merge!

jurgenvinju:
Thanks @sungshik !

PR #490, opened 2024-10-24
Error recovery for Rascal

sungshik:
### History

  - A few [initial commits](https://github.com/usethesource/rascal-language-servers/compare/7d24fc7..2d5dc04) to do a first integration of error recovery
  - #475
  - #476
  - #491
  - #492
  - #512
  - #622
  - #632

### Building a VSIX package

Run the following script:
    
```sh
# Installing `rascal`
git clone https://github.com/usethesource/rascal.git
cd rascal
git checkout feat/error-recovery
mvn install -Drascal.compile.skip -Drascal.tutor.skip -DskipTests
RASCAL_VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
cd ..

# Building the VSIX package
git clone https://github.com/usethesource/rascal-language-servers.git
cd rascal-language-servers/rascal-lsp
git checkout error-recovery/rascal
mvn versions:use-dep-version -Dincludes=org.rascalmpl:rascal -DdepVersion=$RASCAL_VERSION -DforceVersion=true
cd ../rascal-vscode-extension
npm ci
cd ..
./build.sh -fd
./package.sh
cd ..

# Cleaning up
mv rascal-language-servers/rascal-vscode-extension/*.vsix .
yes | rm -r rascal rascal-language-servers
```


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java
```diff
@@ -45,64 +59,253 @@
...
+    private final Supplier<CompletableFuture<T>> initializeAndGet;
+
+    public Debouncer(Duration period,
+            Supplier<@Nullable CompletableFuture<T>> getIfInitialized,
+            Supplier<CompletableFuture<T>> initializeAndGet) {
+
+        this(Math.toIntExact(period.toMillis()), getIfInitialized, initializeAndGet);
+    }
+
+    public Debouncer(int period,
```
DavyLandman, line 237: Can we write some junit tests for these classes? Like to make sure it behaves as intended, and later, people don't accidentally break it?

rascal-lsp/src/main/rascal/lang/rascal/lsp/Outline.rsc
```diff
@@ -31,38 +31,58 @@ import String;
...
-        case (Declaration) `<Tags _> <Visibility _> tag <Kind k> <Name name> on <{Type ","}+ ts>;`:
-            children += [symbol(clean("<name>"), \key(), name@\loc, detail="tag <k> <name> on <ts>")];
+    children = [];
 
-        case (Declaration) `<Tags _> <Visibility _> data <UserType u> <CommonKeywordParameters kws>;`: {
-            kwlist = [symbol(".<k.name>", property(), k@\loc, detail="<k.\type>") | kws is present, KeywordFormal k <- kws.keywordFormalList];
-            children += [symbol("<u.name>", struct(), u@\loc, detail="data <u> <kws>", children=kwlist)];
+    top-down-break visit (m) {
+        case decl: (Declaration) `<Tags _> <Visibility _> <Type t> <{Variable ","}+ vars>;`:
+            if (!hasErrors(decl)) {
```
jurgenvinju: I'm missing something here. If a node like this matches then it has enough structure to produce the symbol. What does it matter if there are errors inside?

jurgenvinju: The point of the error trees is that they will be there but won't match with any of the expected language constructs. So you can ignore then typically.

For the case where you encounter one during recursion, a match on the error tree has to be added that ignores it or does something useful like a short yield: `"<error tree>"[..25]`

jurgenvinju: Yield of an error tree is guaranteed to work: "<myStat>" will yield the entire string that was accepted to fill in for a real statement.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java
```diff
@@ -45,64 +67,178 @@
...
+    private CompletableFuture<Update> parseAndGetCurrent(Duration delay) {
+        var update = current;
+        if (update.isParsing()) {
+            return CompletableFuture.completedFuture(update);
+        } else {
+            return parseAndGetCurrentDebouncer.get(delay);
+        }
+    }
+
+    private class Update {
```
DavyLandman, line 155: please add a small description to this class, since `Update` is so generic?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java
```diff
@@ -45,64 +67,178 @@
...
+                parser
+                    .apply(location, content)
+                    .whenComplete((t, e) -> {
+                        // Prepare result values for futures
+                        var tree = new Versioned<>(version, t);
+                        var diagnostics = new Versioned<>(version, toDiagnostics(t, e));
+
+                        // Complete future to get the tree
+                        if (t == null) {
+                            treeAsync.completeExceptionally(e);
```
DavyLandman, line 199: loooks like we should do this before building `tree` and `diagnostics`? or are those other functions able to deal with null trees?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -227,7 +221,8 @@ public void connect(LanguageClient client) {
     @Override
     public void didOpen(DidOpenTextDocumentParams params) {
         logger.debug("Did Open file: {}", params.getTextDocument());
-        handleParsingErrors(open(params.getTextDocument()));
+        TextDocumentState file = open(params.getTextDocument());
+        handleParsingErrors(file, file.getCurrentDiagnosticsAsync()); // No debounce
```
DavyLandman, line 225: what if we introduce 2 globals in this class?

```java
private static final Duration NO_DEBOUNCE = Duration.ZERO;
private static final Duration NORMAL_DEBOUNCE = Duration.ofMillis(800);
```

And we use them in the 2 calls? so that we don't need this comment?

(and share them in some way with `RascalTextDocumentService.java`)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -278,37 +273,16 @@ private void triggerBuilder(TextDocumentIdentifier doc) {
...
-                    new Range(new Position(0,0), new Position(0,1)),
-                    "Parsing failed: " + excp.getMessage(),
-                    DiagnosticSeverity.Error,
-                    "Rascal Parser");
-            }
-            logger.trace("Finished parsing tree, reporting new parse error: {} for: {}", newParseError, file.getLocation());
-            facts(file.getLocation()).reportParseErrors(file.getLocation(), version,
-                newParseError == null ? Collections.emptyList() : Collections.singletonList(newParseError));
-            return null;
+    private void handleParsingErrors(TextDocumentState file, CompletableFuture<Versioned<List<Diagnostic>>> diagnosticsAsync) {
```
DavyLandman, line 281: Observation: in this codebase we've tried to keep the LSP mapping in this class, and out of the other classes. For this function this has shifted to the `TextDocumentState` class. 

I'm not sure if maybe we should pass that mapping as an argument or something?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
```diff
@@ -52,11 +54,18 @@
...
 import io.usethesource.vallang.IString;
 import io.usethesource.vallang.IValue;
+import io.usethesource.vallang.IValueFactory;
 
 public class Diagnostics {
     private static final Logger logger = LogManager.getLogger(Diagnostics.class);
     private static final Map<String, DiagnosticSeverity> severityMap;
 
+    // Note: DiagnosticSeverity.Hint only highlightes a single character!
+    static DiagnosticSeverity errorLocationHighlight = DiagnosticSeverity.Error;
```
DavyLandman, line 64: are they meant to be package private and writable (also by other classes)? and why are some null? not to mention I don't like such a global state, what if we want to have it different for the rascal parser vs the DSL parsers?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
```diff
@@ -71,9 +80,64 @@ public static <K, V> Map<K, List<V>> groupByKey(Stream<Entry<K, V>> diagnostics)
     }
 
     public static Diagnostic translateDiagnostic(ParseError e, ColumnMaps cm) {
-        return new Diagnostic(toRange(e, cm), e.getMessage(), DiagnosticSeverity.Error, "parser");
+        var message = e.getMessage() + " (irrecoverable)";
```
DavyLandman, line 83: for DSLs, we don't know if error recovery is enabled, so this extra message part is not accurate.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
```diff
@@ -71,9 +80,64 @@ public static <K, V> Map<K, List<V>> groupByKey(Stream<Entry<K, V>> diagnostics)
...
+
+        if (prefixHighlight != null) {
+            // The recognized prefix
+            int prefixLength = skippedLoc.getOffset()-errorTreeLoc.getOffset();
+            if (prefixLength > 0) {
+                ISourceLocation prefixLoc = factory.sourceLocation(errorTreeLoc,
+                        errorTreeLoc.getOffset(), skippedLoc.getOffset()-errorTreeLoc.getOffset(),
+                        errorTreeLoc.getBeginLine(), skippedLoc.getBeginLine(),
+                        errorTreeLoc.getBeginColumn(), skippedLoc.getBeginColumn());
+                diagnostics.add(new Diagnostic(toRange(prefixLoc, cm), "Recovered parse error prefix", DiagnosticSeverity.Error, "parser"));
```
DavyLandman, line 122: why is this not using `prefixHighlight` variable?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
```diff
@@ -121,14 +185,27 @@ public static Diagnostic translateDiagnostic(IConstructor d, Range range) {
...
-                loc.getOffset(), loc.getLength() + 1,
-                loc.getBeginLine(), loc.getBeginColumn(),
-                loc.getEndLine(), loc.getEndColumn() + 1);
+            try {
+                loc = ValueFactoryFactory.getValueFactory().sourceLocation(loc,
+                    loc.getOffset(), loc.getLength() + 1,
+                    loc.getBeginLine(), loc.getBeginColumn(),
+                    loc.getEndLine(), loc.getEndColumn() + 1);
+            } catch (Throwable t) {
+                logger.trace("Cannot extend 0-width location for parse error: " + t.getMessage());
```
DavyLandman, line 205: Why do we catch `Throwable`? is there not a more specific exception to catch? In general we shouldn't be catching `Throwable` since that also catches `Error`. I think `Exception` is the most generic we should catch, but even then, is there no more specific one?

nit: log messages should use `"{}"` interpolation instead of string concat. Also there is an overload for log4j2 that allows you just pass in the whole exception I think.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/RascalServices.java
```diff
@@ -27,19 +27,39 @@
...
+        RascalParser parser = new RascalParser();
+        ITree tree = parser.parse(
+            Parser.START_MODULE, loc.getURI(), input,
+            new NoActionExecutor(),
+            new DefaultNodeFlattener<>(),
+            new UPTRNodeFactory(true),
+            new ToTokenRecoverer(loc.getURI(), parser, new StackNodeIdDispenser(parser)));
+
+        // Recover
+        ParseErrorRecovery recoverer = new ParseErrorRecovery(VALUE_FACTORY);
```
DavyLandman, line 62: `RascalLanguageServices` puts a single instance of this in a field? Why do we have to re-create it here?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/concurrent/DebouncedSupplier.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+import java.time.Duration;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public class DebouncedSupplier<T> {
```
DavyLandman, line 38: let's add a small note on this class to explain why it's there.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/concurrent/DebouncedSupplier.java
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Gets the result of `supplier` with the given timeout, as follows:
+     *   - if the next get *doesn't* happen before the given timeout, then the
+     *     current get completes with the result of `supplier` upon timeout;
+     *   - if the next get *does* happen before the given timeout, then the
+     *     current get completes with the result of the next get upon
+     *     availability.
+     */
+    public CompletableFuture<T> get(long timeout, TimeUnit unit) {
```
DavyLandman, line 70: This looks a lot like the `InterruptableFuture` design. Can we merge these? Or maybe use them here?

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
```diff
@@ -36,17 +36,22 @@ module demo::lang::pico::LanguageServer
 import util::LanguageServer;
 import util::IDEServices;
 import ParseTree;
+import util::ParseErrorRecovery;
 import util::Reflective;
 import lang::pico::\syntax::Main;
 
+private Tree picoParser(str input, loc origin) {
+    return parse(#start[Program], input, origin, allowRecovery=true, filters={createParseErrorFilter(false)});
```
DavyLandman, line 44: is there no way to use the `parser` function to build a function that does this?

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
```diff
@@ -36,17 +36,22 @@ module demo::lang::pico::LanguageServer
...
+}
+
 @synopsis{A language server is simply a set of ((LanguageService))s.}
 @description{
 Each ((LanguageService)) for pico is implemented as a function.
 Here we group all services such that the LSP server can link them
 with the ((Language)) definition later.
 }
 set[LanguageService] picoLanguageServer() = {
```
DavyLandman, line 53: since this one is also used for the other tests, I would like to think of a way to still test that without error recovery on, things also keep working.

DavyLandman (commented):
Quite some work and a impressive feature. I know this has had quite some iterations. So I've reviewed it in a way to say: so where did we end up, and what do we like about that, and what should we maybe polish a bit more.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/concurrent/DebouncedSupplier.java
sungshik: I'm not sure. The debounce logic is non-trivial I think (and, e.g., crucially relies on `completeOnTimeout`, which `InterruptableFuture` doesn't have yet), and I wonder if it will get easier when the two are somehow merged. The reduction in LOC won't be much (excluding comments and whitespace, this class is only 23 lines). Along the same lines, I wonder if using `InterruptableFuture` here would make the code simpler/easier to maintain, as `CompletableFuture` is more familiar, better documented, etc.

That said, I'm fine to spend a morning to investigate more deeply if the two can be combined somehow.

sungshik (commented):
Hi Davy, thanks for the comments. Pieter already addressed a few of them; here are the remaining ones.

DavyLandman (approved):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java
```diff
@@ -45,64 +67,178 @@
...
+    private final AtomicReference<@MonotonicNonNull Versioned<ITree>> lastWithoutErrors;
+    private final AtomicReference<@MonotonicNonNull Versioned<ITree>> last;
 
-    private final ISourceLocation file;
-    @SuppressWarnings("java:S3077") // we are use volatile correctly
-    private volatile Versioned<String> currentContent;
-    @SuppressWarnings("java:S3077") // we are use volatile correctly
-    private volatile @MonotonicNonNull Versioned<ITree> lastFullTree;
-    @SuppressWarnings("java:S3077") // we are use volatile correctly
-    private volatile CompletableFuture<Versioned<ITree>> currentTree;
```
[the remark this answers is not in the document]
sungshik: - Debounce (800ms): When `didChange` happens (but nothing else yet)
  - Immediate: When any actual operation is requested

So, as long as a parse tree isn't needed yet, it's (pre)computed only after debouncing, but as soon as any operation needs it, the debounce is cut short and the tree is immediately computed.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParserOnlyContribution.java
```diff
@@ -97,7 +102,8 @@ private static Either<IFunction, Exception> loadParser(ParserSpecification spec)
         try {
             logger.debug("Loading parser {} at {}", reifiedType, spec.getParserLocation());
             // this hides all the loading and instantiation details of Rascal-generated parsers
-            var parser = vf.loadParser(reifiedType, spec.getParserLocation(), VF.bool(spec.getAllowAmbiguity()), VF.bool(false), VF.bool(false), VF.bool(false), vf.set());
+            var parser = vf.loadParser(reifiedType, spec.getParserLocation(), VF.bool(spec.getAllowAmbiguity()), VF.integer(MAX_AMB_DEPTH),
```
DavyLandman, line 105: the `ParserSpecification` should get a new flag (also documented in the json side) to enable error recovery, and be able to pass in these flags, instead of getting these defaults.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -91,6 +92,7 @@
 
 public class RascalLanguageServices {
     private static final IValueFactory VF = IRascalValueFactory.getInstance();
+    private static final ParseErrorRecovery RECOVERY = new ParseErrorRecovery(IRascalValueFactory.getInstance());
```
DavyLandman, line 95: here we do store this recovery instance, so that might seem wise in that other class?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/RascalServices.java
```diff
@@ -27,19 +27,42 @@
...
+        ITree tree = parser.parse(
+            Parser.START_MODULE, loc.getURI(), input, MAX_AMB_DEPTH,
+            new NoActionExecutor(),
+            new DefaultNodeFlattener<>(),
+            new UPTRNodeFactory(true),
+            new ToTokenRecoverer(loc.getURI(), parser, new StackNodeIdDispenser(parser), MAX_RECOVERY_ATTEMPTS, MAX_RECOVERY_TOKENS));
+
+        // Recover
+        ParseErrorRecovery recoverer = new ParseErrorRecovery(VALUE_FACTORY);
+        return (ITree) recoverer.disambiguateParseErrors(tree, TRUE);
```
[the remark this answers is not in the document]
DavyLandman: I would feel safer if we have a small comment that explains that here, for the future developer that is picking that task up.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
```diff
@@ -121,14 +193,27 @@ public static Diagnostic translateDiagnostic(IConstructor d, Range range) {
...
-                loc.getOffset(), loc.getLength() + 1,
-                loc.getBeginLine(), loc.getBeginColumn(),
-                loc.getEndLine(), loc.getEndColumn() + 1);
+            try {
+                loc = ValueFactoryFactory.getValueFactory().sourceLocation(loc,
+                    loc.getOffset(), loc.getLength() + 1,
+                    loc.getBeginLine(), loc.getBeginColumn(),
+                    loc.getEndLine(), loc.getEndColumn() + 1);
+            } catch (IllegalArgumentException e) {
+                logger.trace("Cannot extend 0-width location for parse error: {}", e.getMessage());
```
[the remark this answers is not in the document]
sungshik: In very rare cases (I could only reproduce it when a file contains only empty lines), an exception would be thrown when the new location was created in the try-block. Assuming this was just a minor oversight of a valid corner case, I just added an exception handler to cover it.

Only today I noticed that there is actually a bug in the call to create the source location: the order of the args should be `beginLine`, `endLine`, `beginColumn`, `endColumn`, instead of the way it was written. Fixed this and removed the exception handler.

rascal-vscode-extension/src/lsp/ParameterizedLanguageServer.ts
```diff
@@ -183,6 +183,14 @@ export interface ParserSpecification {
     nonTerminalIsStart?: boolean;
     /** allow ambiguities during parsing, default: false */
     allowAmbiguity?: boolean;
+    /** limit the maximum nesting depth of ambiguities, ambiguities will be pruned below this level. Only of interest if allowAmbiguity or allowRecovery are set to true. default: 2 */
+    maxAmbDepth?: number;
+    /** Allow error recovery during parsing resulting in error trees. default: false */
+    allowRecovery?: boolean;
```
DavyLandman, line 189: Yes 👍🏼 

nit: Let's move `allowRecovery` as the first keyword, and then put a newline before it, to group the recovery config parameters together.

rascal-vscode-extension/src/test/vscode-suite/repl.test.ts
```diff
@@ -78,7 +78,7 @@ describe('REPL', function () {
         const repl = new RascalREPL(bench, driver);
         await repl.connect();
         expect(repl.lastOutput).is.equal("5\nint: 0");
-    }).timeout(Delays.extremelySlow * 3);
+    }).timeout(Delays.extremelySlow * 4);
```
DavyLandman, line 81: is this still needed? or was this a debugging change?

DavyLandman (approved):
Yes, thanks, this has improved 👍🏼

rascal-vscode-extension/src/lsp/ParameterizedLanguageServer.ts
PieterOlivier: I have now added newlines to better show this grouping.

PR #491, opened 2024-10-25
Both outline and codelenses support can now handle error trees

PieterOlivier:
Both `locateCodeLenses` (in `RascalLanguageServices.java`) and `outlineRascalModule` (in `Outline.rsc`) would crash if they encountered trees with error nodes.

This PR fixes these problems.

In case of the outline, elements with errors are removed from the outline.

In case of the code lenses, top-level elements with errors are ignored for code lenses.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
@@ -51,6 +51,7 @@ import org.eclipse.lsp4j.jsonrpc.messages.ResponseError;
 import org.eclipse.lsp4j.jsonrpc.messages.ResponseErrorCode;
 import org.rascalmpl.exceptions.Throw;
 import org.rascalmpl.interpreter.Evaluator;
+import org.rascalmpl.library.util.ErrorRecovery;
 import org.rascalmpl.library.util.PathConfig;
 import org.rascalmpl.values.IRascalValueFactory;
 import org.rascalmpl.values.functions.IFunction;
@@ -76,6 +77,7 @@ import io.usethesource.vallang.type.TypeStore;
 
 public class RascalLanguageServices {
     private static final IValueFactory VF = IRascalValueFactory.getInstance();
+    private static final ErrorRecovery RECOVERY = new ErrorRecovery(IRascalValueFactory.getInstance());
 
     private static final Logger logger = LogManager.getLogger(RascalLanguageServices.class);
 
@@ -236,8 +238,13 @@ public class RascalLanguageServices {
         List<CodeLensSuggestion> result = new ArrayList<>(2);
         result.add(new CodeLensSuggestion(module, "Import in new Rascal terminal", "rascalmpl.importModule", moduleName));
 
-        for (IValue topLevel : TreeAdapter
-            .getListASTArgs(TreeAdapter.getArg(TreeAdapter.getArg(tree, "body"), "toplevels"))) {
+        ITree body = TreeAdapter.getArg(tree, "body");
+        ITree toplevels = TreeAdapter.getArg(body, "toplevels");
+        for (IValue topLevel : TreeAdapter.getListASTArgs(toplevels)) {
+            if (RECOVERY.hasErrors((ITree) topLevel)) {
+                continue;
+            }
+
             ITree decl = TreeAdapter.getArg((ITree) topLevel, "declaration");
             if ("function".equals(TreeAdapter.getConstructorName(decl))) {
                 ITree signature = TreeAdapter.getArg(TreeAdapter.getArg(decl, "functionDeclaration"), "signature");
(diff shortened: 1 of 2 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -76,6 +77,7 @@
 
 public class RascalLanguageServices {
     private static final IValueFactory VF = IRascalValueFactory.getInstance();
+    private static final ErrorRecovery RECOVERY = new ErrorRecovery(IRascalValueFactory.getInstance());
```
sungshik, line 80:
```suggestion
    private static final ErrorRecovery RECOVERY = new ErrorRecovery(VF);
```

rascal-lsp/src/main/rascal/lang/rascal/lsp/Outline.rsc
```diff
@@ -31,38 +31,58 @@ import String;
...
-            children += [symbol(clean("<name>"), \key(), name@\loc, detail="tag <k> <name> on <ts>")];
+    children = [];
 
-        case (Declaration) `<Tags _> <Visibility _> data <UserType u> <CommonKeywordParameters kws>;`: {
-            kwlist = [symbol(".<k.name>", property(), k@\loc, detail="<k.\type>") | kws is present, KeywordFormal k <- kws.keywordFormalList];
-            children += [symbol("<u.name>", struct(), u@\loc, detail="data <u> <kws>", children=kwlist)];
+    top-down-break visit (m) {
+        case decl: (Declaration) `<Tags _> <Visibility _> <Type t> <{Variable ","}+ vars>;`:
+            if (!hasErrors(decl)) {
+                children += [symbol(clean("<v.name>"), variable(), v@\loc, detail="variable <t> <v>") | v <- vars, !hasErrors(v)];
```
sungshik, line 48: Is `!hasErrors(v)` needed in the comprehension? (I.e., does `hasErrors(decl)` look only at top-level errors in `decl`?)

rascal-lsp/src/main/rascal/lang/rascal/lsp/Outline.rsc
sungshik, line 1: We should consider adding more unit(-ish) tests to rascal-lsp (as Davy also suggested in a comment about the debouncer, and I think it's relevant also for the semantic tokenizer). Not necessarily in scope of this PR, but perhaps as part of the whole feature.

sungshik (approved):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
PieterOlivier: That will not work: `VF` is of type `IValueFactory` while we need a `IRascalValueFactory` here.

rascal-lsp/src/main/rascal/lang/rascal/lsp/Outline.rsc
```diff
@@ -31,38 +31,58 @@ import String;
...
+            }
+
+        case decl: (Declaration) `<Tags _> <Visibility _> anno <Type t> <Type ot>@<Name name>;`:
```
jurgenvinju, line 51: once the match succeeds we now this node is not in error, so we don't need to check `hasErrors` for that.. but...

rascal-lsp/src/main/rascal/lang/rascal/lsp/Outline.rsc
```diff
@@ -31,38 +31,58 @@ import String;
...
+            if (!hasErrors(decl)) {
+                children +=  [symbol(clean("<name>"), field(), t@\loc, detail="anno <t> <ot>")];
+            }
+
+        case decl: (Declaration) `<Tags _> <Visibility _> alias <UserType u> = <Type al>;`:
+            if (!hasErrors(decl)) {
+                children += [symbol(clean("<u.name>"), struct(), u@\loc, detail="<u> = <al>")];
```
jurgenvinju, line 58: but here we use field projecten `u.name` and `u` might not even have a name if `UserType u` is an error tree. This is going to happen all over the place for error trees, so I'd like a more reusable/generic solution then having to guard all the field projections. It will become a cross-cutting and scattered concern everywhere otherwise, while error trees are a builtin language concept that features like `.` projection {c,sh,w}ould handle gracefully.

jurgenvinju: I think we may assume that if a programmer uses `u.name` then in that place `u.name` normally does not fail, and so it should also not fail for error children.

* if u is an error tree, and the error "dot" is after the `name` field, then field projection can still work as it was supposed to.
* otherwise if the "dot" is before the `name` field or on it, then it may be possible to return an empty stub error tree for `name` instead of failing.

If we have such a semantics for field projection, there would be a lot less pressure on user code for dealing with all kinds of exceptions. In general for productions with `n` children, there are I guess something like `n!/(2 * (n-2)!)`  possible combinations of having and not having error children, so that can become untenable.

Another option is to rewrite all field projections to pattern matches, which would fail in the absence of a field. But in this example that would lead to a lot of superfluous case distintions:

```rascal
syntax UserType
	= name: QualifiedName name 
	| parametric: QualifiedName name >> "[" "[" {Type ","}+ parameters "]" ;
```

Here we really do not care about the difference between the two rules but rather their commonality which is fine by the `.name` field projection. So I guess pattern matching is not a solution here.

PR #492, opened 2024-10-25
Process parse errors (including error nodes) in `TextDocumentState`

sungshik:
Before this PR, parse trees were processed *twice* to find error nodes (i.e., the exact same computation was repeated a second time): first in `TextDocumentState` (to figure out if `lastTreeWithoutErrors` should be updated), and next in `RascalTextDocumentService` (to translate error nodes to LSP diagnostics).

After this PR, all processing of parse errors (including error nodes) is moved out of `RascalTextDocumentService` into `TextDocumentState`. There is a new API call `getCurrentDiagnosticsAsync` (optionally with a debounce delay) for consumers to get diagnostics, similar to the existing `getCurrentTreeAsync`.


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java
```diff
@@ -158,22 +194,58 @@ private void parseIfNotParsing() {
...
-                            treeAsync.completeExceptionally(error);
                         }
+
+                        // Complete future to get diagnostics
+                        diagnosticsAsync.complete(diagnostics);
                     });
             }
         }
+
+        private List<Diagnostic> toDiagnostics(ITree tree, Throwable excp) {
```
sungshik, line 220: The code in this method is copied verbatim from `RascalTextDocumentService`. I might like to refactor it a little bit after the first review round, but thought it'd be easier to keep it like this for now.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
```diff
@@ -215,51 +215,20 @@ private TextDocumentState updateContents(VersionedTextDocumentIdentifier doc, St
...
-
-            if (tree != null) {
-                RascalValueFactory valueFactory = (RascalValueFactory) ValueFactoryFactory.getValueFactory();
-                IList errors = new ErrorRecovery(valueFactory).findAllErrors(tree.get());
-                for (IValue error : errors) {
-                    ITree errorTree = (ITree) error;
-                    parseErrors.add(Diagnostics.translateErrorRecoveryDiagnostic(errorTree, columns));
-                }
-            }
-
```
sungshik, line 249: This code was moved verbatim to `TextDocumentState`

sungshik (commented):
Additional comments

PieterOlivier:
Looking good! I do think you should should move the `Debouncer` and `Update` classes to their own files. You can introduce a new package if you do not want expose these classes to the rest of the codebase.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java
```diff
@@ -158,22 +194,58 @@ private void parseIfNotParsing() {
...
-                    .thenApply(t -> new Versioned<>(version, t))
-                    .whenComplete((t, error) -> {
-                        if (t != null) {
-                            var errors = RECOVERY.findAllErrors(t.get());
-                            if (errors.isEmpty()) {
-                                Versioned.replaceIfNewer(lastWithoutErrors, t);
+                    .whenComplete((t, e) -> {
+
+                        // Prepare result values for futures
+                        var tree = new Versioned<>(version, t);
```
PieterOlivier, line 200: As a `<tree,diagnostics>` pair seem to belong together (not suprisingly), would ik make sense to have one `Versioned` object with a version and a pair of `<tree,diagnostics>`?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
```diff
@@ -73,13 +73,14 @@ public static <K, V> Map<K, List<V>> groupByKey(Stream<Entry<K, V>> diagnostics)
     }
 
     public static Diagnostic translateDiagnostic(ParseError e, ColumnMaps cm) {
-        return new Diagnostic(toRange(e, cm), e.getMessage(), DiagnosticSeverity.Error, "parser");
+        var message = e.getMessage() + " (irrecoverable)";
```
PieterOlivier, line 76: Nice, especially as we are still building confidence in error recovery so instantly knowing when recovery succeeded and when not is very helpful|!

PR #497, opened 2024-11-01
Semantic tokenizer improvements: tests, simplification, syntax-in-syntax bugfix

### LINKED ISSUE #456: Semantic tokenizer makes mistakes when a `syntax` tree (with category) has `syntax` children
**Describe the bug**

```rascal
```

**To Reproduce**

```
void f() {
    int  x = 3;
    real y = 3.14;
    rat  z = 3/14;
}
```

**Expected behavior**

**Screenshots**

image

**Additional context**

```rascal
syntax Literal
    | ...
    ;

syntax  IntegerLiteral  = ...
lexical RealLiteral     = ...
lexical RationalLiteral = ...
```

sungshik:
This PR:

  - adds tests for the semantic tokenizer;
  - removes duplicate logic for computing maximal tokens;
  - fixes #456.

Before this PR:

image

After this PR:

image

See this history for details:

  - #494
  - #500
  - ~~#501~~ (replaced by <span>#</span>502)
  - #502
  - 69ab630


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizerTester.java
```diff
@@ -0,0 +1,57 @@
...
+        var tokens = new IValue[encoded.size() / 5];
+        var tokenTypes = tokenizer.capabilities().getTokenTypes();
+
+        for (int i = 0; i < encoded.size(); i += 5) {
+            var deltaLine     = values.integer(encoded.get(i));
+            var deltaStart    = values.integer(encoded.get(i + 1));
+            var length        = values.integer(encoded.get(i + 2));
+            var tokenType     = values.string(tokenTypes.get(encoded.get(i + 3)));
+            var tokenModifier = values.string(""); // Token modifiers aren't supported yet
+            tokens[i / 5] = values.tuple(deltaLine, deltaStart, length, tokenType, tokenModifier);
```
DavyLandman, line 52: nit: we  normally use list writers and append on them, instead of keeping this array around.

sungshik: Ah, thanks, I didn't realize. The writer even has `appendTuple` 🙂

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/LanguageContributionsMultiplexer.java
```diff
@@ -159,6 +161,8 @@ private synchronized void calculateRouting() {
         hasReferences = anyTrue(ILanguageContributions::hasReferences);
         hasImplementation = anyTrue(ILanguageContributions::hasImplementation);
 
+        specialCaseHighlighting = anyTrue(ILanguageContributions::specialCaseHighlighting);
```
DavyLandman, line 164: for the parser we do something special, we say `firstOrFail` can we do something similar here as well? so that we only pickup the parsing status from the first contribution?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizerTester.java
```diff
@@ -0,0 +1,57 @@
...
+import io.usethesource.vallang.IConstructor;
+import io.usethesource.vallang.IValue;
+
+public class SemanticTokenizerTester {
+
+    public IValue toTokens(IConstructor tree, IBool useSpecialCaseHighlighting, IBool applyRascalCategoryPatch) {
+        var tokenizer = new SemanticTokenizer(applyRascalCategoryPatch.getValue());
+        var encoded = tokenizer.semanticTokensFull((ITree) tree, useSpecialCaseHighlighting.getValue()).getData();
+
+        var values = IRascalValueFactory.getInstance();
```
DavyLandman, line 42: a rascal library function should get the value factory injected in the constructor of the class.

rascal-lsp/src/main/rascal/lang/rascal/tests/semanticTokenizer/Util.rsc
```diff
@@ -0,0 +1,175 @@
...
+    = [a | Actual a: <<_, _, _, tokenType, _>, _, _> <- actuals];
+
+private list[Actual] filterByString(list[Actual] actuals, str string)
+    = [a | a <- actuals, contains(a.s, string)];
+
+//
+// Expects
+//
+
+data Expect // Represents...
```
DavyLandman, line 141: nit: newline after this one is a bit unexpected?

rascal-lsp/src/main/rascal/lang/rascal/tests/semanticTokenizer/Util.rsc
```diff
@@ -0,0 +1,175 @@
...
+
+    // ...the expectation that the `n`-th/first/last occurrence of `string` is
+    // in a token of type `tokenType`
+    = expectNth(int n, str string, str tokenType)
+    | expectFirst(str string, str tokenType)
+    | expectLast(str string, str tokenType)
+
+    // ...the expectation that each occurrence of `string` is not in a token of
+    // type `tokenType`
+    | expectEachNot(str string, str tokenType);
```
DavyLandman, line 151: nit: normally for a multiline adt we put the `;` on the next line.

rascal-lsp/src/main/rascal/lang/rascal/tests/semanticTokenizer/Util.rsc
```diff
@@ -0,0 +1,175 @@
...
+//
+// Length conversion: Characters are counted as 16-bit units in LSP, whereas
+// they are counted as 32-bit units in Rascal, so lengths need to be converted.
+// The difference manifests only in the presence of surrogate pairs.
+//
+
+alias Length32 = int;
+alias Length16 = int;
+
+private Length32 toLength32(Length16 n16, str input) {
```
DavyLandman, line 106: I'm very happy we have tests, but a bit sad we have to write all of this.

rascal-vscode-extension/src/lsp/ParameterizedLanguageServer.ts
```diff
@@ -183,7 +183,8 @@ export interface ParserSpecification {
     nonTerminalIsStart?: boolean;
     /** allow ambiguities during parsing, default: false */
     allowAmbiguity?: boolean;
-
+    /** apply the special case for highlighting syntax-in-syntax, default: true */
```
DavyLandman, line 186: maybe also add the warning here that it will get replaced soon.

DavyLandman (changes requested):
Looks good, a few small things I noticed.

rascal-lsp/src/main/rascal/lang/rascal/tests/semanticTokenizer/Util.rsc
sungshik: Agreed...

DavyLandman (approved):
LGTM, I also checked with @jurgenvinju and he also approves of the change to the `parsing` constructor.

PR #501, opened 2024-11-04
Add token collector modes

### LINKED ISSUE #456: Semantic tokenizer makes mistakes when a `syntax` tree (with category) has `syntax` children

sungshik:
### This PR

This PR adds support to run the token collector in one of the following modes:

  - `INNER_OVER_OUTER_LEGACY` (default for DSLs): In this mode, the behavior of the semantic tokenizer is exactly the same as before this PR, namely inner-over-outer for nested categories, except syntax-in-syntax. No changes are needed to existing DSL grammars.

  - `INNER_OVER_OUTER_STRICT` (default for Rascal): Essentially the same as `INNER_OVER_OUTER_LEGACY` but without the syntax-in-syntax special case. This fixes #456.
  
  - `OUTER_OVER_INNER`: In this mode, outer categories take precedence over inner categories. For some grammars, this [might be more intuitive](https://github.com/usethesource/rascal-language-servers/issues/456#issuecomment-2382842966). Furthermore, this mode turns out to be already useful to tokenize ambiguities (i.e.,  if a subtree is part of an ambiguity, then ambiguity tokens should be generated, while any non-ambiguity category inside the subtree should be ignored).

Further reading about inner-over-outer vs. outer-over-inner: #456

### Usage

The defaults can be explicitly overridden in grammars by annotating a production with the following tags:

```rascal
@categoryNesting="innerOverOuterLegacy"
@categoryNesting="innerOverOuterStrict"
@categoryNesting="outerOverInner"
```

The declared mode will be used in the whole subtree starting from the annotated production. (Unless a subtree inside the subtree has its own `@categoryNesting` annotation.) In particular, users ready to migrate their DSLs to `INNER_OVER_OUTER_STRICT` only need to annotate the `start` production of the grammar with a `@categoryNesting` annotation.

### Examples

  - Migrating Pico to `INNER_OVER_OUTER_STRICT`:
   
    ```rascal
    module lang::pico::\syntax::Main

    import ParseTree;

    start syntax Program 
      = @categoryNesting="innerOverOuterStrict" program: "begin" Declarations decls {Statement  ";"}* body "end" ;

    ...
    ```

    (Pico is unaffected by the syntax-in-syntax special case, so legacy/strict mode is the same for Pico.)

  - Defining [reals in terms of ints](https://github.com/usethesource/rascal-language-servers/issues/456#issuecomment-2378861654):

    ```rascal
    lexical Ints = @category="integer" [0-9]+ !>> [0-9];
    lexical Reals = @categoryNesting="outerOverInner" @category="real" Ints "." Ints?;
    ```

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
@@ -430,13 +430,13 @@ public class SemanticTokenizer implements ISemanticTokens {
         public void collect(ITree tree) {
 
             // For Rascal, the default is the *strict* inner-over-outer mode
-            // (without syntax-in-syntax special case) that fixes #456.
+            // (without syntax-in-syntax special case) that fixes #456
             if (patch instanceof RascalCategoryPatch) {
                 collect(tree, Mode.INNER_OVER_OUTER_STRICT, null);
             }
 
             // For DSLs, the default is the *legacy* inner-over-outer mode (with
-            // syntax-in-syntax special case) to be backward-compatible.
+            // syntax-in-syntax special case) to be backward-compatible
             else {
                 collect(tree, Mode.INNER_OVER_OUTER_LEGACY, null);
             }
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
@@ -414,7 +414,7 @@ public class SemanticTokenizer implements ISemanticTokens {
             INNER_OVER_OUTER_LEGACY; // Syntax-in-syntax special case
 
             public static Mode toMode(String s) {
-                switch (s) {
+            switch (s) {
                 case "outerOverInner":
                     return Mode.OUTER_OVER_INNER;
                 case "innerOverOuterStrict":
@@ -557,7 +557,7 @@ public class SemanticTokenizer implements ISemanticTokens {
                 }
             }
             return ifAbsent;
-	    }
+        }
     }
 
     // The idea behind the patch is to dynamically map productions in Rascal's
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
@@ -48,6 +48,7 @@ import org.eclipse.lsp4j.SemanticTokensDelta;
 import org.eclipse.lsp4j.SemanticTokensLegend;
 import org.eclipse.lsp4j.SemanticTokensWithRegistrationOptions;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
+import org.rascalmpl.values.RascalValueFactory;
 import org.rascalmpl.values.parsetrees.ITree;
 import org.rascalmpl.values.parsetrees.ProductionAdapter;
 import org.rascalmpl.values.parsetrees.SymbolAdapter;
@@ -57,6 +58,7 @@ import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 
 import io.usethesource.vallang.IConstructor;
+import io.usethesource.vallang.INode;
 import io.usethesource.vallang.IString;
 import io.usethesource.vallang.IValue;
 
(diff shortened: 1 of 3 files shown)


sungshik (commented):
Additional comments

sungshik:
Additional comment: I feel the user interface is a bit clunky and welcome suggestions. (Maybe @jurgenvinju, @DavyLandman, or @rodinaarssen has ideas?) For now, I settled for these lengthy tags (e.g., `@categoryNesting="innerOverOuterLegacy"`) because I expect:

 1. these tags will be rarely used, so the annoyance will hopefully be rare as well 😉;
 2. in the rare cases they're written, it will probably be just at the `start` non-terminal, so it's likely to be limited to a specific place;
 3. in the rare cases they're read, the lengthiness is descriptive and might actually be useful to understand what the tag does (more so than, e.g., `@nesting="innerOverOuterLegacy"` or just `@innerOverOuterLegacy`).

Other considerations:

  - When strict mode becomes the default, the need to override the default might (should?) become even smaller
  - We probably need to revisit this at some point in the context of https://github.com/usethesource/rascal/issues/1928

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -454,21 +485,28 @@ private void collectAppl(ITree tree, @Nullable String parentCategory) {
...
-                        ProductionAdapter.isSort(TreeAdapter.getProduction((ITree) child))) {
-                    collect((ITree) child, null);
-                } else {
-                    collect((ITree) child, category);
-                }
+
+                // Special case (syntax-in-syntax): In legacy mode, do *not*
+                // propagate `category` when both `tree` and `child` are syntax
+                // non-terminals.
+                var specialCase =
```
PieterOlivier, line 492: There is a `isContextFree` method in `ProductionAdapter`, maybe this condition can be written even more clearly when using that?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -502,6 +540,22 @@ private static boolean isKeyword(ITree tree) {
 
             return true;
         }
+
+        // Derived from `org.rascalmpl.values.parsetrees.ProductionAdapter.getCategory`
+        private static Mode getModeIfPresent(IConstructor prod, Mode ifAbsent) {
```
PieterOlivier, line 545: It might be a good idea to submit a PR for Rascal to refactor `ProductionAdapter.getCategory` and extract `getAttribute(String name)`.

I understand you cannot use that method yet, but after the next release of Rascal you can simplify this code to use the `getAttribute` method.

PieterOlivier (approved):

DavyLandman:
Whatever we end up doing, I think it should be mentioned along side the `parse` contribution documentation in `util::LanguageServer`.

I like that we have a way to offer the user the fallback of the old behavior. We should make sure that we explain this in the release notes, and also announce that it will switch in the future. So either rewrite the grammar and opt-in to the new behavior, or explicitaly set you want the old behavior for now. 

With regards to the names of the categories, I agree, they are a bit verbose. Some points for this:

1. Maybe we can rename `@categoryNesting` to `@highlighting`? 
2. We should move away from strings, and instead have an adt (defined in `util::LanguageServer`) that has the 3 options. So it would become `@highlighting=outerOverInner()`. Although this would make a grammar suddenly depend on the `rascal-lsp` project, which we tend to avoid.
3. We could also consider dropping the descriptive names, and going for `data HighlightingMode = legacy() | current();` although that doesn't leave us much room for future additions. (although I hope we won't be switching it up more).
4. Alternatively we could also not be so configurable and only add a `@newHighlighting=true` flag to the grammar, such that
5. Another alternative which would be less invasive to the grammar would be to add it as a keyword parameter to the `parse` contribution, such that a it's very local to the rascal-lsp `util::LanguageServer` module.

I think my preference would be 5.

jurgenvinju:
Hi cool stuff!!ighlighting is not only a feature of the LSP but also in other standard library features with ANSI coloring. 

So I think we should port these features there, which will also make the option `data` types available to all via the stdlib. We can't have the dependency from grammars on the LSP like Davy day said.

DavyLandman:
I think @sungshik moved to #502 (point 5) to make this less of a big change. Especially since this is about a bugfix around some weird behavior we introduced, that wasn't compatible with the previous interpretations.

And indeed, we could think of this feature as a generic rascal feature/annotation, but move that to a later stage than the bugfix.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
sungshik: Good suggestion. I ended up not using tags after all, so this code isn't needed anymore, but I filed an [issue](https://github.com/usethesource/rascal/issues/2072) to do this improvement at some point anyway.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
sungshik: Using `isContextFree` would slightly change the semantics of the condition. As we're keeping the special case only for backward-compatibility at this point (it is planned to be removed in a future release), I think we need to keep the current formulation.

sungshik:
Hi all, yes, sorry about that. I started to work on Davy's alternative 5 in a separate branch (#502), which is almost ready. I'll close this PR to avoid further confusion.

PR #502, opened 2024-11-05
Add special-case-highlighting parameter to parsing service

### LINKED ISSUE #456: Semantic tokenizer makes mistakes when a `syntax` tree (with category) has `syntax` children

sungshik:
### Overview

Before this PR, the semantic highlighter applied a special case for `syntax`-in-`syntax` (sub)trees. However, this gave rise to unexpected behavior and caused #456.

After this PR:

  - For Rascal, the special case is no longer applied. This fixes #456.
  - For DSLs, the special case continues to be applied for backward-compatibility. That is, for now, no changes are needed to existing grammars. However, users are encouraged to update existing grammars that rely on the special case, as it's deprecated.
  
To control whether or not the special case is applied for DSLs, this PR temporarily -- for the next few release cycles -- adds keyword parameter `usesSpecialCaseHighlighting` to constructor `parsing` of ADT `LanguageServer`. 

  - Present release cycle: The default is `true`. Existing grammars that rely on the special case will have the expected highlighting. They could be updated (encouraged), and `usesSpecialCaseHighlighting` could be set to `true` (encouraged). 
  - Future release cycle (short-term): The default is `false`. Existing grammars that rely on the special case may not have the expected highlighting. Either they should be updated (encouraged), or `usesSpecialCaseHighlighting` should be set to `true` (discouraged).
  - Future release cycle (mid-term): `usesSpecialCaseHighlighting` will be removed (i.e., the special case ceases to exist). Existing grammars that rely on the special case may not have the expected highlighting. They must be updated. 

### A few details

The updated constructor:

https://github.com/usethesource/rascal-language-servers/blob/9175367921454b0e80ed1bf1c5ad30e7838ff3fe/rascal-lsp/src/main/rascal/util/LanguageServer.rsc#L248-L249

The updated docs:

https://github.com/usethesource/rascal-language-servers/blob/9175367921454b0e80ed1bf1c5ad30e7838ff3fe/rascal-lsp/src/main/rascal/util/LanguageServer.rsc#L173-L183

The updated Pico language server as an example:

https://github.com/usethesource/rascal-language-servers/blob/9175367921454b0e80ed1bf1c5ad30e7838ff3fe/rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc#L48-L50


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/ISemanticTokens.java
sungshik, line 1: This interface didn't seem to be meaningfully used anywhere (nowhere used as a type; only one implementing class). However, changes were now required to some of the method signatures. I thought I'd just remove it instead... (But I'd be happy to put it back if I missed something.)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -74,24 +71,12 @@ public SemanticTokenizer(boolean rascal) {
...
 
-    @Override
-    public Either<SemanticTokens, SemanticTokensDelta> semanticTokensFullDelta(String previousId, ITree tree) {
-        return Either.forLeft(semanticTokensFull(tree));
-    }
-
-    @Override
-    public SemanticTokens semanticTokensRange(Range range, ITree tree) {
-        return semanticTokensFull(tree);
-    }
```
sungshik, line 92: Unused functions (but I still needed to update them because the signature of `semanticTokensFull` changed). I thought I'd just remove them instead... If we need them in the future, we can add them again.

sungshik (commented):
Additional comments

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/ISemanticTokens.java
sungshik, line 1: This interface didn't seem to be meaningfully used anywhere (nowhere used as a type; only one implementing class). However, changes were now required to some of the method signatures. I thought I'd just remove it instead... (But I'd be happy to put it back if I missed something.)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SemanticTokenizer.java
```diff
@@ -74,24 +71,12 @@ public SemanticTokenizer(boolean rascal) {
...
```
sungshik, line 92: Unused functions (but I still needed to update them because the signature of `semanticTokensFull` changed). I thought I'd just remove them instead... If we need them in the future, we can add them again.

DavyLandman (changes requested):
We're missing the typescript side of the dedicated parser configuration.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/ISemanticTokens.java
sungshik: Added in 1-2 March 2021 as part of the initial implementation of the semantic tokenizer (6a73d54 and a0e0e16). After that, only license updates.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParserOnlyContribution.java
```diff
@@ -228,6 +228,11 @@ public CompletableFuture<Boolean> hasInlayHint() {
         return CompletableFuture.completedFuture(false);
     }
 
+    @Override
+    public CompletableFuture<Boolean> specialCaseHighlighting() {
+        return CompletableFuture.completedFuture(false);
```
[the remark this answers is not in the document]
sungshik: Thanks, you're right. Fixed in 22f28d9.

PR #511, opened 2024-11-08
Implemented error recovery support in parametric language servers

PieterOlivier:
This PR leverages the error recovery support in `TextDocumentService` to also support error recovery for parametric language services.
The Pico demo has been extended with error recovery support.

The first step when supporting error recovery is to change the parser to do error recovery and parse tree disambiguation. The disambiguation if needed because error trees are usually highly ambiguous. The `createParseErrorFilter` function returns a filter to disambiguate error trees so only a single alternative is left (based on simple heuristics so in some cases you might want to implement your own filter):
```
private Tree picoParser(str input, loc origin) {
    return parse(#start[Program], input, origin, allowRecovery=true, filters={createParseErrorFilter(false)});
}
...
set[LanguageService] picoLanguageServer() = {
    parsing(picoParser, usesSpecialCaseHighlighting = false),
```
After that, the individual services need to be able to work even when error trees are present. In case of Pico, only a small change was needed to `picoDocumentSymbolService`, all other services kept working without changes.


rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
```diff
@@ -195,7 +195,7 @@ void main() {
             pathConfig(),
             "Pico",
             {"pico", "pico-new"},
-            "demo::lang::pico::NewLanguageServer",
+            "demo::lang::pico::LanguageServer",
```
sungshik, line 198: This fix should probably be released asap. I made a separate PR (#513) that targets `main` just for this.

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
```diff
@@ -74,7 +74,7 @@ symbol search in the editor.
 }
 list[DocumentSymbol] picoDocumentSymbolService(start[Program] input)
   = [symbol("<input.src>", DocumentSymbolKind::\file(), input.src, children=[
-      *[symbol("<var.id>", \variable(), var.src) | /IdType var := input]
+      *[symbol("<var.id>", \variable(), var.src) | /IdType var := input, var has id]
```
sungshik, line 77: `var has id` is basically the check to see if it's an error node, right? If so, for didactic purposes, it might be instructive to spell it out explicitly, like:

```suggestion
      *[symbol("<var.id>", \variable(), var.src) | /IdType var := input, !hasParseError(var)]
```

With:

```rascal
bool hasParseError(var) = var has id;
```

Or something along those lines. It would be a concrete example to demonstrate how to make a service resilient to recoverable parse errors.

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
sungshik, line 1: ### Observation

The analyzer and the outliner can be inconsistent.

I suspect that maybe the analyzer and the outliner actually do produce consistent output, but that there's an issue with management of (versions of) diagnostics. I remember the code to do that is quite complicated and was tricky to get right -- and maybe we didn't after all.

### Example 1: Consistent

Consider this screenshot:

image

  - The analyzer does report that `b` and `c` are undefined
  - The outliner doesn't show `b` and `c`

### Example 2: Inconsistent

Consider this screenshot:

image

  - The analyzer does report `c` is undefined
  - The outliner does show `c`

### Example 3: Inconsistent

Consider this screenshot:

image

  - The analyzer doesn't report `b` is undefined
  - The outliner doesn't show `b`

sungshik (changes requested):
The analyzer runs on parse trees with errors! 😲 That's beyond my expectations. Really cool 🙂 There are some inconsistencies between analyzer diagnostics and the outline, though, so that requires some additional tweaking. I suspect it might be a bug in version management of diagnostics; see the comment below for more details.

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
PieterOlivier: We can just use the `util::ErrorRecovery::hasErrors` method. In general this is expensive because the whole tree is traversed, but in this case the tree is small so that is not a problem.

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
PieterOlivier: That PR has been merged so I have removed this change and merged changes from main into `error-recovery/rascal` and  subsequently `error-recovery/rascal` into `error-recovery/parametric`.

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
PieterOlivier: There is definitely something fishy going on. Lets dive into this together.

sungshik: This turned out to be a simpler issue than I expected: in some of the above cases, the analyzer crashed because of parse errors that it couldn't deal with. The addition of `!hasErrors(...)` somewhere fixed it. Screenshots of the same examples as before:

  - **Example 1:**
    image

  - **Example 2:**
    image

  - **Example 3:**
    image

sungshik:
I fixed the analyzer issue (see [comment](https://github.com/usethesource/rascal-language-servers/pull/511#discussion_r1836661410)), but I'm not sure if it addresses all potential issues in the analyzer. Remaining questions:

  - This is fine, I think? (Or can `id` have errors?)

    https://github.com/usethesource/rascal-language-servers/blob/5a36693e1f3e1ada0e7cb1d54ce7849cbab074a3/rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc#L100-L101

  - Do we need to check `hasErrors(var)` here, too? Not sure because `src` is maybe always defined.

    https://github.com/usethesource/rascal-language-servers/blob/5a36693e1f3e1ada0e7cb1d54ce7849cbab074a3/rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc#L103-L104

  - What about here? (Probably the same as the first question)

    https://github.com/usethesource/rascal-language-servers/blob/5a36693e1f3e1ada0e7cb1d54ce7849cbab074a3/rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc#L119-L123

PieterOlivier:
Ik denk dat in deze gevallen niets hoeft te gebeuren omdat lexicals nooit error trees kunnen zijn.

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
```diff
@@ -116,6 +116,13 @@ describe('IDE', function () {
         await ide.hasErrorSquiggly(editor);
     });
 
+    it("error recovery works", async function() {
+        const editor = await ide.openModule(TestWorkspace.mainFile);
+        await ide.hasSyntaxHighlighting(editor);
+        await editor.setTextAtLine(2, "1 2 3");
+        await ide.hasRecoveredError(editor);
```
sungshik, line 123: No `Delays.slow` second arg needed here?

rascal-vscode-extension/src/test/vscode-suite/utils.ts
```diff
@@ -179,6 +179,26 @@ function scopedElementLocated(scope:WebElement, selector: Locator): WebElementCo
...
+    return new WebElementCondition("locating element in scope", async (_driver) => {
+        try {
+            const result = await scope.findElements(selector);
+            if (result && result.length > 0) {
+                for(const element of result) {
+                    if ((await element.getRect()).y >= minTop) {
+                        return element;
+                    }
+                }
+                //return result[0] ?? null;
```
sungshik, line 192: Nit: Remove comment

rascal-vscode-extension/src/test/vscode-suite/utils.ts
```diff
@@ -234,6 +254,11 @@ export class IDEOperations {
         return this.driver.wait(until.elementLocated(By.className("squiggly-error")), timeout, message, 50);
     }
 
+    hasRecoveredError(editor: TextEditor, timeout = Delays.normal, message = "Missing recovered parse error"): Promise<WebElement> {
+        // We need to differentiate between real parse errors (error at first line) and recovered parse error (error at parse position).
+        return this.driver.wait(minTopElementLocated(editor, By.className("squiggly-error"), 20), timeout, message, 50);
```
sungshik, line 259: Here are some alternatives to consider, to make the test less brittle:

  - Measure somehow the offset of the bottom of line 1 instead of hardcoding it to `20`.
  - Make two syntax errors at the same time, so there are two squiggly elements. It's a bit indirect, but it can happen only when error recovery is enabled.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/model/ParametricFileFacts.java
```diff
@@ -319,20 +319,25 @@ private void sendDiagnostics() {
                 logger.debug("Cannot send diagnostics since the client hasn't been registered yet");
                 return;
             }
-            var messages = Lists.union(
-                unwrap(parserDiagnostics),
-                unwrap(analyzerDiagnostics),
-                unwrap(builderDiagnostics));
-            logger.trace("Sending diagnostics for {}. {} messages", file, messages.size());
+
+            // Read the atomic references once for the whole method, to ensure
```
DavyLandman, line 323: why was unwrapped removed? now there are 2 `.get` function calls? (I mean, maybe we can find a better way for the unwrap function, or add a comment next to the unwrap)

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
```diff
@@ -36,17 +36,22 @@ module demo::lang::pico::LanguageServer
...
+
 @synopsis{A language server is simply a set of ((LanguageService))s.}
 @description{
 Each ((LanguageService)) for pico is implemented as a function.
 Here we group all services such that the LSP server can link them
 with the ((Language)) definition later.
 }
 set[LanguageService] picoLanguageServer() = {
-    parsing(parser(#start[Program]), usesSpecialCaseHighlighting = false),
+    parsing(picoParser, usesSpecialCaseHighlighting = false),
```
DavyLandman, line 54: Remind me: now we're depending on the evaluator lock right? There was no current way around it? (because we're calling the `filters`)

rascal-vscode-extension/src/test/vscode-suite/utils.ts
DavyLandman: or at the very least, make 20 a parameter tot he ``hasRecoveredError`?

Could we also say: we need both an squiggly-error plus syntax highlighting?

runUItests.sh
```diff
@@ -22,6 +22,6 @@ npm run compile-tests
 # test what was compiled
 
 exec npx extest setup-and-run out/test/vscode-suite/*.test.js \
-    --code_version 1.82.3 \
+    --code_version 1.98.0 \
```
DavyLandman, line 25: why are we going this high? we want to make sure we're running against the "oldest" version we can support.

DavyLandman (approved):
Some small comments, looks good.

rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
PieterOlivier: That is true. But we have discussed in the past that it might be possible to recognize this specific situation: when calling the parser with only the error disambiguation filter we can forfeit calling the evaluator and do the filtering at the Java level.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/model/ParametricFileFacts.java
PieterOlivier: `unwrap` was removed because we have extended the `logger.trace` call with version information. The old `unwrap` method just dropped the version info.

rascal-vscode-extension/src/test/vscode-suite/utils.ts
PieterOlivier: I implemented both suggestions:
- Introduce and check two syntax errors
- Test for syntax highlighting

runUItests.sh
PieterOlivier: The `runUITests.sh` script now takes the version number from `package.json`.

PR #512, opened 2024-11-08
Simplify the debouncer

sungshik:
Before this PR: The debouncer used `CompletableFuture.delayedExecutor` to schedule futures with a delay. Once scheduled in this way, a delayed future *can't* be interrupted during the delay. Our use case requires such interruptions, though, so we had a rather complicated workaround.

After this PR: The debouncer uses `CompletableFuture.completeOnTimeout` to schedule futures with a delay. Once scheduled in this way, a delayed future *can* be interrupted during the delay, just by using a "normal" completion. The logic to distinguish between a normal completion and a timeout completion is straightforward. Overall, the debouncer has become a lot simpler.


PieterOlivier:
This is so much better!

My only concern is with the expensive tests. These tests have a lot of value but they also slow down the CI considerably. We should have a discussion if we want to pay this price on every CI run. Do we run the UI tests on a daily basis? If so we might consider moving these tests to that daily build run instead of in the CI.

PR #513, opened 2024-11-08
Fix configuration of Pico language server

sungshik:
We missed a rename of Pico's language server module. @PieterOlivier discovered and fixed it in a commit to the error recovery branch (097d81619b6dbbf16b0d0439d2cd488e1daad74b), but it may still take some time before that branch is merged into `main` (possibly in a next release cycle).


DavyLandman:
Why did the test still work? Or do they still run the old module?

jurgenvinju:
The tests still run the old demo so that the backwards compatibility layer is certified. Let's change that when people have had the chance to move you the new interfaces.

jurgenvinju:
Thanks @sungshik @PieterOlivier

PR #525, opened 2024-11-18
Rename performance improvements

toinehartman:
After testing with large Rascal projects (`rascal-core`, `typepal`), some low-hanging performance improvements to the rename refactoring presented themselves. Otherwise, the rename refactoring now mainly spends time in the type checker, which has some performance improvements under development.


DavyLandman (approved):
Good that you tackled a few of the causes of slowdown 👍🏼 

As a general performance tip:

```rascal
if (x <- [0..9999], x % 33 == 0) { return x; }
```

is correct, but written as a loop doesn't lean on the backtracker to do the iteration.

```rascal
for (x <- [0..9999], x % 33 == 0) { return x; }
```

will often be faster, or even:

```
for (x <- [0..9999]) { if (x % 33 == 0) return x; }
```

toinehartman:
> a loop doesn't lean on the backtracker to do the iteration

@DavyLandman Thanks, will keep that in mind and apply it when refactoring the renaming code later on.

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/WorkspaceInfo.rsc
```diff
@@ -294,7 +293,7 @@ set[loc] rascalGetOverloadedDefs(WorkspaceInfo ws, set[loc] defs, MayOverloadFun
             defPaths = defPathStep o selectedFields;
         } else {
             // Find definitions in the reached scope, and definitions within those definitions (transitively)
-            defPaths = defPathStep o invert(defScopes);
+            defPaths = defPathStep o (ws.defines<idRole, scope, defined>)[role]+;
```
toinehartman, line 296: This is the main performance improvement; way less potential overloads to check.

PR #526, opened 2024-11-19
Fix duplicate locations in rename edits.

toinehartman:
In some cases, names of definitions also appear as use locations in the TModel. In those cases, the rename logic would mark that location as both a definition and use.
Since the addition of annotated edits, locations are not deduplicated, so VS Code would receive multiple edits for the same location, and fail to apply all of them. Many renames would fail due to this, depending on what kind of name the cursor was on.

This PR fixes this behaviour; it postpones attaching annotations to most rename locations, simplifying location deduplication (and, as a side-effect, de-duplicates the annotations as well). It also adds assertions against duplicate edit locations to the test framework.

### DIFF
--- a/rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
+++ b/rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
@@ -62,6 +62,8 @@ import util::Reflective;
 
 alias Edits = tuple[list[DocumentEdit], map[ChangeAnnotationId, ChangeAnnotation]];
 
+private str MANDATORY_CHANGE_DESCRIPTION = "These changes are required for a correct renaming. They can be previewed here, but it is not advised to disable them.";
+
 // Rascal compiler-specific extension
 void throwAnyErrors(list[ModuleMessages] mmsgs) {
     for (mmsg <- mmsgs) {
@@ -192,13 +194,13 @@ private set[IllegalRenameReason] rascalCollectIllegalRenames(WorkspaceInfo ws, s
 private str rascalEscapeName(str name) = name in getRascalReservedIdentifiers() ? "\\<name>" : name;
 
 // Find the smallest trees of defined non-terminal type with a source location in `useDefs`
-private map[loc, loc] rascalFindNamesInUseDefs(start[Module] m, set[loc] useDefs) {
-    map[loc, loc] useDefNameAt = ();
+private rel[loc, loc] rascalFindNamesInUseDefs(start[Module] m, set[loc] useDefs) {
+    rel[loc, loc] nameOfUseDef = {};
     useDefsToDo = useDefs;
     visit(m.top) {
         case t: appl(prod(_, _, _), _): {
             if (t.src in useDefsToDo && just(nameLoc) := rascalLocationOfName(t)) {
-                useDefNameAt[t.src] = nameLoc;
+                nameOfUseDef += <nameLoc, t.src>;
                 useDefsToDo -= t.src;
             }
         }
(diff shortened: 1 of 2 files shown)


rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Util.rsc
```diff
@@ -108,6 +108,13 @@ bool isShorterTuple(tuple[loc, &T] t1, tuple[loc, &T] t2) = isShorter(t1[0], t2[
 }
 bool byOffset(loc l1, loc l2) = l1.offset < l2.offset;
 
+@synopsis{
+    Gives the longest location in a non-empty set of locations.
+    On a tie, gives one of the longest locations.
+}
+loc maxByLength(set[loc] locs:{_, *_}) = // at least one element
```
DavyLandman, line 115: nit: if there is at least one, that can also be used as the initial case (instead of an `unkown` loc).

```rascal
loc maxByLength({loc atLeastOne, set[loc] *rest}) 
    = (atLeastOne | it.length > l.length ? it : l | l.locs);
```

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
```diff
@@ -229,25 +231,32 @@ Maybe[loc] rascalLocationOfName(Nonterminal nt) = just(nt.src);
...
-                 | l <- renameLocs]>;
+    rel[loc l, Maybe[ChangeAnnotationId] ann, bool isDef] renames =
+        {<l, a, true>  | <l, a> <- defs}
+      + {<l, a, false> | <l, a> <- uses};
+    rel[loc name, loc useDef] nameOfUseDef = rascalFindNamesInUseDefs(m, renames.l);
+
+    ChangeAnnotationId defAnno = registerChangeAnnotation("Definitions", MANDATORY_CHANGE_DESCRIPTION, false);
+    ChangeAnnotationId useAnno = registerChangeAnnotation("References", MANDATORY_CHANGE_DESCRIPTION, false);
+
+    // Note: if the implementer of the rename logic has attached annotations to multiple rename suggestions that have the same
```
DavyLandman, line 249: this reads like it should be written in a different way? as in, that the comprehension below is too simple?

DavyLandman (approved):
LGTM, 2 small comments.

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Util.rsc
toinehartman: This is actually unused. Added it for an earlier implementation, and then rewrote. Will remove!

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
toinehartman: Well, the we do not have that information at all here currently. For rename locations that already have an annotation added at this stage (which are only `has` uses at this point in time), we only have the id of the annotation, but not its params. That still points in the direction of a necessary refactoring, probably. But I wanted to get the fix out first.

toinehartman: We could consider keeping a pair of `<loc, Maybe[ChangeAnnotation]>` instead, decouple the two using an `id` later, before returning to the Java world. Or even in Java.

toinehartman: Made an issue for this: https://github.com/usethesource/rascal-language-servers/issues/527

Issue #355, opened 2024-02-10
Fix rascal-lsp incorrect implementation of the LSP Document synchronisation for DSLs


DavyLandman:
During debugging of #354 I've discovered we implemented LSP incorrectly. For keeping it simple, I'll focus on the Parametric implementation, the Rascal implementation has roughly the same problems, but will require a different solution I suspect.

## Specification vs current implementation

I'll be quoting form the [LSP specification 3.17](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_synchronization) a bit.

> [Text Document Synchronization](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_synchronization)
> ...
> The document open notification is sent from the client to the server to signal newly opened text documents. The document’s content is now managed by the client and **the server must not try to read** the document’s content using the document’s Uri. Open in this sense means it is managed by the client. It doesn’t necessarily mean that its content is presented in an editor.

Currently our implementation starts to keep track of the document contents after an `open` event. Request that target this current document will use that state, and it's parse tree to provide callbacks to rascal lsp. So for the document in question, we respect the rules, we are not reading from disk.

However:

1. `exists` & `lastModified` and other functions in `IO` will go to disk, and potentially correspond to the wrong "state" of the documents. As a change in LSP does not have to mean a change visible on disk.
2. Other files that are read (references to other modules) are always read from disk, even though LSP clearly state that if those documents are also tracked by LSP, we are not supposed to read them from disk ourselves. We've seen this cause issues since we cannot take into account unsaved changes.
3. If not careful, usage of the current rascal-lsp can cause strange caching issues (the mismatch between `lastModified` and the unsaved state in the editor is a known example).

> [DidOpenTextDocument](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didOpen)
> ..
> Note that a server’s ability to fulfill requests is independent of whether a text document is open or closed.

Currently if VS Code asks us information about a document that is closed (or never opened before) we throw an exception. Even if we've seen the document before. The correct behavior is that is non-opened, we are allowed to read the file from disk, and answer the question that is asked.

> [VersionedTextDocumentIdentifier](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#versionedTextDocumentIdentifier)
> An identifier to denote a specific version of a text document. This information usually flows from the client to the server.

[...]

PR #531, opened 2024-11-26
Improve handling of qualified names when renaming

toinehartman:
- [x] Support renaming fully qualified modules names (instead of last name segment only)
- [x] Reject renaming of modules from qualified declaration names.
- [x] Explicitly do not support external imports
- [x] Support module renames from explorer view.
- [x] Escape name segments (`\syntax`) and v.v.
- [x] Resolve module names via dependencies
- [x] Use 'did rename' instead of 'will rename' callback, to prevent problems in case the evaluator is locked
- [x] Add UI test for module move
- [x] Change qualified names on folder rename

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
@@ -155,17 +155,19 @@ public abstract class BaseLanguageServer {
     public static void startLanguageServer(ExecutorService threadPool, Function<ExecutorService, IBaseTextDocumentService> docServiceProvider, BiFunction<ExecutorService, IBaseTextDocumentService, BaseWorkspaceService> workspaceServiceProvider, int portNumber) {
         logger.info("Starting Rascal Language Server: {}", getVersion());
 
-        var docService = docServiceProvider.apply(threadPool);
-        var wsService = workspaceServiceProvider.apply(threadPool, docService);
-        docService.pair(wsService);
-
         if (DEPLOY_MODE) {
+            var docService = docServiceProvider.apply(threadPool);
+            var wsService = workspaceServiceProvider.apply(threadPool, docService);
+            docService.pair(wsService);
             startLSP(constructLSPClient(capturedIn, capturedOut, new ActualLanguageServer(() -> System.exit(0), docService, wsService)));
         }
         else {
             try (ServerSocket serverSocket = new ServerSocket(portNumber, 0, InetAddress.getByName("127.0.0.1"))) {
                 logger.info("Rascal LSP server listens on port number: {}", portNumber);
                 while (true) {
+                    var docService = docServiceProvider.apply(threadPool);
+                    var wsService = workspaceServiceProvider.apply(threadPool, docService);
+                    docService.pair(wsService);
                     startLSP(constructLSPClient(serverSocket.accept(), new ActualLanguageServer(() -> {}, docService, wsService)));
                 }
             } catch (IOException e) {
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
@@ -42,6 +42,8 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutionException;
 import java.util.function.Supplier;
+
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.checkerframework.checker.nullness.qual.Nullable;
@@ -149,17 +151,19 @@ public abstract class BaseLanguageServer {
     }
 
     @SuppressWarnings({"java:S2189", "java:S106"})
-    public static void startLanguageServer(Supplier<IBaseTextDocumentService> service, int portNumber) {
+    public static void startLanguageServer(Supplier<Pair<IBaseTextDocumentService, BaseWorkspaceService>> service, int portNumber) {
         logger.info("Starting Rascal Language Server: {}", getVersion());
 
         if (DEPLOY_MODE) {
-            startLSP(constructLSPClient(capturedIn, capturedOut, new ActualLanguageServer(() -> System.exit(0), service.get())));
+            var services = service.get();
+            startLSP(constructLSPClient(capturedIn, capturedOut, new ActualLanguageServer(() -> System.exit(0), services.getLeft(), services.getRight())));
         }
         else {
             try (ServerSocket serverSocket = new ServerSocket(portNumber, 0, InetAddress.getByName("127.0.0.1"))) {
                 logger.info("Rascal LSP server listens on port number: {}", portNumber);
                 while (true) {
-                    startLSP(constructLSPClient(serverSocket.accept(), new ActualLanguageServer(() -> {}, service.get())));
+                    var services = service.get();
+                    startLSP(constructLSPClient(serverSocket.accept(), new ActualLanguageServer(() -> {}, services.getLeft(), services.getRight())));
                 }
             } catch (IOException e) {
                 logger.fatal("Failure to start TCP server", e);
@@ -206,10 +210,10 @@ public abstract class BaseLanguageServer {
         private final Runnable onExit;
         private IDEServicesConfiguration ideServicesConfiguration;
 
-        private ActualLanguageServer(Runnable onExit, IBaseTextDocumentService lspDocumentService) {
+        private ActualLanguageServer(Runnable onExit, IBaseTextDocumentService lspDocumentService, BaseWorkspaceService lspWorkspaceService) {
             this.onExit = onExit;
             this.lspDocumentService = lspDocumentService;
-            this.lspWorkspaceService = new BaseWorkspaceService(lspDocumentService);
+            this.lspWorkspaceService = lspWorkspaceService;
             reg.registerLogical(new ProjectURIResolver(this::resolveProjectLocation));
             reg.registerLogical(new TargetURIResolver(this::resolveProjectLocation));
         }
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
@@ -41,9 +41,9 @@ import java.util.Properties;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutionException;
-import java.util.function.Supplier;
+import java.util.concurrent.ExecutorService;
+import java.util.function.Function;
 
-import org.apache.commons.lang3.tuple.Pair;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.checkerframework.checker.nullness.qual.Nullable;
@@ -151,19 +151,21 @@ public abstract class BaseLanguageServer {
     }
 
     @SuppressWarnings({"java:S2189", "java:S106"})
-    public static void startLanguageServer(Supplier<Pair<IBaseTextDocumentService, BaseWorkspaceService>> service, int portNumber) {
+    public static void startLanguageServer(ExecutorService threadPool, Function<ExecutorService, IBaseTextDocumentService> docServiceProvider, Function<IBaseTextDocumentService, BaseWorkspaceService> workspaceServiceProvider, int portNumber) {
         logger.info("Starting Rascal Language Server: {}", getVersion());
 
+        var docService = docServiceProvider.apply(threadPool);
+        var wsService = workspaceServiceProvider.apply(docService);
+        docService.pair(wsService);
+
         if (DEPLOY_MODE) {
-            var services = service.get();
-            startLSP(constructLSPClient(capturedIn, capturedOut, new ActualLanguageServer(() -> System.exit(0), services.getLeft(), services.getRight())));
+            startLSP(constructLSPClient(capturedIn, capturedOut, new ActualLanguageServer(() -> System.exit(0), docService, wsService)));
         }
         else {
             try (ServerSocket serverSocket = new ServerSocket(portNumber, 0, InetAddress.getByName("127.0.0.1"))) {
                 logger.info("Rascal LSP server listens on port number: {}", portNumber);
                 while (true) {
-                    var services = service.get();
-                    startLSP(constructLSPClient(serverSocket.accept(), new ActualLanguageServer(() -> {}, services.getLeft(), services.getRight())));
+                    startLSP(constructLSPClient(serverSocket.accept(), new ActualLanguageServer(() -> {}, docService, wsService)));
                 }
             } catch (IOException e) {
                 logger.fatal("Failure to start TCP server", e);
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
@@ -42,6 +42,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
+import java.util.function.BiFunction;
 import java.util.function.Function;
 
 import org.apache.logging.log4j.LogManager;
@@ -151,11 +152,11 @@ public abstract class BaseLanguageServer {
     }
 
     @SuppressWarnings({"java:S2189", "java:S106"})
-    public static void startLanguageServer(ExecutorService threadPool, Function<ExecutorService, IBaseTextDocumentService> docServiceProvider, Function<IBaseTextDocumentService, BaseWorkspaceService> workspaceServiceProvider, int portNumber) {
+    public static void startLanguageServer(ExecutorService threadPool, Function<ExecutorService, IBaseTextDocumentService> docServiceProvider, BiFunction<ExecutorService, IBaseTextDocumentService, BaseWorkspaceService> workspaceServiceProvider, int portNumber) {
         logger.info("Starting Rascal Language Server: {}", getVersion());
 
         var docService = docServiceProvider.apply(threadPool);
-        var wsService = workspaceServiceProvider.apply(docService);
+        var wsService = workspaceServiceProvider.apply(threadPool, docService);
         docService.pair(wsService);
 
         if (DEPLOY_MODE) {
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseWorkspaceService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseWorkspaceService.java
@@ -30,6 +30,8 @@ import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.ExecutorService;
+
 import com.google.gson.JsonPrimitive;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.eclipse.lsp4j.ClientCapabilities;
@@ -51,12 +53,15 @@ public class BaseWorkspaceService implements WorkspaceService, LanguageClientAwa
     public static final String RASCAL_META_COMMAND = "rascal-meta-command";
     public static final String RASCAL_COMMAND = "rascal-command";
 
+    private final ExecutorService ownExecuter;
+
     private final IBaseTextDocumentService documentService;
     private final CopyOnWriteArrayList<WorkspaceFolder> workspaceFolders = new CopyOnWriteArrayList<>();
 
 
-    protected BaseWorkspaceService(IBaseTextDocumentService documentService) {
+    protected BaseWorkspaceService(ExecutorService exec, IBaseTextDocumentService documentService) {
         this.documentService = documentService;
+        this.ownExecuter = exec;
     }
 
 
@@ -122,6 +127,9 @@ public class BaseWorkspaceService implements WorkspaceService, LanguageClientAwa
     }
 
 
+    protected final ExecutorService getExecuter() {
+        return ownExecuter;
+    }
 
 
 }
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseWorkspaceService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseWorkspaceService.java
@@ -57,7 +57,6 @@ public class BaseWorkspaceService implements WorkspaceService, LanguageClientAwa
 
     protected BaseWorkspaceService(IBaseTextDocumentService documentService) {
         this.documentService = documentService;
-        documentService.pair(this);
     }
 
 
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseWorkspaceService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseWorkspaceService.java
@@ -37,6 +37,7 @@ import org.eclipse.lsp4j.DidChangeConfigurationParams;
 import org.eclipse.lsp4j.DidChangeWatchedFilesParams;
 import org.eclipse.lsp4j.DidChangeWorkspaceFoldersParams;
 import org.eclipse.lsp4j.ExecuteCommandParams;
+import org.eclipse.lsp4j.FileOperationsServerCapabilities;
 import org.eclipse.lsp4j.ServerCapabilities;
 import org.eclipse.lsp4j.WorkspaceFolder;
 import org.eclipse.lsp4j.WorkspaceFoldersOptions;
@@ -54,7 +55,7 @@ public class BaseWorkspaceService implements WorkspaceService, LanguageClientAwa
     private final CopyOnWriteArrayList<WorkspaceFolder> workspaceFolders = new CopyOnWriteArrayList<>();
 
 
-    BaseWorkspaceService(IBaseTextDocumentService documentService) {
+    protected BaseWorkspaceService(IBaseTextDocumentService documentService) {
         this.documentService = documentService;
         documentService.pair(this);
     }
@@ -69,10 +70,12 @@ public class BaseWorkspaceService implements WorkspaceService, LanguageClientAwa
         var clientWorkspaceCap = clientCap.getWorkspace();
 
         if (clientWorkspaceCap != null && Boolean.TRUE.equals(clientWorkspaceCap.getWorkspaceFolders())) {
-            var workspaceCap = new WorkspaceFoldersOptions();
-            workspaceCap.setSupported(true);
-            workspaceCap.setChangeNotifications(true);
-            capabilities.setWorkspace(new WorkspaceServerCapabilities(workspaceCap));
+            var workspaceOpts = new WorkspaceFoldersOptions();
+            workspaceOpts.setSupported(true);
+            workspaceOpts.setChangeNotifications(true);
+            var workspaceCap = new WorkspaceServerCapabilities(workspaceOpts);
+            workspaceCap.setFileOperations(new FileOperationsServerCapabilities());
+            capabilities.setWorkspace(workspaceCap);
         }
 
     }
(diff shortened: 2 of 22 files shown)


rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Util.rsc
```diff
@@ -129,3 +130,9 @@ bool(&T, &T) desc(bool(&T, &T) f) {
         return f(t2, t1);
     };
 }
+
+set[&T] flatMap(set[&S] ss, set[&T](&S) f) =
```
DavyLandman, line 134: might be more compact to write as a reducer, as then we can use set concat?

```rascal
= ({} | it + f(s) | s <- ss);
```

(internally, there is __some__ performance gain of writing a single set comprehension, but I would only do that if it's in the profiles.)

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/rename/Modules.rsc
```diff
@@ -0,0 +1,92 @@
...
+import util::FileSystem;
+import util::Reflective;
+
+private tuple[str, loc] fullQualifiedName(QualifiedName qn) = <"<qn>", qn.src>;
+private tuple[str, loc] qualifiedPrefix(QualifiedName qn) {
+    list[Name] names = [n | n <- qn.names];
+    if (size(names) <= 1) return <"", |unknown:///|>;
+
+    str fullName = "<qn>";
+    str namePrefix = substring(fullName, 0, findLast(fullName, "::"));
```
DavyLandman, line 50: rascal strings support for ranges, so we can write this as `fullName[..findLast(fullName,"::")]`

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/rename/Modules.rsc
```diff
@@ -0,0 +1,92 @@
...
+}
+
+private bool isReachable(PathConfig toProject, PathConfig fromProject) =
+    toProject == fromProject           // Both configs belong to the same project
+ || toProject.bin in fromProject.libs; // The using project can import the declaring project
+
+list[TextEdit] getChanges(loc f, PathConfig wsProject, rel[str oldName, str newName, PathConfig pcfg] qualifiedNameChanges) {
+    list[TextEdit] changes = [];
+
+    start[Module] m = parseModuleWithSpacesCached(f);
```
DavyLandman, line 63: this doesn't get the file contents from the open editors, of which we actually have the trees. Should we find a way to provide them? or do we wait for the changes that solve #355 to get merged?

DavyLandman (changes requested):
Looks good, with some exceptions that I marked down.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
```diff
@@ -0,0 +1,113 @@
...
+        super.initialize(clientCap, currentWorkspaceFolders, capabilities);
+
+        capabilities.getWorkspace().getFileOperations().setDidRename(new FileOperationOptions(List.of(new FileOperationFilter(new FileOperationPattern("**/*.rsc")))));
+    }
+
+    @Override
+    public void connect(LanguageClient client) {
+        super.connect(client);
+        this.client = client;
+        this.rascalServices = new RascalLanguageServices((RascalTextDocumentService) docService, this, (IBaseLanguageClient) client, ownExecuter);
```
DavyLandman, line 87: This means we start new evaluators for everything just for this workspace service, thats a bit excessive. we have to share the instance with the `RascalTextDocumentService`

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
```diff
@@ -0,0 +1,113 @@
...
+    private final ExecutorService ownExecuter;
+    private @MonotonicNonNull RascalLanguageServices rascalServices;
+    private @MonotonicNonNull FileFacts facts;
+    private @MonotonicNonNull LanguageClient client;
+
+    RascalWorkspaceService(ExecutorService exec, IBaseTextDocumentService documentService) {
+        super(documentService);
+        this.ownExecuter = exec;
+        this.docService = documentService;
+        new ColumnMaps(docService::getContents);
```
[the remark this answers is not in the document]
toinehartman: This was left over (from a suggested code fix). Removed.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
```diff
@@ -0,0 +1,113 @@
...
+
+        try {
+            rascalServices.getModuleRenames(params.getFiles(), workspaceFolders, facts::getPathConfig).get()
+                .thenApply(edits -> DocumentChanges.translateDocumentChanges(docService, edits))
+                .thenCompose(docChanges -> client.applyEdit(new ApplyWorkspaceEditParams(docChanges)))
+                .thenAccept(editResponse -> {
+                    if (!editResponse.isApplied()) {
+                        throw new RuntimeException("Applying module rename failed: " + editResponse.getFailureReason());
+                    }
+                }).join();
```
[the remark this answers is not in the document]
toinehartman: Thanks, this was indeed the wrong approach. Now using `exceptionally` instead, and got rid of the `try-catch` here as well.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseWorkspaceService.java
```diff
@@ -69,10 +69,12 @@ public void initialize(ClientCapabilities clientCap, @Nullable List<WorkspaceFol
...
         if (clientWorkspaceCap != null && Boolean.TRUE.equals(clientWorkspaceCap.getWorkspaceFolders())) {
-            var workspaceCap = new WorkspaceFoldersOptions();
-            workspaceCap.setSupported(true);
-            workspaceCap.setChangeNotifications(true);
-            capabilities.setWorkspace(new WorkspaceServerCapabilities(workspaceCap));
+            var workspaceOpts = new WorkspaceFoldersOptions();
+            workspaceOpts.setSupported(true);
+            workspaceOpts.setChangeNotifications(true);
+            var workspaceCap = new WorkspaceServerCapabilities(workspaceOpts);
+            workspaceCap.setFileOperations(new FileOperationsServerCapabilities());
```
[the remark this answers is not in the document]
toinehartman: I thought the creation of some of those objects here wouldn't hurt, but it might. Moved it.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IBaseTextDocumentService.java
```diff
@@ -44,5 +44,5 @@ public interface IBaseTextDocumentService extends TextDocumentService {
     void unregisterLanguage(LanguageParameter lang);
     CompletableFuture<IValue> executeCommand(String languageName, String command);
     LineColumnOffsetMap getColumnMap(ISourceLocation file);
-
+    String getContents(ISourceLocation file);
```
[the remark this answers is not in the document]
toinehartman: The reason I opted for this signature is because both implementing classes already had a function like this. But after some refactorings, this is not required anymore, so I'll revert to how it was.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -235,6 +240,81 @@ public InterruptibleFuture<ITuple> getRename(ITree module, Position cursor, Set<
...
+            .collect(Collectors.toList());
+
+        Set<ITuple> nameMapping = renames.stream()
+            .map(rename -> {
+                ISourceLocation currentLoc = sourceLocationFromUri(rename.getOldUri());
+                ISourceLocation newLoc = sourceLocationFromUri(rename.getNewUri());
+
+                ISourceLocation currentWsFolder = findContainingWorkspaceFolder(currentLoc, sortedWorkspaceFolders)
+                    .orElseThrow(() -> new ResponseErrorException(new ResponseError(ResponseErrorCode.InvalidParams,
+                        String.format("Cannot move %s, since that location is outside the current workspace", currentLoc), null)));
```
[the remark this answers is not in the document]
toinehartman: Sharp. Changed!

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
```diff
@@ -158,6 +166,13 @@ private String getContents(ISourceLocation file) {
         }
     }
 
+    Set<ISourceLocation> getFolderContents(ISourceLocation folder) {
```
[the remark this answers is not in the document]
toinehartman: Unused, removed

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -235,6 +240,81 @@ public InterruptibleFuture<ITuple> getRename(ITree module, Position cursor, Set<
...
+            .sorted((o1, o2) -> o1.toString().compareTo(o2.toString()))
+            .collect(Collectors.toList());
+
+        Set<ITuple> nameMapping = renames.stream()
+            .map(rename -> {
+                ISourceLocation currentLoc = sourceLocationFromUri(rename.getOldUri());
+                ISourceLocation newLoc = sourceLocationFromUri(rename.getNewUri());
+
+                ISourceLocation currentWsFolder = findContainingWorkspaceFolder(currentLoc, sortedWorkspaceFolders)
+                    .orElseThrow(() -> new ResponseErrorException(new ResponseError(ResponseErrorCode.InvalidParams,
```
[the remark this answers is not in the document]
toinehartman: Since this does not return a promise to the LSP and therefore this exception is not caught nicely, maybe all of these should just be `RuntimeException`

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -149,17 +152,21 @@ public void write(JsonWriter target, IValue value) throws IOException {
...
         if (DEPLOY_MODE) {
-            startLSP(constructLSPClient(capturedIn, capturedOut, new ActualLanguageServer(() -> System.exit(0), service.get())));
+            startLSP(constructLSPClient(capturedIn, capturedOut, new ActualLanguageServer(() -> System.exit(0), docService, wsService)));
         }
         else {
             try (ServerSocket serverSocket = new ServerSocket(portNumber, 0, InetAddress.getByName("127.0.0.1"))) {
                 logger.info("Rascal LSP server listens on port number: {}", portNumber);
                 while (true) {
-                    startLSP(constructLSPClient(serverSocket.accept(), new ActualLanguageServer(() -> {}, service.get())));
+                    startLSP(constructLSPClient(serverSocket.accept(), new ActualLanguageServer(() -> {}, docService, wsService)));
```
DavyLandman, line 169: Since this happens in a loop, the semantics have now changed.

This code exists so when developing 2nd level, you can restart VS Code, and then we automatically restart the LSP instance to get a fresh one.  So we should reconstruct them every time we start the loop.

toinehartman: Right. Fixed!

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -235,6 +240,81 @@ public InterruptibleFuture<ITuple> getRename(ITree module, Position cursor, Set<
...
+                        String.format("Cannot move file to %s, since that location is outside the current workspace", newLoc), null)));
+
+                if (!currentWsFolder.equals(newWsFolder)) {
+                    String commonProjPrefix = StringUtils.getCommonPrefix(currentWsFolder.toString(), newWsFolder.toString());
+                    String currentProject = StringUtils.removeStart(currentWsFolder.toString(), commonProjPrefix);
+                    String newProject = StringUtils.removeStart(newWsFolder.toString(), commonProjPrefix);
+
+                    throw new ResponseErrorException(new ResponseError(ResponseErrorCode.RequestFailed,
+                        String.format("Moving files between projects (from %s to %s) is not supported", currentProject, newProject), null));
```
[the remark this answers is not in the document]
toinehartman: The tricky thing is that this could require adding a dependency on the project the the module was moved to. We could consider supporting this when this dependency already exists, or prompt the user fix the dependencies themselves.

https://github.com/usethesource/rascal-language-servers/issues/542

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -235,6 +245,115 @@ public InterruptibleFuture<ITuple> getRename(ITree module, Position cursor, Set<
...
+                writer.insert(VF.tuple(currentName, newName, addResources(pcfg)));
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+        }
+
+        return writer.done();
+    }
+
+    private static String readFile(ISourceLocation loc, String charset) throws IOException {
```
DavyLandman, line 301: This is almost a copy from `ParametricTextDocumentService::getContents` with the exception that it doesn't look at the editor state, and doesn't use IOUtils but has the classic copy loop.

Is there a reason to keep this like this?

I would also advice against passing in the charset (unless we really know). In general, let the uri resolver take care of the charset.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -235,6 +245,115 @@ public InterruptibleFuture<ITuple> getRename(ITree module, Position cursor, Set<
...
+    private CompletableFuture<ITree> getEditorTreeOrParse(ISourceLocation loc, Map<ISourceLocation, TextDocumentState> documents) throws IOException {
+        if(documents.containsKey(loc)) {
+            var file = documents.get(loc);
+            return file.getCurrentTreeAsync()
+                .thenApply(Versioned::get)
+                .handle((t, r) -> (t == null ? file.getMostRecentTree().get() : t))
+                ;
+        }
+
+        var input = readFile(loc, "UTF-8");
```
DavyLandman, line 323: we don't know if the file is encoded in utf8. and always better to use `StandardCharsets.UTF8` instead of a string literal.

also, we should read the file inside the `CompletableFuture`, not outside of it.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -235,6 +245,115 @@ public InterruptibleFuture<ITuple> getRename(ITree module, Position cursor, Set<
...
+        if (fileRenames.isEmpty()) {
+            return InterruptibleFuture.completedFuture(emptyResult);
+        }
+
+        final ISet qualifiedNameChanges = qualfiedNameChangesFromRenames(fileRenames, workspaceFolders, getPathConfig);
+
+        return runEvaluator("Rascal module rename", semanticEvaluator, eval -> {
+            IFunction rascalGetPathConfig = eval.getFunctionValueFactory().function(getPathConfigType, (t, u) -> addResources(getPathConfig.apply((ISourceLocation) t[0])));
+            final var treeType = tf.abstractDataType(store, "Tree");
+            IFunction rascalGetModuleTree = eval.getFunctionValueFactory().function(tf.functionType(treeType, tf.tupleType(tf.sourceLocationType()), tf.tupleEmpty()), (args, kwArgs) -> {
```
DavyLandman, line 338: I get why this is here, so that we can look at unsaved state. But since the fix for #355 is on it's way, isn't this a bit of overkill? (especially since the rest of the rename code also doesn't work with unsaved changes I think?)

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
```diff
@@ -188,6 +188,31 @@ describe('IDE', function () {
         expect(editorText).to.contain("i -2");
     });
 
+    it("renaming files works", async() => {
```
DavyLandman, line 191: nice test :+1:

DavyLandman (changes requested):
Some small changes, looks mostly fine 👍🏼

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
toinehartman: I will remove the explicit charset. Otherwise, this is more or less copied from `Prelude.readFile`, like you suggested. But I agree; the approach in `ParametricTextDocumentService::getContents` is a lot more readable. Will use that instead.

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
toinehartman: Had to add this, since I think consistently verifying at least basic functionality of this callback is difficult otherwise.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -235,6 +240,79 @@ public InterruptibleFuture<ITuple> getRename(ITree module, Position cursor, Set<
...
+        return writer.done();
+    }
+
+    public InterruptibleFuture<ITuple> getModuleRenames(List<FileRename> fileRenames, Set<ISourceLocation> workspaceFolders, Function<ISourceLocation, PathConfig> getPathConfig) {
+        var emptyResult = VF.tuple(VF.list(), VF.map());
+        if (fileRenames.isEmpty()) {
+            return InterruptibleFuture.completedFuture(emptyResult);
+        }
+
+        final ISet qualifiedNameChanges = qualfiedNameChangesFromRenames(fileRenames, workspaceFolders, getPathConfig);
```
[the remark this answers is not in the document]
toinehartman: Thanks. I do understand that especially path config computation could be expensive and should not lock the thread. Fixed.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
```diff
@@ -335,6 +387,24 @@ public CompletableFuture<List<FoldingRange>> foldingRange(FoldingRangeRequestPar
...
+
+        rascalServices.getModuleRenames(params.getFiles(), workspaceFolders, facts::getPathConfig).get()
+            .thenApply(edits -> DocumentChanges.translateDocumentChanges(this, edits))
+            .thenCompose(docChanges -> client.applyEdit(new ApplyWorkspaceEditParams(docChanges)))
+            .thenAccept(editResponse -> {
+                if (!editResponse.isApplied()) {
+                    throw new RuntimeException("Applying module rename failed: " + editResponse.getFailureReason());
+                }
+            })
+            .exceptionally(e -> {
```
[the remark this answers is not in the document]
toinehartman: Correct. Checked this with a forced error. Now fixed.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
toinehartman: Alright. Removed this.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -235,6 +246,84 @@ public InterruptibleFuture<ITuple> getRename(ITree module, Position cursor, Set<
...
+
+    public InterruptibleFuture<ITuple> getModuleRenames(List<FileRename> fileRenames, Set<ISourceLocation> workspaceFolders, Function<ISourceLocation, PathConfig> getPathConfig, Map<ISourceLocation, TextDocumentState> documents) {
+        var emptyResult = VF.tuple(VF.list(), VF.map());
+        if (fileRenames.isEmpty()) {
+            return InterruptibleFuture.completedFuture(emptyResult);
+        }
+
+        return runEvaluator("Rascal module rename", semanticEvaluator, eval -> {
+            IFunction rascalGetPathConfig = eval.getFunctionValueFactory().function(getPathConfigType, (t, u) -> addResources(getPathConfig.apply((ISourceLocation) t[0])));
+            ISet qualifiedNameChanges = qualfiedNameChangesFromRenames(fileRenames, workspaceFolders, getPathConfig);
```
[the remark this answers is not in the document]
toinehartman: Done. Now this function returns a `CompletableFuture` instead of an `InterruptibleFuture`. We still call `InterruptibleFuture::get`, just in a different function.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Override
+    public void didRenameFiles(RenameFilesParams params) {
+        logger.debug("workspace/didRenameFiles: {}", params.getFiles());
+
+        CompletableFuture.supplyAsync(() -> workspaceFolders()
+            .stream()
+            .map(f -> Locations.toLoc(f.getUri()))
+            .collect(Collectors.toSet()), getExecuter())
+                .thenAccept(folders -> ((RascalTextDocumentService) docService).didRenameFiles(params, folders));
```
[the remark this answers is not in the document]
toinehartman: Thanks. This indent was enforced by checkstyle, but now I understand that I might have confused checkstyle with where I put the parentheses. Much better now.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -235,6 +233,36 @@ public InterruptibleFuture<ITuple> getRename(ITree module, Position cursor, Set<
         }, VF.tuple(VF.list(), VF.map()), exec, false, client);
     }
 
+    private ISourceLocation sourceLocationFromUri(String uri) {
```
DavyLandman, line 236: We should add this (in a different PR) to the `URIUtil` in rascal, and call it `assumeCorrectLocation` (to mirror the `assumeCorrect` in that helper class).

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -235,6 +233,36 @@ public InterruptibleFuture<ITuple> getRename(ITree module, Position cursor, Set<
...
+    public CompletableFuture<ITuple> getModuleRenames(List<FileRename> fileRenames, Set<ISourceLocation> workspaceFolders, Function<ISourceLocation, PathConfig> getPathConfig, Map<ISourceLocation, TextDocumentState> documents) {
+        var emptyResult = VF.tuple(VF.list(), VF.map());
+        if (fileRenames.isEmpty()) {
+            return CompletableFuture.completedFuture(emptyResult);
+        }
+
+        return CompletableFuture.supplyAsync(() -> fileRenames.stream()
+                .map(r -> VF.tuple(sourceLocationFromUri(r.getOldUri()), sourceLocationFromUri(r.getNewUri())))
+                .collect(VF.listWriter())
+            )
```
DavyLandman, line 253: yes! much improved, but please pass in our `exec` as 2nd argument for this `supplyAsync` just to make sure we're running on the same threadpool (which makes it nicer for shutdown).

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
```diff
@@ -285,6 +296,56 @@ public CompletableFuture<Either<List<? extends Location>, List<? extends Locatio
             ;
     }
 
+    private ITree getCursorFromPosition(ISourceLocation file, ITree moduleTree, Position p) {
+        Position rascalCursorPos = Locations.toRascalPosition(file, p, columns);
+
+        // Find all trees containing the cursor, in ascending order of size
+        IList focusList = TreeSearch.computeFocusList(moduleTree, rascalCursorPos.getLine(), rascalCursorPos.getCharacter());
+        List<String> sortNames = focusList.stream()
+            .map(tree -> ProductionAdapter.getSortName(TreeAdapter.getProduction((ITree) tree)))
```
DavyLandman, line 305: nit: can be split in 2 (or 3) maps without any function closures:

```java
.map(ITree.class::cast)
.map(TreeAdapter::getProduction)
.map(ProductionAdapter::getSortName)
```

if this is better, I'm not always sure, but it does lead to less long lines?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
```diff
@@ -285,6 +296,56 @@ public CompletableFuture<Either<List<? extends Location>, List<? extends Locatio
             ;
     }
 
+    private ITree getCursorFromPosition(ISourceLocation file, ITree moduleTree, Position p) {
```
DavyLandman, line 299: maybe this function should get a better name? something like `findQualifiedNameUnderCursor`?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
```diff
@@ -0,0 +1,93 @@
...
+    public void didRenameFiles(RenameFilesParams params) {
+        logger.debug("workspace/didRenameFiles: {}", params.getFiles());
+
+        CompletableFuture.supplyAsync(() ->
+            workspaceFolders()
+                .stream()
+                .map(f -> Locations.toLoc(f.getUri()))
+                .collect(Collectors.toSet())
+            , getExecuter()
+        ).thenAccept(folders -> ((RascalTextDocumentService) docService).didRenameFiles(params, folders));
```
DavyLandman, line 91: nit: we should move this cast to the constructor, and then throw an exception if we get a non `RascalTextDocumentService` as argument.

DavyLandman (approved):
Some minor comments, but looks good. Also thinks @sungshik for the help with the UI tests.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
toinehartman: Agreed, replaced

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
toinehartman: Since we now delegate to a function of the interface, we don't need this case at all. But perhaps, we should still check the instance type in the constructor?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
toinehartman: https://github.com/usethesource/rascal/pull/2113

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
DavyLandman: well, if we don't need a cast, let's not cast indeed :) and don't add more requirements than needed, it'll only hurts us later.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
DavyLandman: tnx 👍🏼

DavyLandman:
@toinehartman looks like it's ready to merge, want to do the honors?

toinehartman:
After off-line discussion, we reasoned that the issue is mostly solved by reordering the edits that renamings produce. Some async behaviour will remaing due to the nature of the LSP, and is acceptable. Merging.

PR #543, opened 2024-12-13
Avoid registering the same scheme multiple times

urbanfly:
When multiple `ParameterizedLanguageServer` instances are created in the same extension, the same schemes are registered multiple times - failing.

This PR avoids registering a scheme multiple times by filtering the previously registered schemes.

```
rejected promise not handled within 1 second: Error: a provider for the scheme 'zip' is already registered
extensionHostProcess.js:155
stack trace: Error: a provider for the scheme 'zip' is already registered
    at o.registerFileSystemProvider (...\AppData\Local\Programs\Microsoft VS Code\resources\app\out\vs\workbench\api\node\extensionHostProcess.js:161:94201)
    at Object.registerFileSystemProvider (...\AppData\Local\Programs\Microsoft VS Code\resources\app\out\vs\workbench\api\node\extensionHostProcess.js:171:48861)
    at ...\node_modules\@usethesource\rascal-vscode-dsl-lsp-server\lib\fs\RascalFileSystemProviders.js:51:1
    at Array.forEach (<anonymous>)
    at RascalFileSystemProvider.registerSchemes (...\node_modules\@usethesource\rascal-vscode-dsl-lsp-server\lib\fs\RascalFileSystemProviders.js:51:1)
    at ...\node_modules\@usethesource\rascal-vscode-dsl-lsp-server\lib\lsp\RascalLSPConnection.js:56:1
    at runNextTicks (node:internal/process/task_queues:60:5)
    at processImmediate (node:internal/timers:449:9)
    at process.callbackTrampoline (node:internal/async_hooks:130:17)
```

[VS Code API docs](https://code.visualstudio.com/api/references/vscode-api#workspace) say:

> There can only be one provider per scheme and an error is being thrown when a scheme has been claimed by another provider or when it is reserved.

### DIFF
--- a/rascal-vscode-extension/src/fs/RascalFileSystemProviders.ts
+++ b/rascal-vscode-extension/src/fs/RascalFileSystemProviders.ts
@@ -47,19 +47,30 @@ export class RascalFileSystemProvider implements vscode.FileSystemProvider {
         });
     }
 
-
-
-    registerSchemes(schemes:string[]):void {
-        schemes
+    /**
+     * Attemptes to register all schemes.
+     * @param schemes The list of schemes to register for this provider
+     * @returns `true` if all schemes were registered successfully; `false` if any schemes failed to register.
+     */
+    tryRegisterSchemes(schemes: string[]): boolean {
+        return schemes
             .filter(s => !this.protectedSchemes.includes(s))
+            // we add support for schemes that look inside a jar
+            .concat(schemes
+                .filter(s => s !== "jar" && s !== "zip" && s !== "compressed")
+                .map(s => "jar+" + s))
             .filter(isUnknownFileSystem)
-            .forEach(s => vscode.workspace.registerFileSystemProvider(s, this));
-        // we add support for schemes that look inside a jar
-        schemes
-            .filter(s => s !== "jar" && s !== "zip" && s !== "compressed")
-            .map(s => "jar+" + s)
-            .filter(isUnknownFileSystem)
-            .forEach(s => vscode.workspace.registerFileSystemProvider(s, this));
+            .map(s => {
+                try {
+                    vscode.workspace.registerFileSystemProvider(s, this);
+                    this.client.info(`Scheme registered: ${s}`);
+                    return true;
+                } catch (error) {
+                    this.client.error(`Unable to register scheme: ${s}\n${error}`);
+                    return false;
+                }
+            })
+            .every(b => b);
     }
 
     watch(uri: vscode.Uri, options: { recursive: boolean; excludes: string[]; }): vscode.Disposable {
--- a/rascal-vscode-extension/src/fs/RascalFileSystemProviders.ts
+++ b/rascal-vscode-extension/src/fs/RascalFileSystemProviders.ts
@@ -63,11 +63,17 @@ export class RascalFileSystemProvider implements vscode.FileSystemProvider {
             .map(s => {
                 try {
                     vscode.workspace.registerFileSystemProvider(s, this);
-                    this.client.info(`Scheme registered: ${s}`);
+                    this.client.debug(`Rascal VFS registered scheme: ${s}`);
                     return true;
                 } catch (error) {
-                    this.client.error(`Unable to register scheme: ${s}\n${error}`);
-                    return false;
+                    if (isUnknownFileSystem(s)) {
+                        this.client.error(`Unable to register scheme: ${s}\n${error}`);
+                        return false;
+                    }
+                    else {
+                        this.client.debug(`Rascal VFS lost the race to register scheme: ${s}, which in most cases is fine`);
+                        return true;
+                    }
                 }
             })
             .every(b => b);
--- a/rascal-vscode-extension/src/fs/RascalFileSystemProviders.ts
+++ b/rascal-vscode-extension/src/fs/RascalFileSystemProviders.ts
@@ -50,33 +50,28 @@ export class RascalFileSystemProvider implements vscode.FileSystemProvider {
     /**
      * Attemptes to register all schemes.
      * @param schemes The list of schemes to register for this provider
-     * @returns `true` if all schemes were registered successfully; `false` if any schemes failed to register.
      */
-    tryRegisterSchemes(schemes: string[]): boolean {
-        return schemes
+    tryRegisterSchemes(schemes: string[]) {
+        schemes
             .filter(s => !this.protectedSchemes.includes(s))
             // we add support for schemes that look inside a jar
             .concat(schemes
                 .filter(s => s !== "jar" && s !== "zip" && s !== "compressed")
                 .map(s => "jar+" + s))
             .filter(isUnknownFileSystem)
-            .map(s => {
+            .forEach(s => {
                 try {
                     vscode.workspace.registerFileSystemProvider(s, this);
                     this.client.debug(`Rascal VFS registered scheme: ${s}`);
-                    return true;
                 } catch (error) {
                     if (isUnknownFileSystem(s)) {
                         this.client.error(`Unable to register scheme: ${s}\n${error}`);
-                        return false;
                     }
                     else {
                         this.client.debug(`Rascal VFS lost the race to register scheme: ${s}, which in most cases is fine`);
-                        return true;
                     }
                 }
-            })
-            .every(b => b);
+            });
     }
 
     watch(uri: vscode.Uri, options: { recursive: boolean; excludes: string[]; }): vscode.Disposable {
--- a/rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
+++ b/rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
@@ -63,10 +63,7 @@ export async function activateLanguageClient(
 
     schemesReply.then( schemes => {
         vfsServer.ignoreSchemes(schemes);
-        const allRegistered = new RascalFileSystemProvider(client).tryRegisterSchemes(schemes);
-        if (!allRegistered) {
-            client.warn(`At least one of the expected schemes failed to register.\nExpected schemes: ${schemes}`);
-        }
+        new RascalFileSystemProvider(client).tryRegisterSchemes(schemes);
     });
 
     return client;
--- a/rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
+++ b/rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
@@ -62,8 +62,12 @@ export async function activateLanguageClient(
     const schemesReply = client.sendRequest<string[]>("rascal/filesystem/schemes");
 
     schemesReply.then( schemes => {
-        vfsServer.ignoreSchemes(schemes);
-        new RascalFileSystemProvider(client).registerSchemes(schemes);
+        // avoid registering the same schemes multiple times
+        const knownRascalNativeSchemes = vfsServer.getIgnoredSchemes();
+        const remainingSchemes = schemes.filter(s => !knownRascalNativeSchemes.includes(s));
+        if (remainingSchemes.length) {
+            new RascalFileSystemProvider(client).registerSchemes(schemes);
+        }
     });
 
 
--- a/rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
+++ b/rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
@@ -65,7 +65,7 @@ export async function activateLanguageClient(
         vfsServer.ignoreSchemes(schemes);
         const allRegistered = new RascalFileSystemProvider(client).tryRegisterSchemes(schemes);
         if (!allRegistered) {
-            client.warn("At least one of the expected schemes failed to register.", { expectedSchemes: schemes });
+            client.warn(`At least one of the expected schemes failed to register.\nExpected schemes: ${schemes}`);
         }
     });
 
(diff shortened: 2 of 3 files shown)


DavyLandman:
Yes this is an annoying error message, but I don't think this is the right fix. As it's in essence a race between parallel registrations of the VFS on the VS Code side.

In regular rascal mode:

1. Rascal
2. parametic server

most cases, 1. wins (as it tends to get started first). you don't see this message

In deployed mode:

1. parametric server DSL 1
2. parametric server DSL 2
3. ...

These are all started in parallel, and the 2 instences don't know of each other, or at least, they don't know if they're the only one thats there.

(there is also the edge case of where a scheme might already by another vs code extensions, but let's ignore that).

So, to make sure rascal schemes work in VS Code (so you can open a `lib://...` or a `|project://..` URI in VS Code and the contents are correct) we have to register which rascal schemes are available to VS Code. So we call the `registerFileSystemProvider` fuction in VS Code, and forward it to the java process. The function notes:

> There can only be one provider per scheme and an error is being thrown when a scheme has been claimed by another provider or when it is reserved.

And the only way we could check if maybe someone else already did register a certain scheme is the `isWritableFileSystem` function, which returns true/false/undefined. where undefined means: we don't know about this scheme. This could still get into a race, but would reduce the chances of it.

I see 2 strategies to get rid of this message:

1. check `isWritableFileSystem` just before `registerFileSystemProvider`
2. swallow the exception that `registerFileSystemProvider` throws.

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -62,8 +62,13 @@ export async function activateLanguageClient(
...
 
     schemesReply.then( schemes => {
-        vfsServer.ignoreSchemes(schemes);
-        new RascalFileSystemProvider(client).registerSchemes(schemes);
+        // avoid registering the same schemes multiple times
+        const knownRascalNativeSchemes = vfsServer.getIgnoredSchemes();
+        const remainingSchemes = schemes.filter(s => !knownRascalNativeSchemes.includes(s));
+        if (remainingSchemes.length) {
+            vfsServer.ignoreSchemes(remainingSchemes);
+            new RascalFileSystemProvider(client).registerSchemes(schemes);
```
DavyLandman, line 70: Should it not be `remainingSchemes`?

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -62,8 +62,13 @@ export async function activateLanguageClient(
     const schemesReply = client.sendRequest<string[]>("rascal/filesystem/schemes");
 
     schemesReply.then( schemes => {
-        vfsServer.ignoreSchemes(schemes);
-        new RascalFileSystemProvider(client).registerSchemes(schemes);
+        // avoid registering the same schemes multiple times
+        const knownRascalNativeSchemes = vfsServer.getIgnoredSchemes();
```
DavyLandman, line 66: the function name `getIgnoredSchemes` is confusing. (especially in context to the field & the variable name).

DavyLandman (changes requested):
While it's an okay workaround, and will indeed reduce the chances of the exception appearing, I think we should do something that's also save against other kinds of collisions (see my comment)

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
urbanfly: Agree. The internal names and the public function `ignoreSchemes` is already awkward.

urbanfly:
> but not one that actually hurts us

It hurts a little bit. We are in the process of adding telemetry and these constantly show up as `unhandlederror` events. There is no way (that I know of) to avoid these errors as a consumer of this package.

urbanfly:
> Are you okay updating the PR with these fixes?

Yes, I will.

urbanfly:
> So, to make sure rascal schemes work in VS Code (so you can open a `lib://...` or a `|project://..` URI in VS Code and the contents are correct) we have to register which rascal schemes are available to VS Code. So we call the `registerFileSystemProvider` function in VS Code, and forward it to the java process.

Can this be an optional feature that the consumer can opt out of w/ a new parameter on `ParameterizedLanguageServer`?

Some ext. probably don't need this ability. And if there are multiple `ParameterizedLanguageServer` and only one of them has registered the scheme - will it work reliably anyway?

urbanfly:
> I see 2 strategies to get rid of this message:
> 
> 1. check `isWritableFileSystem` just before `registerFileSystemProvider`
> 2. swallow the exception that `registerFileSystemProvider` throws.

`isWritableFileSystem` is already being checked just before `registerFileSystemProvider`

https://github.com/usethesource/rascal-language-servers/blob/3c5ff77797c604c0d525e2eb7ad25ebea0ce4d74/rascal-vscode-extension/src/fs/RascalFileSystemProviders.ts#L53-L56

https://github.com/usethesource/rascal-language-servers/blob/3c5ff77797c604c0d525e2eb7ad25ebea0ce4d74/rascal-vscode-extension/src/fs/RascalFileSystemProviders.ts#L116-L118

DavyLandman:
> Can this be an optional feature that the consumer can opt out of w/ a new parameter on `ParameterizedLanguageServer`?

The cases that break would be very subtle, I rather we don't get into that. It really depends on the kind of locs you're using, and sending to VS Code. 




> `isWritableFileSystem` is already being checked just before `registerFileSystemProvider`

Ah right, I thought I did something like that already. Okay, so it's a proper race now, between that check and the subsequent register. So just catch the exception is the wisest thing, we could even skip that whole preflight check and just catch the error?

That should solve it once and or all.

urbanfly:
I changed the PR to now attempt to register all schemes and log (info) when they succeed and log (error) when they fail. These log messages show up in the language-specific OutputChannel.

<details><summary>Example of log output</summary>
<p>

```log
[Info  - 4:28:30 PM] Scheme registered: zip
[Info  - 4:28:30 PM] Scheme registered: std
[Info  - 4:28:30 PM] Scheme registered: memory
[Info  - 4:28:30 PM] Scheme registered: lib
[Info  - 4:28:30 PM] Scheme registered: jar
[Info  - 4:28:30 PM] Scheme registered: compressed
[Info  - 4:28:30 PM] Scheme registered: PATH
[Info  - 4:28:30 PM] Scheme registered: cwd
[Info  - 4:28:30 PM] Scheme registered: tmp
[Info  - 4:28:30 PM] Scheme registered: project
[Info  - 4:28:30 PM] Scheme registered: home
[Info  - 4:28:30 PM] Scheme registered: target
[Info  - 4:28:30 PM] Scheme registered: jar+std
[Info  - 4:28:30 PM] Scheme registered: jar+file
[Info  - 4:28:30 PM] Scheme registered: jar+memory
[Info  - 4:28:30 PM] Scheme registered: jar+lib
[Info  - 4:28:30 PM] Scheme registered: jar+http
[Info  - 4:28:30 PM] Scheme registered: jar+https
[Info  - 4:28:30 PM] Scheme registered: jar+PATH
[Info  - 4:28:30 PM] Scheme registered: jar+cwd
[Info  - 4:28:30 PM] Scheme registered: jar+tmp
[Info  - 4:28:30 PM] Scheme registered: jar+project
[Info  - 4:28:30 PM] Scheme registered: jar+home
[Info  - 4:28:30 PM] Scheme registered: jar+target
```

or

```log
[Error - 4:28:30 PM] Unable to register scheme: zip
Error: a provider for the scheme 'zip' is already registered
[Error - 4:28:30 PM] Unable to register scheme: std
Error: a provider for the scheme 'std' is already registered
[Error - 4:28:30 PM] Unable to register scheme: memory
Error: a provider for the scheme 'memory' is already registered
[Error - 4:28:30 PM] Unable to register scheme: lib
Error: a provider for the scheme 'lib' is already registered
[Error - 4:28:30 PM] Unable to register scheme: jar
Error: a provider for the scheme 'jar' is already registered
[Error - 4:28:30 PM] Unable to register scheme: compressed
Error: a provider for the scheme 'compressed' is already registered
[Error - 4:28:30 PM] Unable to register scheme: PATH
Error: a provider for the scheme 'PATH' is already registered
[Error - 4:28:30 PM] Unable to register scheme: cwd
Error: a provider for the scheme 'cwd' is already registered
[Error - 4:28:30 PM] Unable to register scheme: tmp
Error: a provider for the scheme 'tmp' is already registered
[Error - 4:28:30 PM] Unable to register scheme: project
Error: a provider for the scheme 'project' is already registered
[Error - 4:28:30 PM] Unable to register scheme: home
Error: a provider for the scheme 'home' is already registered
[Error - 4:28:30 PM] Unable to register scheme: target
Error: a provider for the scheme 'target' is already registered
[Error - 4:28:30 PM] Unable to register scheme: jar+std
Error: a provider for the scheme 'jar+std' is already registered
[Error - 4:28:30 PM] Unable to register scheme: jar+file
Error: a provider for the scheme 'jar+file' is already registered
[Error - 4:28:30 PM] Unable to register scheme: jar+memory
Error: a provider for the scheme 'jar+memory' is already registered
[Error - 4:28:30 PM] Unable to register scheme: jar+lib
Error: a provider for the scheme 'jar+lib' is already registered
[Error - 4:28:30 PM] Unable to register scheme: jar+http
Error: a provider for the scheme 'jar+http' is already registered
[Error - 4:28:30 PM] Unable to register scheme: jar+https
Error: a provider for the scheme 'jar+https' is already registered
[Error - 4:28:30 PM] Unable to register scheme: jar+PATH
Error: a provider for the scheme 'jar+PATH' is already registered
[Error - 4:28:30 PM] Unable to register scheme: jar+cwd
Error: a provider for the scheme 'jar+cwd' is already registered
[Error - 4:28:30 PM] Unable to register scheme: jar+tmp
Error: a provider for the scheme 'jar+tmp' is already registered
[Error - 4:28:30 PM] Unable to register scheme: jar+project
Error: a provider for the scheme 'jar+project' is already registered
[Error - 4:28:30 PM] Unable to register scheme: jar+home
Error: a provider for the scheme 'jar+home' is already registered
[Error - 4:28:30 PM] Unable to register scheme: jar+target
Error: a provider for the scheme 'jar+target' is already registered
[Warn  - 4:28:30 PM] At least one of the expected schemes failed to register.
```

</p>
</details> 

This _does_ change the behavior slightly - when the original code failed to register `zip`, it would not attempt to register any other schemes; the async continuation would throw and be unhandled. This change will attempt to register all schemes - even if one of them fails.

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -63,10 +63,12 @@ export async function activateLanguageClient(
 
     schemesReply.then( schemes => {
         vfsServer.ignoreSchemes(schemes);
-        new RascalFileSystemProvider(client).registerSchemes(schemes);
+        const allRegistered = new RascalFileSystemProvider(client).tryRegisterSchemes(schemes);
+        if (!allRegistered) {
+            client.warn(`At least one of the expected schemes failed to register.\nExpected schemes: ${schemes}`);
```
DavyLandman, line 68: not sure if this should be a warning, as we expect races? So sometimes it happens, and that's fine?

rascal-vscode-extension/src/fs/RascalFileSystemProviders.ts
```diff
@@ -47,19 +47,30 @@ export class RascalFileSystemProvider implements vscode.FileSystemProvider {
...
-            .map(s => "jar+" + s)
-            .filter(isUnknownFileSystem)
-            .forEach(s => vscode.workspace.registerFileSystemProvider(s, this));
+            .map(s => {
+                try {
+                    vscode.workspace.registerFileSystemProvider(s, this);
+                    this.client.info(`Scheme registered: ${s}`);
+                    return true;
+                } catch (error) {
+                    this.client.error(`Unable to register scheme: ${s}\n${error}`);
```
DavyLandman, line 69: Can we tweak this a bit to reflect that most likely you can ignore this error?

rascal-vscode-extension/src/fs/RascalFileSystemProviders.ts
```diff
@@ -47,19 +47,30 @@ export class RascalFileSystemProvider implements vscode.FileSystemProvider {
...
+    /**
+     * Attemptes to register all schemes.
+     * @param schemes The list of schemes to register for this provider
+     * @returns `true` if all schemes were registered successfully; `false` if any schemes failed to register.
+     */
+    tryRegisterSchemes(schemes: string[]): boolean {
+        return schemes
             .filter(s => !this.protectedSchemes.includes(s))
+            // we add support for schemes that look inside a jar
+            .concat(schemes
```
DavyLandman, line 59: ah yes, that's nice :) just concat them 👍🏼

DavyLandman (changes requested):
Looks good, let's just try and reduce the logging output a bit? People get confused when there are many log messages that say something broke. (I fully understand it's useful for debugging this fix).

I like the behavior that we continue even if one failed, much nicer 👍🏼

rascal-vscode-extension/src/fs/RascalFileSystemProviders.ts
urbanfly: I see you made this change. Thanks. 👍

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
urbanfly: Considering that the other messages are now either debug on success or uninteresting failures, and only an error in a truly unexpected cases - and that this will only be false in truly unexpected cases - are you ok with leaving this as a warning?

DavyLandman: I was wondering about dropping this message or not. What's your opinion?

(I picked up the PR, since we're doing a small bugfix release, and wanted to have this one in there as well)

urbanfly: I'm ok with dropping it.

PR #544, opened 2024-12-16
Fix LSP document synchronization

rodinaarssen:
The LSP server is responsible for the content of files that are open in the editor. This was not (fully/correctly) implemented in rascal-lsp (#355). During resolution of a source location, the fallback resolver now rewrites source locations of open files, and file operations on such a location are handled by the new `LSPOpenFileResolver` class. This works for both Rascal and DSL files.

This depends on https://github.com/usethesource/rascal/pull/2101 to function correctly.


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
```diff
@@ -488,4 +490,14 @@ private static <T> CompletableFuture<T> recoverExceptions(CompletableFuture<T> f
                     return defaultValue.get();
                 });
     }
+
+    @Override
+    public boolean isManagingFile(ISourceLocation file) {
+        return facts.containsFile(file);
```
DavyLandman, line 496: why is this not looking at documents? (just like the Parametric one, where both look at the same container?)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/FallbackResolver.java
```diff
@@ -31,33 +31,55 @@
...
+
+    public static FallbackResolver getInstance() {
+        if (instance == null) {
+            throw new IllegalStateException("FallbackResolver accessed before initialization");
+        }
+        return instance;
+    }
+
+    public FallbackResolver() {
+        instance = this;
```
DavyLandman, line 81: I think we should add a small comment on why we're doing this

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/FallbackResolver.java
```diff
@@ -250,5 +272,45 @@ public void unwatch(ISourceLocation root, Consumer<ISourceLocationChanged> watch
...
+            }
+        }
+        return false;
+    }
 
+    @Override
+    public ISourceLocation resolve(ISourceLocation input) throws IOException {
+        if (isFileManaged(input)) {
+            try {
+                return URIUtil.changeScheme(input, "lsp+" + input.getScheme()).top();
```
DavyLandman, line 289: why the `.top`? that would loose information on the URI that we cannot recover?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/FallbackResolver.java
```diff
@@ -250,5 +272,45 @@ public void unwatch(ISourceLocation root, Consumer<ISourceLocationChanged> watch
...
+        }
+        return input;
+    }
+
+    @Override
+    public String authority() {
+        throw new UnsupportedOperationException("'authority' not supported by fallback resolver");
+    }
+
+    private Set<IBaseTextDocumentService> textDocumentServices = new HashSet<>();
```
DavyLandman, line 302: nit: should be final

notnit: not a threadsafe container, since we're only iterating over it, never clearing any, or using the key of it, an `CopyOnWriteArrayList` might be the better option.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/FallbackResolver.java
```diff
@@ -250,5 +272,45 @@ public void unwatch(ISourceLocation root, Consumer<ISourceLocationChanged> watch
...
+        textDocumentServices.add(service);
+    }
+
+    public TextDocumentState getDocumentState(ISourceLocation file) {
+        for (final var service : textDocumentServices) {
+            if (service.isManagingFile(file)) {
+                return service.getDocumentState(file);
+            }
+        }
+        return null;
```
DavyLandman, line 314: is there an alternative to `null`? should we throw an exception at this point? since `isFileManaged` is true?

what happens to a uri that was generated while the editor was open, but refenced after it is closed? do we unresolve it to the original URI?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileResolver.java
```diff
@@ -0,0 +1,104 @@
...
+import io.usethesource.vallang.ISourceLocation;
+
+public class LSPOpenFileResolver implements ISourceLocationInput {
+
+    @Override
+    public InputStream getInputStream(ISourceLocation uri) throws IOException {
+        var fallbackResolver = FallbackResolver.getInstance();
+        uri = stripLspPrefix(uri);
+        if (fallbackResolver.isFileManaged(uri)) {
+            return new ByteArrayInputStream(fallbackResolver.getDocumentState(uri).getCurrentContent().get().getBytes(StandardCharsets.UTF_16));
```
DavyLandman, line 48: There is a risk of a race between `isFileManaged`  and `getDocumentState`. Maybe it gets better if we merge them? Would also cut down on the map lookups.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileResolver.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    public boolean exists(ISourceLocation uri) {
+        return FallbackResolver.getInstance().isFileManaged(stripLspPrefix(uri));
+    }
+
+    @Override
+    public long lastModified(ISourceLocation uri) throws IOException {
+        return FallbackResolver.getInstance().getDocumentState(stripLspPrefix(uri)).getLastModified();
```
DavyLandman, line 65: can throw a NPE in case `getDocumentState` returns a `null`.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileResolver.java
```diff
@@ -0,0 +1,104 @@
...
+            } catch (URISyntaxException e) {
+                // fall through
+            }
+        }
+        return uri;
+    }
+
+    @Override
+    public String[] list(ISourceLocation uri) throws IOException {
+        throw new UnsupportedOperationException("Unimplemented method 'list'");
```
DavyLandman, line 91: let's turn this into an `IOException` that says you can't do this on a file?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Versioned.java
```diff
@@ -31,10 +31,12 @@
 public class Versioned<T> {
     private final int version;
     private final T object;
+    private final long timestamp;
 
     public Versioned(int version, T object) {
         this.version = version;
         this.object = object;
+        this.timestamp = System.currentTimeMillis();
```
DavyLandman, line 39: we have no guarantee that the versions are monotomic in their construction. so version 10 might be constructed on thread A at time 20, while version 9 got scheduled on thread B at time 22.

This gets especially messy when `tpl` files are around where we depend on the "newer" relationship. What I'm wondering is if we could, register the timestamp as close to the time we receive the `didChange` message? (it might be close to here, I haven't checked)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/Locations.java
```diff
@@ -41,6 +41,29 @@
 import io.usethesource.vallang.IValue;
 
 public class Locations {
+    public static ISourceLocation toClientLocation(ISourceLocation loc) throws IOException {
```
DavyLandman, line 44: I think we should add some notes on these functions to help future developer pick the right one, as to remain compatible with this redirection feature.

DavyLandman (commented):
Looks mostly good, I have some small concerns that I would like your input on.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/FallbackResolver.java
rodinaarssen: The offset/length information is restored by `resolveAndFixOffsets`. Not clearing it here would result in "double offset". I've added a comment

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Versioned.java
rodinaarssen: The timestamp of receiving a `didChange` message is now propagated.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileResolver.java
rodinaarssen: These have been merged

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/FallbackResolver.java
rodinaarssen: This has been rewritten.

We currently do not unresolve to the original URI. I'm not convinced we should as it might lead to unexpected lower timestamps?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/FallbackResolver.java
rodinaarssen: Fixed

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/FallbackResolver.java
rodinaarssen: Added a comment.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/FallbackResolver.java
```diff
@@ -250,5 +274,49 @@ public void unwatch(ISourceLocation root, Consumer<ISourceLocationChanged> watch
...
+
+    private final List<IBaseTextDocumentService> textDocumentServices = new CopyOnWriteArrayList<>();
+
+    public void registerTextDocumentService(IBaseTextDocumentService service) {
+        textDocumentServices.add(service);
+    }
+
+    public TextDocumentState getDocumentState(ISourceLocation file) throws IOException {
+        for (final var service : textDocumentServices) {
+            final var state = service.getDocumentState(file);
```
DavyLandman, line 315: nit: why are they marked final in this file? (also in the other loop?)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileResolver.java
DavyLandman: Has that been committed & pushed?

Since on 47 we do `isFileManaged` and on 48 we then do `getDocumentState`?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileResolver.java
DavyLandman: has that change been pushed? it still allows for the NPE to occur

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/Locations.java
```diff
@@ -35,12 +35,37 @@
 import org.eclipse.lsp4j.TextDocumentItem;
 import org.rascalmpl.uri.URIResolverRegistry;
 import org.rascalmpl.uri.URIUtil;
-import org.rascalmpl.values.IRascalValueFactory;
 
 import io.usethesource.vallang.ISourceLocation;
 import io.usethesource.vallang.IValue;
 
 public class Locations {
+    // Source locations of files that are opened in the IDE are redirected by prefixing the scheme with "lsp+".
```
DavyLandman, line 43: nit: maybe turn this into a java doc?

DavyLandman (commented):
Looks good, but I'm missing at least 1 test, just to help us not break this in the future.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileResolver.java
rodinaarssen: This was a oversight, thanks.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/FallbackResolver.java
rodinaarssen: No particular reason. They've been removed for a more uniform code base.

rodinaarssen:
This PR is on hold until a release of `rascal` containing usethesource/rascal#2101 is made, and a version of `rascal-core` compatible with it.

DavyLandman:
@rodinaarssen if you merge with the main branch, it should contain the fixes needed to finish this branch.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/Locations.java
```diff
@@ -35,12 +35,41 @@
...
+ * Source locations of files that are opened in the IDE are redirected by prefixing the scheme with "lsp+".
+ * These lsp-redirected source locations must not leak outside of the LSP server. The `toClientLocation` methods
+ * strip source locations of their "lsp+" prefix.
+ */
 public class Locations {
+    public static ISourceLocation toClientLocation(ISourceLocation loc) throws IOException {
+        var result = URIResolverRegistry.getInstance().logicalToPhysical(loc);
+        if (result.getScheme().startsWith("lsp+")) {
+            try {
+                result = URIUtil.changeScheme(result, result.getScheme().substring("lsp+".length()));
```
DavyLandman, line 54: looks like a clone of this extists in `LSPOpenFileResolver`? can we merge these (or perhaps call the function in `LSPOpenFileResolver` here?

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
```diff
@@ -275,4 +275,15 @@ describe('IDE', function () {
...
+
+    it("editor contents used for open files", async() => {
+        const importerEditor = await ide.openModule(TestWorkspace.importerFile);
+        const importeeEditor = await ide.openModule(TestWorkspace.importeeFile);
+
+        await importeeEditor.typeTextAt(3, 1, "public str foo;");
+        await ide.openModule(TestWorkspace.importerFile);
+
+        await ide.triggerTypeChecker(importerEditor, {waitForFinish : true});
+        await ide.hasErrorSquiggly(importerEditor);
```
DavyLandman, line 287: nice and clean test 👍🏼

DavyLandman (approved):
Very happy to see it works now :+1:

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/Locations.java
rodinaarssen: Good catch!

PR #551, opened 2024-12-24
Interact with type checker using module names.

toinehartman:
As discussed with @PaulKlint, this is a more robust way of interacting with the type checker. Internally, it uses names to represent modules, and only uses locations when TPLs are outdated.


DavyLandman (approved):
LGTM: however, the test seem to fail?

PR #553, opened 2024-12-27
Update example folder to package language servers as VS Code extensions

sungshik:
This PR fixes a few small things (and aligns with the current versions) in the example folder to package a language server as a VS Code extension.


DavyLandman:
Thanks!

PR #557, opened 2025-01-08
Fixed over-eager leading backslash removal in Rascal outline

### LINKED ISSUE #89: Rascal outline does not properly handle escaped characters in literals
image

rodinaarssen:
This solves an error in the outline of Rascal source files. Fixes #89.

### DIFF
--- a/rascal-lsp/src/main/rascal/lang/rascal/lsp/DocumentSymbols.rsc
+++ b/rascal-lsp/src/main/rascal/lang/rascal/lsp/DocumentSymbols.rsc
@@ -88,7 +88,7 @@ list[DocumentSymbol] documentRascalSymbols(start[Module] \mod) {
 }
 
 // remove leading backslash
-str clean(/\\<rest:.*>/) = clean(rest);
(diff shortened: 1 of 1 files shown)


DavyLandman:
What happens now in case of an newline char? I can hardly imagine this fixing it, it such a small fix ;)

rodinaarssen:
The regular expression that was changed in this PR intended to strip a leading backslash from the input, returning all characters trailing the backslash character. Instead, it would match against any backslash character in the input. New line characters are changed to a space in another overload of `clean`.

In the outline case, the string that is fed into `clean` is (cutting some corners), for a production `P`, `"<p.syms>"`. In the failing example from #89, this corresponds to the string `"\"BBB\\nBBB\""` (note that the backslash and `n` are separate characters here). The "leading backslash" `clean` overload would match against the `\\` at index 4 and incorrectly return `"nBBB\""`. Now, it correctly returns the input.

DavyLandman:
aah, right, it was for the case of an constructor named `\if` or `\while`. right :) thanks for explaining.

PR #559, opened 2025-01-10
Fix rename refactoring type errors

toinehartman:


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -215,7 +215,7 @@ public InterruptibleFuture<ITuple> getRename(ITree cursorTree, Set<ISourceLocati
             } catch (Throw e) {
                 if (e.getException() instanceof IConstructor) {
                     var exception = (IConstructor)e.getException();
-                    if (exception.getType().getAbstractDataType().getName().equals("RenameException")) {
+                    if (exception.getName().contains("Rename")) {
```
PieterOlivier, line 218: Is there anyway that now or in the future user code is called that could throw an exception? If so, is "having Rename" in the name discerning enough? Maybe we can also check the type including the "package" part to make sure it is an "official" Rename exception?

rascal-lsp/src/main/rascal/lang/rascal/tests/rename/Grammars.rsc
```diff
@@ -27,6 +27,7 @@ POSSIBILITY OF SUCH DAMAGE.
 module lang::rascal::tests::rename::Grammars
 
 import lang::rascal::tests::rename::TestUtils;
+import lang::rascal::lsp::refactor::Exception;
```
PieterOlivier, line 30: Extra import but no use of the import? The same goes for imports in `ProjectOnDisk.rsc`.

toinehartman: These imports fix errors regarding undefined `@expected` exceptions in these modules.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
toinehartman: I agree that this check is not very accurate. For now, however, I do not see a way to check for the origin of the exception.

toinehartman:
Some type errors remain in the tests. Those should be solved by the fix for https://github.com/usethesource/rascal/issues/2130.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
toinehartman: Reverted

PR #564, opened 2025-01-28
Add custom view for easy access to the Rascal debugger

rodinaarssen:
This PR adds a custom view to the "Run and Debug" view container in VS Code. This view provides users with a list of running Rascal REPLs within VS Code, with a button to start a debugging session for REPLs that have been initialized and are not already actively being debugged. The active Rascal terminal (if any) is highlighted for convenience.

For a visual impression, see the attached screenshot.

image

Each line represents a single active Rascal terminal. The highlighted line represents the terminal that is currently visible in the UI. The bug button to the right starts a debug session for the associated terminal. The bug icon to the left indicates that a debug session is already active for that terminal.

### DIFF
--- a/rascal-vscode-extension/package.json
+++ b/rascal-vscode-extension/package.json
@@ -55,26 +55,13 @@
           "light": "./assets/images/rascal-logo-v2.1.svg",
           "dark": "./assets/images/rascal-logo-v2.1.svg"
         }
-      },
-      {
-        "command": "rascalmpl.stopDebuggerForRepl",
-        "title": "Stop Rascal debugger for REPL",
-        "icon": {
-          "light": "./assets/images/rascal-logo-v2.1.svg",
-          "dark": "./assets/images/rascal-logo-v2.1.svg"
-        }
       }
     ],
     "menus": {
       "view/item/context": [
         {
           "command": "rascalmpl.startDebuggerForRepl",
-          "when": "view == rascalmpl-debugger-view && viewItem == 'isNotDebugging'",
-          "group": "inline"
-        },
-        {
-          "command": "rascalmpl.stopDebuggerForRepl",
-          "when": "view == rascalmpl-debugger-view && viewItem == 'isDebugging'",
+          "when": "view == rascalmpl-debugger-view && viewItem == 'canStartDebugging'",
           "group": "inline"
         }
       ]
--- a/rascal-vscode-extension/package.json
+++ b/rascal-vscode-extension/package.json
@@ -47,8 +47,28 @@
       {
         "command": "rascalmpl.importModule",
         "title": "Start Rascal Terminal and Import this module"
+      },
+      {
+        "command": "rascalmpl.startDebuggerForRepl",
+        "title": "Start Rascal debugger for REPL",
+        "icon": "$(debug)"
       }
     ],
+    "menus": {
+      "commandPalette": [
+        {
+          "command": "rascalmpl.startDebuggerForRepl",
+          "when": "false"
+        }
+      ],
+      "view/item/context": [
+        {
+          "command": "rascalmpl.startDebuggerForRepl",
+          "when": "view == rascalmpl-debugger-view && viewItem == 'canStartDebugging'",
+          "group": "inline"
+        }
+      ]
+    },
     "languages": [
       {
         "id": "rascalmpl",
--- a/rascal-vscode-extension/package.json
+++ b/rascal-vscode-extension/package.json
@@ -55,6 +55,12 @@
       }
     ],
     "menus": {
+      "commandPalette": [
+        {
+          "command": "rascalmpl.startDebuggerForRepl",
+          "when": "false"
+        }
+      ],
       "view/item/context": [
         {
           "command": "rascalmpl.startDebuggerForRepl",
--- a/rascal-vscode-extension/package.json
+++ b/rascal-vscode-extension/package.json
@@ -47,8 +47,38 @@
       {
         "command": "rascalmpl.importModule",
         "title": "Start Rascal Terminal and Import this module"
+      },
+      {
+        "command": "rascalmpl.startDebuggerForRepl",
+        "title": "Start Rascal debugger for REPL",
+        "icon": {
+          "light": "./assets/images/rascal-logo-v2.1.svg",
+          "dark": "./assets/images/rascal-logo-v2.1.svg"
+        }
+      },
+      {
+        "command": "rascalmpl.stopDebuggerForRepl",
+        "title": "Stop Rascal debugger for REPL",
+        "icon": {
+          "light": "./assets/images/rascal-logo-v2.1.svg",
+          "dark": "./assets/images/rascal-logo-v2.1.svg"
+        }
       }
     ],
+    "menus": {
+      "view/item/context": [
+        {
+          "command": "rascalmpl.startDebuggerForRepl",
+          "when": "view == rascalmpl-debugger-view && viewItem == 'isNotDebugging'",
+          "group": "inline"
+        },
+        {
+          "command": "rascalmpl.stopDebuggerForRepl",
+          "when": "view == rascalmpl-debugger-view && viewItem == 'isDebugging'",
+          "group": "inline"
+        }
+      ]
+    },
     "languages": [
       {
         "id": "rascalmpl",
@@ -89,12 +119,24 @@
           "icon": "./assets/images/rascal-logo-v2.1.svg",
           "visibility": "collapsed"
         }
+      ],
+      "debug": [
+        {
+          "id": "rascalmpl-debugger-view",
+          "name": "Rascal debugger",
+          "icon": "./assets/images/rascal-logo-v2.1.svg",
+          "visibility": "visible"
+        }
       ]
     },
     "viewsWelcome": [
       {
         "view": "rascalmpl-configuration-view",
         "contents": "No Rascal Projects found in the workspace"
+      },
+      {
+        "view": "rascalmpl-debugger-view",
+        "contents": "No active Rascal REPLs found"
       }
     ],
     "breakpoints": [
--- a/rascal-vscode-extension/package.json
+++ b/rascal-vscode-extension/package.json
@@ -47,28 +47,8 @@
       {
         "command": "rascalmpl.importModule",
         "title": "Start Rascal Terminal and Import this module"
-      },
-      {
-        "command": "rascalmpl.startDebuggerForRepl",
-        "title": "Start Rascal debugger for REPL",
-        "icon": "$(debug)"
       }
     ],
-    "menus": {
-      "commandPalette": [
-        {
-          "command": "rascalmpl.startDebuggerForRepl",
-          "when": "false"
-        }
-      ],
-      "view/item/context": [
-        {
-          "command": "rascalmpl.startDebuggerForRepl",
-          "when": "view == rascalmpl-debugger-view && viewItem == 'canStartDebugging'",
-          "group": "inline"
-        }
-      ]
-    },
     "languages": [
       {
         "id": "rascalmpl",
--- a/rascal-vscode-extension/package.json
+++ b/rascal-vscode-extension/package.json
@@ -108,7 +108,7 @@
         {
           "id": "rascalmpl-debugger-view",
           "name": "Rascal debugger",
-          "icon": "./assets/images/rascal-logo-v2.1.svg",
+          "icon": "$(debug)",
           "visibility": "visible"
         }
       ]
--- a/rascal-vscode-extension/package.json
+++ b/rascal-vscode-extension/package.json
@@ -51,10 +51,7 @@
       {
         "command": "rascalmpl.startDebuggerForRepl",
         "title": "Start Rascal debugger for REPL",
-        "icon": {
-          "light": "./assets/images/rascal-logo-v2.1.svg",
-          "dark": "./assets/images/rascal-logo-v2.1.svg"
-        }
+        "icon": "$(debug)"
       }
     ],
     "menus": {
--- a/rascal-vscode-extension/package.json
+++ b/rascal-vscode-extension/package.json
@@ -107,7 +107,7 @@
       "debug": [
         {
           "id": "rascalmpl-debugger-view",
-          "name": "Rascal debugger",
+          "name": "Debug Rascal Terminal",
           "icon": "$(debug)",
           "visibility": "visible"
         }
--- a/rascal-vscode-extension/src/RascalExtension.ts
+++ b/rascal-vscode-extension/src/RascalExtension.ts
@@ -57,7 +57,7 @@ export class RascalExtension implements vscode.Disposable {
         checkForJVMUpdate();
 
         vscode.window.registerTreeDataProvider('rascalmpl-configuration-view', new RascalLibraryProvider(this.rascal.rascalClient));
-        vscode.window.registerTreeDataProvider('rascalmpl-debugger-view', new RascalDebugViewProvider(this.rascal.rascalDebugClient));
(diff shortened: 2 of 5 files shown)


rascal-vscode-extension/package.json
```diff
@@ -47,8 +47,22 @@
       {
         "command": "rascalmpl.importModule",
         "title": "Start Rascal Terminal and Import this module"
+      },
+      {
+        "command": "rascalmpl.startDebuggerForRepl",
```
DavyLandman, line 52: Reading this correctly: https://code.visualstudio.com/api/extension-guides/command

It looks like we should only register commands here that are user facing. Like, we don't want this one showing up in the Command palette that a user can execute. Have we tried what happens if we only register the command in the typescript code?

rascal-vscode-extension/package.json
```diff
@@ -47,8 +47,22 @@
...
+        "title": "Start Rascal debugger for REPL",
+        "icon": "$(debug)"
       }
     ],
+    "menus": {
+      "view/item/context": [
+        {
+          "command": "rascalmpl.startDebuggerForRepl",
```
DavyLandman, line 60: similar here, if we move the command declaration to an annotation on the list item node, we can get away with the when class I think?

rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -52,9 +53,11 @@ export class RascalExtension implements vscode.Disposable {
         this.registerTerminalCommand();
         this.registerMainRun();
         this.registerImportModule();
+        this.registerDebuggerCommand();
```
DavyLandman, line 56: I think the `RascalDebugViewProvider` should be responsible for registring this command, as without it, it doesn't function.

rascal-vscode-extension/src/dap/RascalDebugClient.ts
```diff
@@ -77,13 +77,11 @@ export class RascalDebugClient {
 
     registerDebugServerPort(processID: number, serverPort: number){
         this.debugSocketServersPorts.set(processID, serverPort);
+        commands.executeCommand('rascalmpl.updateDebugView');
```
DavyLandman, line 80: Hmm this is a bit strange, I think we should use some kind of event system here, where the debug view just subscribes to the RascalDebugClient for updates. vscode library already has a buildin set of classes for this.

rascal-vscode-extension/src/dap/RascalDebugView.ts
```diff
@@ -0,0 +1,118 @@
...
+        });
+        vscode.debug.onDidTerminateDebugSession(_e => {
+            this.changeEmitter.fire(undefined);
+        });
+        vscode.debug.onDidReceiveDebugSessionCustomEvent(_e => {
+            this.changeEmitter.fire(undefined);
+        });
+        context.subscriptions.push(vscode.commands.registerCommand('rascalmpl.updateDebugView', () => {
+            this.changeEmitter.fire(undefined);
+        }));
```
DavyLandman, line 55: Can we make a small function that takes a any generic parameter, and just triggers an `changeEmitter.fire` event, and provide that as event handler for all these functions?

Also, don't all of these `onDid`... handlers need to be pushed to the subscribitions? to make sure we properly release resources on restart/update?

rascal-vscode-extension/src/dap/RascalDebugView.ts
```diff
@@ -0,0 +1,118 @@
...
+        return element;
+    }
+    getChildren(element?: RascalReplNode | undefined): vscode.ProviderResult<RascalReplNode[]> {
+        if (element === undefined) {
+            return this.updateRascalDebugView();
+        }
+        return [];
+    }
+    getParent?(_element: RascalReplNode): vscode.ProviderResult<RascalReplNode> {
+        return undefined;
```
DavyLandman, line 68: nit: add a comment to explain that we only have root items?

DavyLandman (approved):
Nice and compact set of changes to get a great UX feature :+1:

rascal-vscode-extension/src/dap/RascalDebugView.ts
```diff
@@ -0,0 +1,120 @@
...
+        }
+        return processId === await activeTerminal.processId;
+    }
+
+}
+
+export class RascalReplNode extends vscode.TreeItem {
+    serverPort? : number | undefined;
+
+    constructor(label : string | vscode.TreeItemLabel, serverPort : number | undefined, isDebugging : boolean) {
```
DavyLandman, line 115: nit: if we write `public readonly serverPort?: number | undefined` in the constructor definition, it automatically becomes an field.

rascal-vscode-extension/src/dap/RascalDebugView.ts
DavyLandman: (also this `rascalmpl.updateDebugView` is an example of a command not registered in the `package.json` ;)

rascal-vscode-extension/src/RascalExtension.ts
rodinaarssen: Fair point. Updated.

rascal-vscode-extension/package.json
rodinaarssen: You get an error indicating that the TreeViewItem references a command that's not registered. It's possible to prevent the command from showing up in the command palette, though.

Registering the command in the TS code is possible, and the command is then called when the TreeViewItem is clicked. However, we lose the possibility to attach a button icon to the list entries.

I've changed the logic around the command for now.

DavyLandman: Yeah, my only goal is, I don't want users accidentally triggering that command when they find it in the command list.

rascal-vscode-extension/src/dap/RascalDebugClient.ts
rodinaarssen: Thanks for the suggestion

rascal-vscode-extension/package.json
rodinaarssen: As the icon is useful for the user, I've reinstated it.

rascal-vscode-extension/package.json
```diff
@@ -47,8 +47,28 @@
...
+        "command": "rascalmpl.startDebuggerForRepl",
+        "title": "Start Rascal debugger for REPL",
+        "icon": "$(debug)"
       }
     ],
+    "menus": {
+      "commandPalette": [
+        {
+          "command": "rascalmpl.startDebuggerForRepl",
+          "when": "false"
```
DavyLandman, line 61: nice 👍🏼

rascal-vscode-extension/src/dap/RascalDebugClient.ts
```diff
@@ -37,7 +37,8 @@ export class RascalDebugClient {
     debugSocketServersPorts: Map<number, number>; // Terminal processID -> socket server port for debug
     runningDebugSessionsPorts: Set<number>; // Stores all running debug session server ports
 
-
+    private portEventEmitter = new EventEmitter<undefined>();
```
DavyLandman, line 40: can we not give this a better type than `undefined`? something like` [number, number]` if we don't want to make a new class? or just inline anonymous class `{ processId: number; serverPort: number; }`?

rascal-vscode-extension/src/dap/RascalDebugView.ts
```diff
@@ -0,0 +1,116 @@
...
+        const fireEmitter = (_: vscode.Terminal | vscode.DebugSession | undefined) : void => {
+            this.changeEmitter.fire(undefined);
+        };
+        context.subscriptions.push(vscode.window.onDidOpenTerminal(fireEmitter));
+        context.subscriptions.push(vscode.window.onDidCloseTerminal(fireEmitter));
+        context.subscriptions.push(vscode.window.onDidChangeActiveTerminal(fireEmitter));
+        context.subscriptions.push(vscode.debug.onDidStartDebugSession(fireEmitter));
+        context.subscriptions.push(vscode.debug.onDidTerminateDebugSession(fireEmitter));
+
+        context.subscriptions.push(this.rascalDebugClient.portRegistrationEvent(fireEmitter));
```
DavyLandman, line 44: nit: these `on..` functions take a second argument, that allows us to pass in the `context.subscriptions` array.

for example:

```
vscode.window.onDidOpenTerminal(fireEmitter, context.subscriptions);
```

rascal-vscode-extension/src/dap/RascalDebugView.ts
```diff
@@ -0,0 +1,116 @@
...
+        //The Rascal debug view gives a flat list of the opened Rascal terminals. As such, only root items exit in this view.
+        return undefined;
+    }
+    resolveTreeItem?(item: vscode.TreeItem, _element: RascalReplNode, _token: vscode.CancellationToken): vscode.ProviderResult<vscode.TreeItem> {
+        return item;
+    }
+
+    async makeLabel(label: string, processId : number) : Promise<string | vscode.TreeItemLabel> {
+        if (await this.isActiveTerminal(processId)) {
+            return {label: label, highlights : [[0, label.length]]};
```
DavyLandman, line 74: nice way to highlight the active REPL 👍🏼

rascal-vscode-extension/src/dap/RascalDebugView.ts
```diff
@@ -0,0 +1,116 @@
...
+        }
+        return processId === await activeTerminal.processId;
+    }
+
+}
+
+export class RascalReplNode extends vscode.TreeItem {
+    constructor(label : string | vscode.TreeItemLabel, readonly serverPort : number | undefined, isDebugging : boolean) {
+        super(label, vscode.TreeItemCollapsibleState.None);
+        this.serverPort = serverPort;
```
DavyLandman, line 113: nit: is not needed, happens automatically if we define it in the header (although I'm not sure if `readonly` is enough to make it a field, maybe thats why the `this.` was still needed?)

rascal-vscode-extension/src/dap/RascalDebugView.ts
```diff
@@ -0,0 +1,116 @@
...
+                if (serverPort !== undefined && !isDebugging) {
+                    replNode.contextValue = "canStartDebugging";
+                }
+                result.push(replNode);
+            }
+        }
+        return result;
+    }
+
+    async isActiveTerminal(processId : number) : Promise<boolean> {
```
DavyLandman, line 100: since we're rebuiling this list quite frequently (instead of only precisly invalidating the entry we know has changed), I would like to propose we don't `await` on `activeTerminal.processId` for every entry in the loop of open terminals, but determine this once (inside `updateRascalDebugView` function), and pass a boolean into the `makeLabel` function.

DavyLandman (changes requested):
yes, looks very nice.

2 small nits, but 1 point where I think we should be a bit more mindfull of the fact we're running in the UI and shouldn't have a hot loop full of promises.

rascal-vscode-extension/src/dap/RascalDebugView.ts
rodinaarssen: I did not know that, thanks. In fact, it's the third argument.

rascal-vscode-extension/src/dap/RascalDebugClient.ts
rodinaarssen: The debug view currently does not keep state regarding ports and process IDs (or, in fact, copy state from the debug client). When receiving an event, the debug view currently iterates over the open terminals and retrieves data from the debug client. In the current implementation, there is no real need to communicate the `processId:serverPort` pairs. Do you propose to keep a shadow bookkeeping in the view implementation, or do you prefer events having payload in general?

rodinaarssen:
For a visual impression, see the attached screenshot.

image

Each line represents a single active Rascal terminal. The highlighted line represents the terminal that is currently visible in the UI. The bug button to the right starts a debug session for the associated terminal. The bug icon to the left indicates that a debug session is already active for that terminal.

rascal-vscode-extension/src/dap/RascalDebugClient.ts
DavyLandman: I think for now, the fact that we don't keep state in the debug view is fine. I get why that would make the code harder.

So yes, I just prefer events to have a payload :) So that if later we want to use it, it's already there.

rodinaarssen: Okay, fair enough. I agree.

DavyLandman (approved):

rascal-vscode-extension/src/dap/RascalDebugView.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+    getParent?(_element: RascalReplNode): vscode.ProviderResult<RascalReplNode> {
+        //The Rascal debug view gives a flat list of the opened Rascal terminals. As such, only root items exit in this view.
+        return undefined;
+    }
+    resolveTreeItem?(item: vscode.TreeItem, _element: RascalReplNode, _token: vscode.CancellationToken): vscode.ProviderResult<vscode.TreeItem> {
+        return item;
+    }
+
+    makeLabel(label: string, isActiveTerminal : boolean) : string | vscode.TreeItemLabel {
```
urbanfly, line 73: I recommend using a `FileDecoratorProvider` to highlight the active terminal instead of "abusing" the highlights concept.

Each of the `RascalReplNode` will need a unique Uri and the decorator will
- compare the Uri to the `vscode.window.activeTerminal`
- subscribe to the `onDidChangeActiveTerminal` event to refresh the decoration on the previous and newly active Uris

Decorations can give the active node a badge like "🎯" and give currently-debug-enabled nodes a badge like "🐞"

urbanfly:
I know this is late feedback, but there _may be_ a simpler way to achieve a similar level of discoverability for the debugging feature without introducing new UI.

Custom commands can be [added to the Terminal view](https://code.visualstudio.com/api/references/contribution-points#contributes.submenus:~:text=The%20example%20below%20contributes%20to%20an%20arbitrary%20view%20like%20the%20terminal%3A). [Custom state would need to be maintained](https://code.visualstudio.com/api/references/when-clause-contexts#add-a-custom-when-clause-context) when the activeTerminal changes.

```json
"view/title": [
  {
    "command": "rascalmpl.startDebuggerForRepl",
    "when": "view == terminal && rascalTerminalActive && !rascalTerminalDebugging"
    "group": "navigation@1"
  },
  {
    "command": "rascalmpl.stopDebuggerForRepl",
    "when": "view == terminal && rascalTerminalActive && rascalTerminalDebugging"
    "group": "navigation@1"
  }
]
```

rascal-vscode-extension/src/dap/RascalDebugView.ts
DavyLandman: That is a nice trick indeed :) but in a way, "abusing" a different concept. As a terminal is not a file.

DavyLandman:
> I know this is late feedback, but there _may be_ a simpler way to achieve a similar level of discoverability for the debugging feature without introducing new UI.
> 
> Custom commands can be [added to the Terminal view](https://code.visualstudio.com/api/references/contribution-points#contributes.submenus:~:text=The%20example%20below%20contributes%20to%20an%20arbitrary%20view%20like%20the%20terminal%3A). [Custom state would need to be maintained](https://code.visualstudio.com/api/references/when-clause-contexts#add-a-custom-when-clause-context) when the activeTerminal changes.
> 

Hmm, this is a nice addition, those small icons are a bit less discoverable, the current view fits better with how other programs that are running can be debugged. But I do think adding such an icon can be a parallel path to discover the debugger. I'll make it into a new issue to track this idea.

@urbanfly please let me know if you would like to turn that into a small PR?

urbanfly:
I'll consider it.

PR #574, opened 2025-02-13
Fix rename refactoring name check

toinehartman:
This PR fixes issues when renaming something in the Rascal project, reported by @PieterOlivier.

1. When looking for the use of a name in a module, trying all possible sub trees caused a stack overflow. Besides that, pretty-printing all of those sub trees was really expensive. This is now rewritten to a more sensible implementation using `visit`.
2. Rewrite the workspace crawl to use less environments.
3. Make the name check robust against modules with parse errors.

### DIFF
--- a/rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
+++ b/rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
@@ -554,19 +554,16 @@ ProjectFiles preloadFiles(set[loc] workspaceFolders, loc cursorLoc) {
 }
 
 ProjectFiles allWorkspaceFiles(set[loc] workspaceFolders, str cursorName, bool(loc, str) containsName, PathConfig(loc) getPathConfig) {
-    ProjectFiles fs = {};
-    for (folder <- workspaceFolders) {
-        pcfg = getPathConfig(folder);
-        for (srcFolder <- pcfg.srcs) {
-            for (loc f <- find(srcFolder, "rsc")) {
-                // If we do not find any occurrences of the name under the cursor in a module,
-                // we are not interested in loading the model, but we still want to inform the
-                // renaming framework about the existence of the file.
-                fs += <folder, containsName(f, cursorName), f>;
-            }
-        }
-    }
-    return fs;
+    return {
+        // If we do not find any occurrences of the name under the cursor in a module,
+        // we are not interested in loading the model, but we still want to inform the
+        // renaming framework about the existence of the file.
+        <folder, containsName(file, cursorName), file>
+        | folder <- workspaceFolders
+        , PathConfig pcfg := getPathConfig(folder)
+        , srcFolder <- pcfg.srcs
+        , file <- find(srcFolder, "rsc")
+    };
 }
 
 set[TModel] tmodelsForProjectFiles(ProjectFiles projectFiles, set[TModel](set[loc], PathConfig) tmodelsForFiles, PathConfig(loc) getPathConfig) =
--- a/rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
+++ b/rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
@@ -479,21 +479,38 @@ private Cursor rascalGetCursor(TModel ws, Tree cursorT) {
     return cursor(kind, min(locsContainingCursor.l), cursorName);
 }
 
-@memo{maximumSize(1000), expireAfter(minutes=5)}
-private set[str] rascalNameToEquivalentNames(str name) =
-    {name, startsWith(name, "\\") ? name : "\\<name>"};
+private bool rascalContainsName(loc l, str n) {
+    en = rascalEscapeName(n);
 
-private bool rascalContainsName(loc l, str name) {
-    names = rascalNameToEquivalentNames(name);
-    bool matches(Tree t) = "<t>" in names;
+    // Since QualifiedName is the most liberal and all the others are subsets of it,
+    // we default to QualifiedName in case parsing as something else fails.
+    qName = [QualifiedName] n;
+    qNameEsc = [QualifiedName] en;
 
+    Tree tryNameParse(type[&T <: Tree] a, str s) {
+        try {
+            return parse(a, s);
+        } catch _: {
+            return qName;
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
```diff
@@ -479,12 +479,26 @@ private Cursor rascalGetCursor(TModel ws, Tree cursorT) {
...
+    names = rascalNameToEquivalentNames(name);
+    bool matches(Tree t) = "<t>" in names;
+
+    try {
+        m = parseModuleWithSpacesCached(l);
+        visit (m) {
+            case Name n: if (matches(n)) return true;
+            case QualifiedName n: if (matches(n)) return true;
+            case Nonterminal n: if (matches(n)) return true;
+            case NonterminalLabel n: if (matches(n)) return true;
```
DavyLandman, line 496: yes this is already a bit better than the original one, but we still do a lot of unparsing. we can also use pattern matching on trees to take care of it.

Here is a small example of how we can use pattern matching on a few trees to 

```rascal
import ParseTree;

lexical A = [a-z]+ !>>[a-z];
lexical B = "::" [a-z]+ !>> [a-z];

lexical AoB = A | B;

bool match_single_tree_iteration(str name, AoB input) {
    // we build subtrees that are valid candidates
    matches = {[A]name, [B]"::<name>"};
    // then we use <- to pattern match the tree against the set
    return any(/Tree t := input, t <- matches);
}

// this is roughly the same, except that the pattern matcher takes care to go through all subtrees and figure out which might match.
bool match_using_visitor(str name, AoB input) {
    aName = [A]name;
    bName = [B]"::<name>";
    visit (input) {
        case aName: return true;
        case bName: return true;
    }
    return false;
}
```

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
```diff
@@ -523,16 +537,19 @@ ProjectFiles preloadFiles(set[loc] workspaceFolders, loc cursorLoc) {
...
-        | folder <- workspaceFolders
-        , PathConfig pcfg := getPathConfig(folder)
-        , srcFolder <- pcfg.srcs
-        , file <- find(srcFolder, "rsc")
-    };
+    ProjectFiles fs = {};
+    for (folder <- workspaceFolders) {
+        pcfg = getPathConfig(folder);
+        for (srcFolder <- pcfg.srcs) {
+            for (loc f <- find(srcFolder, "rsc")) {
```
DavyLandman, line 544: nit: we can inline loops:

```rascal
 for (srcFolder <- pcfg.srcs, loc f <- find(srcFolder, "rsc")) {
```` 

or even:

```rascal
for (folder <- workspaceFolders, srcFolder <- getPathConfig(folder).srcs, loc f <- find(srcFolder, "rsc")) {
```

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
```diff
@@ -523,16 +537,19 @@ ProjectFiles preloadFiles(set[loc] workspaceFolders, loc cursorLoc) {
...
-        // If we do not find any occurrences of the name under the cursor in a module,
-        // we are not interested in loading the model, but we still want to inform the
-        // renaming framework about the existence of the file.
-        <folder, containsName(file, cursorName), file>
-        | folder <- workspaceFolders
-        , PathConfig pcfg := getPathConfig(folder)
-        , srcFolder <- pcfg.srcs
-        , file <- find(srcFolder, "rsc")
-    };
+    ProjectFiles fs = {};
```
DavyLandman, line 540: this change seems to be due to:

> Rewrite the workspace crawl to use less environments.

How come this is a concern? how much slower is it to write is as a comprehension?

DavyLandman (commented):
I like the idea to fix the scan for the `rascalContainsName` function, I'm less convinced the crawl of the directories is need? do you have some profiles/benchmarks to share?

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
toinehartman: This is more of a general attempt to remove the use of backtracking generators, and turned out to be in the path of debugging this issue. I can not imagine this showing up in any real-use profile. I can revert if you prefer the comprehension. Let me know.

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
toinehartman: It was a bit more complicated than this since we cannot parse all names with all rules, but with some defaulting behaviour, this should be a nice performance improvement :+1:

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
DavyLandman: well, as a reviewer, I look at a change and think: why is this a change that we need. So let's revert is, and keep the comprehension. 

(a variant of: no need for premature optimization)

toinehartman: Ok, reverted!

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
toinehartman: outdated since these changes were reverted

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
```diff
@@ -479,12 +479,43 @@ private Cursor rascalGetCursor(TModel ws, Tree cursorT) {
     return cursor(kind, min(locsContainingCursor.l), cursorName);
 }
 
-private set[str] rascalNameToEquivalentNames(str name) =
-    {name, startsWith(name, "\\") ? name : "\\<name>"};
+private bool rascalContainsName(loc l, str n) {
```
DavyLandman, line 482: If __after profiling__ this function still takes up a lot of time, we could change it so that it returns a function that given a loc returns a boolean. And we only build the partial parse trees for the names once.

PR #575, opened 2025-02-13
Run type checker on dependencies in IDE

rodinaarssen:
When a file is saved in VS Code, the type checker now makes sure that all (transitive) dependencies are type checked first.


rascal-lsp/src/main/rascal/lang/rascal/lsp/IDECheckerWrapper.rsc
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    while (exists(current), isDirectory(current)) {
+        if (exists(current + "META-INF" + "RASCAL.MF")) {
+            return ensureTrailingSpaceInPath(current);
+        }
+        if (!current.parent?) {
+            return isDirectory(member) ? member : member.parent;
+        }
+        current = current.parent;
```
DavyLandman, line 89: what if `current.parent == current`?

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
```diff
@@ -125,6 +125,11 @@ describe('IDE', function () {
         await triggerTypeChecker(editor, TestWorkspace.mainFileTpl, true);
     });
 
+    it("type checker runs on dependencies", async() => {
+        const editor = await ide.openModule(TestWorkspace.libCallFile);
```
DavyLandman, line 129: Doesn't this test depend on the state of previous checks? shouldn't we make sure to delete some tpl files first?

DavyLandman (changes requested):
Looking good, some points where we need to be a bit more defensive.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -158,7 +163,34 @@ private static Map<ISourceLocation, ISet> translateCheckResults(IList messages)
 
     public InterruptibleFuture<Map<ISourceLocation, ISet>> compileFile(ISourceLocation file, PathConfig pcfg,
         Executor exec) {
-        return compileFileList(VF.list(file), pcfg, exec);
+        logger.debug("Running Rascal check for: {} with {}", file, pcfg);
+        return runEvaluator("Rascal check", compilerEvaluator,
+            e -> {
+                IFunction getPathConfig = e.getFunctionValueFactory().function(getPathConfigType, (t, u) -> addResources(rascalTextDocumentService.getFileFacts().getPathConfig((ISourceLocation) t[0])));
```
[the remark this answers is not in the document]
rodinaarssen: Good idea.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -158,7 +163,34 @@ private static Map<ISourceLocation, ISet> translateCheckResults(IList messages)
...
+                IFunction getParseTree = e.getFunctionValueFactory().function(getParseTreeType, (t, u) -> {
```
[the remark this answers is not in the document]
rodinaarssen: Good idea.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigs.java
```diff
@@ -171,7 +171,7 @@ private static boolean hasParentSection(URIResolverRegistry reg, ISourceLocation
 
     private static PathConfig actualBuild(ISourceLocation projectRoot) {
         try {
-            return PathConfig.fromSourceProjectRascalManifest(projectRoot, RascalConfigMode.COMPILER);
+            return PathConfig.fromSourceProjectRascalManifest(projectRoot, RascalConfigMode.INTERPETER);
```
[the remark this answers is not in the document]
rodinaarssen: This has since been reverted.

rascal-lsp/src/main/rascal/lang/rascal/lsp/IDECheckerWrapper.rsc
```diff
@@ -0,0 +1,96 @@
...
+import lang::rascalcore::check::ModuleLocations;
+
+@synopsis{
+    This function is a wrapper for the Rascal type checker, taking into account that dependencies
+    may be open in the current workspace. It takes an overapproximation of the import/extend graph,
+    and calls the type checker on (transitive) dependencies first.
+
+    This function must only be used in an IDE context.
+}
+list[ModuleMessages] checkFile(loc l, node(loc file) getParseTree, PathConfig(loc file) getPathConfig) {
```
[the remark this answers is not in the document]
rodinaarssen: Yes. Changed to `start[Module]` on both ends.

rascal-lsp/src/main/rascal/lang/rascal/lsp/IDECheckerWrapper.rsc
rodinaarssen: Doesn't Rascal always throw a `NoParent` instead of returning the location itself?

rascal-lsp/src/main/rascal/lang/rascal/lsp/IDECheckerWrapper.rsc
```diff
@@ -0,0 +1,96 @@
...
+            dependencies += <currentProject, resolveLocation(mlProject)>;
+            if (ml.extension == "rsc" && ml notin checkedForImports, start[Module] mlpt := getParseTree(ml)) {
+                checkForImports += mlpt;
+            }
+        }
+        checkedForImports += currentSrc;
+        checkForImports -= tree;
+    }
+    modulesPerProject = classify(checkedForImports, loc(loc l) {return inferProjectRoot(l);});
+    return [*check([*modulesPerProject[project]], makeCompilerConfig(getPathConfig(project))) | project <- reverse(order(dependencies)), project in modulesPerProject];
```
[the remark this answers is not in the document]
rodinaarssen: Added some monitor messages.

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
rodinaarssen: The line that follows this one specifies the `tpl` file of the lib file (i.e., the file being called). This `tpl` file is deleted by the test framework before starting the test.

DavyLandman: Ok, nice

DavyLandman (changes requested):
Since this code is hit quite frequent, I thought to review it again, I found a bug in an unreleased resource.

rascal-lsp/src/main/rascal/lang/rascal/lsp/IDECheckerWrapper.rsc
rodinaarssen: The condition in the `if` 'catches' the `NoParent`s

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -151,14 +160,51 @@ private static Map<ISourceLocation, ISet> translateCheckResults(IList messages)
...
+                var tree = rascalTextDocumentService.getFile(resolvedLocation).getMostRecentTree();
+                if (tree != null) {
+                    return tree.get();
+                }
+            } catch (ResponseErrorException e1) {
+                // File is not open in the IDE
+            }
+            // Parse the source file
+            try {
+                var contents = Prelude.consumeInputStream(URIResolverRegistry.getInstance().getCharacterReader(resolvedLocation));
```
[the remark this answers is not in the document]
rodinaarssen: Good catch

rascal-lsp/src/main/rascal/lang/rascal/lsp/IDECheckerWrapper.rsc
```diff
@@ -0,0 +1,133 @@
...
+});
+
+RascalCompilerConfig makeCompilerConfig(PathConfig pcfg) = rascalCompilerConfig(pcfg[resources=pcfg.bin]);
+
+loc locateRascalModule(str fqn, PathConfig pcfg, PathConfig(loc file) getPathConfig) {
+    fileName = makeFileName(fqn);
+    for (dir <- pcfg.srcs, fileLoc := dir + fileName, exists(fileLoc)) {
+        return fileLoc;
+    }
+    if (lib <- pcfg.libs, lib != |lib://rascal/|, dir <- getPathConfig(inferProjectRoot(lib)).srcs, fileLoc := dir + fileName, exists(fileLoc)) {
```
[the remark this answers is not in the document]
rodinaarssen: I figured it'd never be useful to have the checker dive into `lib://rascal/` (and perhaps the same would hold for `lib://typepal/` in that case), but on second thought, what if these projects were open in the IDE? Perhaps this exclusion should just be removed.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -151,14 +160,51 @@ private static Map<ISourceLocation, ISet> translateCheckResults(IList messages)
...
+        return e.getFunctionValueFactory().function(getPathConfigType, (t, u) -> {
+            return addResources(rascalTextDocumentService.getFileFacts().getPathConfig((ISourceLocation) t[0]));
+        });
+    }
+
+    IFunction makeParseTreeGetter(Evaluator e) {
+        return e.getFunctionValueFactory().function(getParseTreeType, (t, u) -> {
+            ISourceLocation resolvedLocation;
+            try {
+                resolvedLocation = URIResolverRegistry.getInstance().logicalToPhysical((ISourceLocation) t[0]);
```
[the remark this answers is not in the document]
rodinaarssen: `RascalTextDocumentService` uses resolved locations in the `documents` map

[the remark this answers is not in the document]
rodinaarssen: At this point I'm unsure about the interplay with `lsp+`. I expect that the key in the `documents` map will still be the resolved location, but I'd have to check.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -97,10 +101,14 @@ public class RascalLanguageServices {
     private final TypeFactory tf = TypeFactory.getInstance();
     private final TypeStore store = new TypeStore();
     private final Type getPathConfigType = tf.functionType(tf.abstractDataType(store, "PathConfig"), tf.tupleType(tf.sourceLocationType()), tf.tupleEmpty());
+    private final IConstructor startModuleConstructor = VF.constructor(RascalValueFactory.Symbol_Start, VF.constructor(RascalValueFactory.Symbol_Sort, VF.string("Module")));
```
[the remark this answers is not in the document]
rodinaarssen: I find it cleaner to have them at the top of a file, especially here, where `getPathConfigType` is used more frequently.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -151,14 +160,51 @@ private static Map<ISourceLocation, ISet> translateCheckResults(IList messages)
         return messages.stream()
             .filter(IConstructor.class::isInstance)
             .map(IConstructor.class::cast)
+            .distinct()
```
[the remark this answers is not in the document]
rodinaarssen: I agree that it's cleaner to do this on the Rascal side.

rascal-lsp/src/main/rascal/lang/rascal/lsp/IDECheckerWrapper.rsc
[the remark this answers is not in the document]
rodinaarssen: I agree

rascal-lsp/src/main/rascal/lang/rascal/lsp/IDECheckerWrapper.rsc
```diff
@@ -0,0 +1,133 @@
...
+    = job("Rascal check", list[ModuleMessages]() {
+    checkForImports = [getParseTree(l)];
+    checkedForImports = {};
+    initialProject = inferProjectRoot(l);
+
+    rel[loc, loc] dependencies = {};
+    
+    msgs = [];
+
+    jobStep("Rascal check", "Building dependency graph");
```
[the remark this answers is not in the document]
rodinaarssen: It's unfortunate that the other overload requires a specification of the amount of work, but I agree that it's better not to be repeating "Rascal check" every time.

rascal-lsp/src/main/rascal/lang/rascal/lsp/IDECheckerWrapper.rsc
```diff
@@ -0,0 +1,134 @@
...
+list[ModuleMessages] checkFile(loc l, start[Module](loc file) getParseTree, PathConfig(loc file) getPathConfig) 
+    = job("Rascal check", list[ModuleMessages](void(str, int) step) {
+    checkForImports = [getParseTree(l)];
+    checkedForImports = {};
+    initialProject = inferProjectRoot(l);
+
+    rel[loc, loc] dependencies = {};
+    
+
+    step("Building dependency graph", 50);
```
DavyLandman, line 61: I know it's possible to grow the total size when we know about more modules. instead of guessing 50.

rascal-lsp/src/main/rascal/lang/rascal/lsp/IDECheckerWrapper.rsc
```diff
@@ -0,0 +1,134 @@
...
+                }
+            } catch _: {
+                ;// Continue
+            }
+        }
+        checkedForImports += currentSrc;
+        checkForImports -= tree;
+    }
+    if (<p,p> <- (dependencies - ident(carrier(dependencies)))+) {
+        return [program(l, error("Detected cyclic dependencies between projects. Fix your project setup.", l))];
```
DavyLandman, line 80: improve the message? `"We cannot support the cyclic dependency between project <A>  and <B>"`

rascal-lsp/src/main/rascal/lang/rascal/lsp/IDECheckerWrapper.rsc
DavyLandman: aaaah, right, my bad.

rascal-lsp/src/main/rascal/lang/rascal/lsp/IDECheckerWrapper.rsc
```diff
@@ -0,0 +1,134 @@
...
+});
+
+loc locateRascalModule(str fqn, PathConfig pcfg, PathConfig(loc file) getPathConfig) {
+    fileName = makeFileName(fqn);
+    // Check the source directories
+    for (dir <- pcfg.srcs, fileLoc := dir + fileName, exists(fileLoc)) {
+        return fileLoc;
+    }
+    // Check the source directories of libraries that are currently open in VS Code
+    if (lib <- pcfg.libs, lib.scheme != "lib", dir <- getPathConfig(inferProjectRoot(lib)).srcs, fileLoc := dir + fileName, exists(fileLoc)) {
```
DavyLandman, line 100: in the rename function we explicitly have a list of current project in the workspace that is passed in. should we also do that? so that we don't try to read from non workspace locations? (especially as the `lib` scheme is soon dissapearing and getting replaced by for example `jar+file` and `mvn`)

rascal-lsp/src/main/rascal/lang/rascal/lsp/IDECheckerWrapper.rsc
```diff
@@ -0,0 +1,134 @@
...
+    }
+    modulesPerProject = classify(checkedForImports, loc(loc l) {return inferProjectRoot(l);});
+    msgs = [];
+    for (project <- reverse(order(dependencies)), project in modulesPerProject, project != initialProject) {
+        step("Checking project `<project.file>`", 1);
+        msgs += check([*modulesPerProject[project]], rascalCompilerConfig(getPathConfig(project))); 
+    }
+    step("Checking <l>", 1);
+    msgs += check([l], rascalCompilerConfig(getPathConfig(initialProject)));
+    return dup(msgs);
```
DavyLandman, line 90: almost makes you want to turn it into a set?

rascal-lsp/src/main/rascal/lang/rascal/lsp/IDECheckerWrapper.rsc
```diff
@@ -0,0 +1,135 @@
...
+        if (!current.parent?) {
+            return isDirectory(member) ? member : member.parent;
+        }
+        current = current.parent;
+    }
+    
+    return ensureTrailingSpaceInPath(current);
+}
+
+loc ensureTrailingSpaceInPath(loc l) = endsWith(l.path, "/") ? l : l[path=l.path + "/"];
```
urbanfly, line 134: Maybe this should be `ensureTrailingSlashInPath`?

rascal-lsp/src/main/rascal/lang/rascal/lsp/IDECheckerWrapper.rsc
rodinaarssen: That sounds like a good idea

rascal-lsp/src/main/rascal/lang/rascal/lsp/IDECheckerWrapper.rsc
DavyLandman: For future ref: the answer is `jobTodo` .

PR #581, opened 2025-02-21
Added a button to error pop-ups to copy the stack trace to the clipboard

rodinaarssen:
The error pop-up in VS Code now has an additional button to copy the stack trace to the clipboard.

image

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
@@ -190,7 +190,7 @@ public class EvaluatorUtil {
                         var content = new StringSelection(stackTrace);
                         Toolkit.getDefaultToolkit().getSystemClipboard().setContents(content, content);
                     }
-                } 
+                }
             });
     }
 
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
@@ -26,7 +26,9 @@
  */
 package org.rascalmpl.vscode.lsp.util;
 
+import java.awt.AWTError;
 import java.awt.Desktop;
+import java.awt.HeadlessException;
 import java.awt.Toolkit;
 import java.awt.datatransfer.StringSelection;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
```diff
@@ -163,23 +165,31 @@ private static void reportInternalError(Throwable e, String task, LanguageClient
         var msg = new ShowMessageRequestParams();
         msg.setMessage(title);
         msg.setType(MessageType.Error);
-        msg.setActions(Arrays.asList(new MessageActionItem("Report on GitHub"), new MessageActionItem("Ignore")));
+        final String ReportLabel = "Report on GitHub";
```
DavyLandman, line 168: It all looks good to me, but maybe this whole function can benefit from some extracted functions.

1. calculate the reason
2. calculate the stack trace
3. open the GH issue
4. copy to clipboard.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
```diff
@@ -163,23 +165,31 @@ private static void reportInternalError(Throwable e, String task, LanguageClient
...
+                        bodyWriter.println(e.getMessage());
+                        bodyWriter.println("```");
+                        bodyWriter.println("Stacktrace:");
+                        bodyWriter.println("```");
+                        bodyWriter.println(stackTrace);
+                        bodyWriter.println("```");
+                        browse("https://github.com/usethesource/rascal-language-servers/issues/new?labels=bug&title=" + URLEncoder.encode(title, StandardCharsets.UTF_8) + "&body=" + URLEncoder.encode(body.toString(), StandardCharsets.UTF_8), client);
+                    } else if (responseTitle.equals(CopyLabel)) {
+                        var content = new StringSelection(stackTrace);
+                        Toolkit.getDefaultToolkit().getSystemClipboard().setContents(content, content);
```
DavyLandman, line 191: have you checked if we should print a small notification? like "copied to clipboard"? to give the user feedback, or is that too noisy?

DavyLandman (approved):
LGTM, just 2 small questions, up to you what to do with them

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
rodinaarssen: It could indeed.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
rodinaarssen: There already is user feedback in the sense that the tooltip closes when a button is clicked. We could generate a notification, but I feel it's not necessary.

PR #547, opened 2024-12-17
Rewrite rename framework

toinehartman:
This reimplements the rename refactoring using the language-parametric renaming framework https://github.com/SWAT-engineering/rascal-rename-framework/. It aims to increase readability, maintainability and performance and almost completely rewrites the existing implementation.

Work included in this PR (separately reviewed):
- #584
- #593
- #594
- #595
- #596 
- #601 
- #605 
- #606 
- #609 
- #612

PR #583, opened 2025-02-24
Rename causes overload at downstream uses of new name

### LINKED ISSUE #582: Rename causes overloads of downstream use of new name
**Describe the bug**
Reported by @DavyLandman 

```rascal
module A

void aa(str s) { }
```

```rascal
module B


import A;
import IO;

void main() { println("aaa"); }
```

Rename `aa` to `println`. This will cause unintended overloads at call sites.

**Expected behavior**
An error that this renaming might change semantics.

toinehartman:
Closes #582.


DavyLandman:
I'm also fine only fixing this bug in the new refactored rename code, if this is too much work to fix in the old codebase

toinehartman:
That might be wiser, indeed. At least I wanted to document these tests and confirm if this was covered by the old implementation. I'll cherry-pick them onto the new implementation.

toinehartman:
Abandoning since we are completely rewriting the rename in #547

PR #593, opened 2025-03-06
Group document edits by file in tests

toinehartman:
The Rascal implementation `executeDocumentEdits`, that the renaming (only) relies on for testing, requires text edits to the same file to be batched under a single document edit. We could impose this invariant on the rename framework, but this might be difficult to guarantee in the presence of create/move/delete edits. Instead, this PR only applies this to the tests of the Rascal rename implementation.


DavyLandman:
I expect this should target a different branch than main?

PR #596, opened 2025-03-14
Rename: fields

toinehartman:
Renaming of positional/keyword fields of constructors, ADTs and collections.

This PR also introduces TModel augmentation. This works around missing definitions and missing/unusable use/def relations in the TModel, until https://github.com/usethesource/rascal/issues/2172 is fixed.

### DIFF
--- a/rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
+++ b/rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
@@ -551,6 +551,10 @@ Tree findCursorInTree(Tree t, loc cursorLoc) {
 list[Tree] extendFocusWithConcreteSyntax([Concrete c, *tail], loc cursorLoc) = [findCursorInTree(c, cursorLoc), c, *tail];
 default list[Tree] extendFocusWithConcreteSyntax(list[Tree] cursor, loc _) = cursor;
 
+@synopsis{
+    Augment the TModel with 'missing' use/def information.
+    Workaround until the typechecker generates this. https://github.com/usethesource/rascal/issues/2172
+}
 TModel augmentTModel(Tree tr, TModel tm, TModel(loc) tmodelForLoc) {
     tm = augmentFieldUses(tr, tm, tmodelForLoc);
     tm = augmentFormalUses(tr, tm);
--- a/rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
+++ b/rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
@@ -562,7 +562,7 @@ TModel augmentTModel(Tree tr, TModel tm, TModel(loc) tmodelForLoc) {
 }
 
 TModel tmodelForLoc(loc l, PathConfig(loc) getPathConfig)
-    = tmodelForTree(parse(#start[Module], l), getPathConfig);
+    = tmodelForTree(parseModuleWithSpaces(l), getPathConfig);
 
 TModel tmodelForTree(Tree t, PathConfig(loc) getPathConfig) {
     loc l = t.src.top;
--- a/rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
+++ b/rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
@@ -599,6 +599,8 @@ set[Define] getCursorDefinitions(list[Tree] cursor, Tree(loc) getTree, TModel(Tr
 
     loc cursorLoc = cursor[0].src;
     TModel tm = getModel(cursor[-1]);
+    if (isUnsupportedCursor(cursor, tm, r)) return {};
+
     for (Tree c <- cursor) {
         if (tm.definitions[c.src]?) {
             return {tm.definitions[c.src]};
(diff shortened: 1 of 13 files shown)


rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
```diff
@@ -551,23 +551,42 @@ Tree findCursorInTree(Tree t, loc cursorLoc) {
...
+    Workaround until the typechecker generates this. https://github.com/usethesource/rascal/issues/2172
+}
+TModel augmentTModel(Tree tr, TModel tm, TModel(loc) tmodelForLoc) {
+    tm = augmentFieldUses(tr, tm, tmodelForLoc);
+    tm = augmentFormalUses(tr, tm);
+    return tm;
+}
+
+TModel tmodelForLoc(loc l, PathConfig(loc) getPathConfig)
+    = tmodelForTree(parse(#start[Module], l), getPathConfig);
```
DavyLandman, line 565: It's better to use the function in `util::Reflective` as that makes sure to not generate a new parser). Also, shouldn't this be a feature of the rename framework? that just calls the parser & chains it to to tmodel function?

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
```diff
@@ -551,23 +551,42 @@ Tree findCursorInTree(Tree t, loc cursorLoc) {
...
+
+    ccfg = rascalCompilerConfig(pcfg);
+    ms = rascalTModelForNames([mname], ccfg, dummy_compile1);
+
+    <found, tm, ms> = getTModelForModule(mname, ms);
+    if (!found) throw ms.messages;
+    return augmentTModel(t, tm, TModel(loc f) {
+        // Prevent endless recursion
+        if (f == l) return tm;
+        return tmodelForLoc(f, getPathConfig);
```
DavyLandman, line 580: this is now skipping the caching of the rename framework?

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
```diff
@@ -551,23 +551,42 @@ Tree findCursorInTree(Tree t, loc cursorLoc) {
...
 default list[Tree] extendFocusWithConcreteSyntax(list[Tree] cursor, loc _) = cursor;
 
+@synopsis{
+    Augment the TModel with 'missing' use/def information.
+    Workaround until the typechecker generates this. https://github.com/usethesource/rascal/issues/2172
+}
+TModel augmentTModel(Tree tr, TModel tm, TModel(loc) tmodelForLoc) {
+    tm = augmentFieldUses(tr, tm, tmodelForLoc);
+    tm = augmentFormalUses(tr, tm);
+    return tm;
```
DavyLandman, line 561: I think it would be wise to set a boolean in the tmodel, to mark that we've augmented it. to avoid doing it multiple times?

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/rename/Parameters.rsc
```diff
@@ -45,51 +46,21 @@ tuple[type[Tree] as, str desc] asType(IdRole idRole) = <#Name, "formal parameter
...
-    { kwArg.name.src
-    | kwArg <- keywordArgs
-    , "<kwArg.name>" == argName};
-private set[loc] rascalGetKeywordArgs(\default(_, {KeywordArgument[Expression] ","}+ keywordArgs), str argName) =
-    { kwArg.name.src
-    | kwArg <- keywordArgs
-    , "<kwArg.name>" == argName};
+@synopsis{Add use/def relations for keyword function parameters, until they exist in the TModel.}
+TModel augmentFormalUses(Tree tr, TModel tm) {
+    void addUseDef(loc use, loc def) { tm = tm[useDef = tm.useDef + <use, def>]; }
```
DavyLandman, line 53: shall we inline this function? since it's only used once?

DavyLandman (approved):
LGTM, also like the augmated design

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
toinehartman: This is only used when augmenting. That happens outside of the framework. The framework caches the augmented TModel, so this happens once per module at most.

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
toinehartman: Models are only augmented once and then passed to the framework, which caches them. This function is never called again.

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
toinehartman: There is a similar chaining function in the `Renamer` now. But this happens 'before' the framework, and does not have access to the `Renamer`.

toinehartman: Will change this to use the pre-generated parser

DavyLandman (approved):

PR #599, opened 2025-03-21
Translate IOExceptions to proper VFS exceptions in VS Code

DavyLandman:


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IRascalFileSystemServices.java
```diff
@@ -73,13 +77,8 @@ default CompletableFuture<SourceLocation> resolveLocation(SourceLocation loc) {
...
                 return SourceLocation.fromRascalLocation(resolved);
-            }
-            catch (URISyntaxException e) {
-                IRascalFileSystemServices__logger.warn("Could not resolve location {} due to {}.", loc, e.getMessage());
-                return loc;
-            }
-            catch (Throwable e) {
-                IRascalFileSystemServices__logger.warn("Could not resolve location {} due to {}.", loc, e.getMessage());
+            } catch (Exception e) {
+                IRascalFileSystemServices__logger.warn("Could not resolve location {}", loc, e);
```
PieterOlivier, line 81: Do we really need to catch all exception types here? That always runs the risk of masking programming errors.  Warnings are easily overlooked.

rascal-vscode-extension/src/fs/RascalFileSystemProviders.ts
```diff
@@ -47,6 +47,27 @@ export class RascalFileSystemProvider implements vscode.FileSystemProvider {
...
 
+    sendRequest<R>(uri : vscode.Uri, method: string): Promise<R>;
+    sendRequest<R, A>(uri : vscode.Uri, method: string, param: A): Promise<R>;
+    sendRequest<R, A>(uri : vscode.Uri, method: string, param?: A): Promise<R> {
+        return this.client.sendRequest<R>(method, param ?? { uri: uri.toString()} )
+            .catch((r: ResponseError) => {
+                if (r !== undefined) {
+                    this.client.info("Got responds error from the file system: ", r);
+                    switch (r.code) {
+                        case -1: throw vscode.FileSystemError.FileExists(uri);
```
PieterOlivier, line 58: There are no constants for these that we can use?

PieterOlivier (approved):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IRascalFileSystemServices.java
DavyLandman: Well, the resolve we generally don't want to crash, normally, yes, we forward all exceptions, but resolve shouldn't crash.

rascal-vscode-extension/src/fs/RascalFileSystemProviders.ts
DavyLandman: sadly no. I mean I would have to repeat the constants in both java and typescript, so their added benefit would be very low.

rascal-vscode-extension/src/fs/RascalFileSystemProviders.ts
```diff
@@ -47,6 +47,27 @@ export class RascalFileSystemProvider implements vscode.FileSystemProvider {
         });
     }
 
+    sendRequest<R>(uri : vscode.Uri, method: string): Promise<R>;
+    sendRequest<R, A>(uri : vscode.Uri, method: string, param: A): Promise<R>;
+    sendRequest<R, A>(uri : vscode.Uri, method: string, param?: A): Promise<R> {
+        return this.client.sendRequest<R>(method, param ?? { uri: uri.toString()} )
+            .catch((r: ResponseError) => {
+                if (r !== undefined) {
+                    this.client.info("Got responds error from the file system: ", r);
```
rodinaarssen, line 56:
```suggestion
                    this.client.info("Got response error from the file system: ", r);
```

rodinaarssen (approved):

rascal-vscode-extension/src/fs/RascalFileSystemProviders.ts
DavyLandman: Thanks, fixed. And also made this a `debug` msg instead of `info`.

PR #601, opened 2025-03-25
Rename: type parameters

toinehartman:
Implement the renaming of type parameters. This works around TModel limitations w.r.t. type parameters as described in https://github.com/usethesource/rascal/issues/2185.


PieterOlivier (approved):

toinehartman:
The TModel augmentation step for type params currently shows up in most rename run profiles with 1-3% of runtime. After some investigation, this is probably caused by https://github.com/usethesource/rascal/issues/2204.

PR #622, opened 2025-04-21
Remove parser debouncer

sungshik:
In a previous PR (#475), as part of the integration of error recovery (#490), parse debouncing was added. Presumably, this was because at the time: (a) there were concerns about error recovery performance; (b) it was still unclear which consumer would need to get which tree at which times.

By now, we've learned performance is fine in the vast majority of cases. Also, it was decided each consumer gets each tree asap (a result of which the debounce time rarely expires). Thus, the original reasons for adding parse debouncing no longer apply. This PR removes it.


PieterOlivier (approved):
Awesome simplfication!

Issue #588, opened 2025-02-27
Rename error due to unrelated @ignoreCompiler annotated module


PieterOlivier:
**Describe the bug**

Renaming a symbol can produce an error for not finding a tpl file of a module.
The module for which the tpl file is not found is a module that starts with
an `@ignoreCompilter{}` annotation. 
This module does not contain the symbol that is being renamed.

**To Reproduce**
Create the following modules:
```
module ignored::RegularModule

int thisIsAUniqueName() = 42;
```
and
```
@ignoreCompiler{Test ignored module}
module ignored::IgnoredModule
```

Now try to rename `thisIsAUniqueName` to something else (that is also unique) in the first module. The rename will fail with a "Cannot read TModel for module `sandbox::rename::ignoredModule` information message.

Note that I first tried this in a workspace that contained more Rascal files. Typecheck warning messages for other unrelated modules showed up below the "Cannot read TModel" message.
As can be seen in the screenshot below in this case the unrelated modules IO and Main?  also show up in the clean workspace case.

**Expected behavior**
The rename should succeed normally.

**Screenshots**

![Image](https://github.com/user-attachments/assets/80fab2c3-1f7a-4d74-831a-ae90de164dff)

toinehartman:
Thanks for reporting! It looks like there are two unrelated issues here:
> This module does not contain the symbol that is being renamed.
In that case, there is no need to type-check it. Why does the renaming still trigger a type-check of this module?

> produce an error for not finding a tpl file of a module
Indeed, the renaming assumes that missing TPLs can only be caused by various errors, but in this case, it might be intended. I am not sure what this annotation should mean. We should probably still throw an error, since we cannot rename modules without TPLs, but the error message should clarify this.

PR #626, opened 2025-04-23
Improve handling of ignored modules in renaming

toinehartman:
Properly support ignored modules in renaming.

Discovered in https://github.com/usethesource/rascal-language-servers/issues/588 and was blocked by https://github.com/usethesource/rascal/issues/2513 for a while.


DavyLandman:
Hmm, I thought the `@ignoreCompiler` is only about test execution? Right @PaulKlint ?

toinehartman:
> Hmm, I thought the @ignoreCompiler is only about test execution?

Currently, the compiler does not produce a TModel for `@ignoredCompiler`-annotated modules. `getTModelForModule` returns `false` and the module status contains `ignored()`.

Btw, this PR currently fails with a run-time error w.r.t. `ignored()`. Already informed @PaulKlint about this; he would look into it.

toinehartman:
@PaulKlint Can you enlighten us on the semantics of `@ignore*` annotations and their differences?

PaulKlint:
We have `@ignore`, `@ignoreInterpreter`, `@ignoreCompiler` with the following meaning:
- `@ignore,`: ignore a test while running the interpreter; compiler does not generate code for it.
- `@ignoreInterpreter`: ignore a test while running the interpreter; compiler does generate code for it and runs it
- `@ignoreCompiler`: completely ignore code fragment (including a test); this allows type incorrect code to be ignored.

Issue #630, opened 2025-04-28
Race between `unregisterLanguage` and `registerLanguage`


sungshik:
I'll add details shortly...

DavyLandman:
I think an missing aspect of this analysis is that the register&unregisters are forwarded over a json-rpc bridge from the REPL to the LSP server. On the LSP server side, these requests are handled in a threadpool, and I don't know precisely what the ordering guarantees are.

toinehartman:
This might be caused by how `registerLanguage`/`unregisterLanguage` are routed:
`REPL -> Rascal LSP -> VS Code -> Rascal LSP`

The work in #856 could enable fixing this.

PR #632, opened 2025-04-28
Workaround for flaky UI test

sungshik:
#### This PR...

...provides a workaround for a flaky UI test.

#### Background

The feature PR (#490) to integrate error recovery into the VS Code extension runs (some of) the UI tests first **without** error recovery and then **with** error recovery. Between without/with, Pico needs to be unregistered. This unregistration is new in the feature PR. However, it turns out there are races between unregistration/registration of DSLs that manifest when executed quickly in succession, as reported in #630. As a result, the UI test sometimes fails.

Note: The race exists already in `main`, so it's beyond the scope of the feature PR to fix it. However, unlike `main`, the feature PR triggers the race (because of the new unregistration, which doesn't exist in `main`). To make sure the workflow reliably succeeds in the feature PR, at least a workaround is needed for now.

Note: The workflow sometimes fails for another UI test, but it's very very flaky. I've been able to trigger the same failure locally in `main` as well, so it's not directly related to error recovery. See #638.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
@@ -104,7 +104,7 @@ public class Diagnostics {
         List<Template> diagnostics = new ArrayList<>();
 
         // Highlight selected parts of the error tree
-        if (ERROR_LOCATION_HIGHLIGHT != null) { // NOSONAR
+        if (ERROR_LOCATION_HIGHLIGHT != null) {
             // Just the error location
             ISourceLocation errorLoc = factory.sourceLocation(skippedLoc,
                     skippedLoc.getOffset(), 1,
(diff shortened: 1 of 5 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
```diff
@@ -104,7 +104,7 @@ public static List<Template> generateParseErrorDiagnostics(ITree errorTree) {
         List<Template> diagnostics = new ArrayList<>();
 
         // Highlight selected parts of the error tree
-        if (ERROR_LOCATION_HIGHLIGHT != null) {
+        if (ERROR_LOCATION_HIGHLIGHT != null) { // NOSONAR
```
PieterOlivier, line 107: I tried to squeeze in `// NOSONAR` comments before but the general consensus was that it is not a good idea to introduce tool-specific comments.

rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
```diff
@@ -53,6 +53,18 @@ parameterizedDescribe(function (errorRecovery: boolean) {
...
         await repl.execute("import demo::lang::pico::OldStyleLanguageServer;");
+
+        // If Pico was registered before as part of another series of tests,
+        // then it needs to be unregistered first (because error recovery
+        // en/disabledness affects which contributors to use). Until issue #630
+        // is fixed (race between `unregister` and `register`), the
+        // unregistration can't reliably be done as part of `main` (tried in
+        // commit `a955a05`). Instead, it's done here and followed by a suitably
+        // long sleep.
+        await repl.execute("import util::LanguageServer;");
```
PieterOlivier, line 64: This seems like a decent temporary fix. Have you investigated what a permanent solution could look like?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
DavyLandman: and more over, I agree with sonar, this is dead code, we should not forget this.

PR #635, opened 2025-05-01
[Renaming] Optimize performance

toinehartman:
Many performance optimization including:
- Only augment TModel when necessary
- Cache augmented TModels (just like 'normal' TModels)
- Cache results of `TModel(loc)` function (ljust like `TModel(Tree)`)
- Only check parse tree for occurrences if file contents match (string based)

Benchmark results after optimizations. Elapsed times are 3-run median, in seconds

| project | role of name under cursor | optimized | `main` | latest release |
|---------|------------------------------|-----------------------------------------|----|-|
| bird | local var | 11 | 86 | 45 |
| bird | global function | 34 | 321 | 87 |
| bird | formal param | 6 | 3 | 45 |
| bird | nonterminal | 15 | 639 | 38 |
| bird | module name | 6 | 163 | 81 |
| typepal | local var | 9 | 41 | 39 |
| typepal | constructor | 29 | 311 | (configuration error) |
| typepal | global var | 1 | 1 | (configuration error) |
| rascal | function | 3 | - | - |
| rascal | local var | 1 | - | 2 |
| rascal | type param | 3 | - | 2 |
| rascal | nonterminal label | (configuration error) | - | - |
| rascal | grammar constructor | 71 | - | - |

### DIFF
--- a/rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
+++ b/rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
@@ -316,7 +316,7 @@ set[Define] getCursorDefinitions(list[Tree] cursor, Tree(loc) _, TModel(Tree) _,
         } else if (defs: {_, *_} := tm.useDef[c.src]) {
             // Cursor at use
             cursorDefs = flatMapPerFile(defs, set[Define](loc f, set[loc] localDefs) {
-                localTm = r.getConfig().augmentedTModelForLoc(f, r);
+                localTm = f == cursorLoc.top ? tm : r.getConfig().augmentedTModelForLoc(f, r);
                 return {localTm.definitions[d] | loc d <- localDefs};
             });
         } else {
--- a/rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
+++ b/rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
@@ -233,12 +233,22 @@ Tree findCursorInTree(Tree t, loc cursorLoc) {
 list[Tree] extendFocusWithConcreteSyntax([Concrete c, *tail], loc cursorLoc) = [findCursorInTree(c, cursorLoc), c, *tail];
 default list[Tree] extendFocusWithConcreteSyntax(list[Tree] cursor, loc _) = cursor;
 
+data AugmentComponents = augmentUses() | augmentDefs();
+
+bool requiresAugmentation(set[Define] defs, {augmentUses(), *_}) = (defs.idRole & {constructorId(), functionId(), fieldId(), keywordFieldId(), keywordFormalId(), typeVarId()}) != {};
+bool requiresAugmentation(set[Define] defs, {augmentDefs()}) = (defs.idRole & {fieldId(), typeVarId()}) != {};
+
+TModel getConditionallyAugmentedTModel(loc l, set[Define] defs, set[AugmentComponents] useOrDef, Renamer r)
+    = requiresAugmentation(defs, useOrDef)
+    ? r.getConfig().augmentedTModelForLoc(l, r)
+    : r.getConfig().tmodelForLoc(l);
+
 @synopsis{
     Augment the TModel with 'missing' use/def information.
     Workaround until the typechecker generates this. https://github.com/usethesource/rascal/issues/2172
 }
-TModel augmentTModel(loc l, TModel tm, PathConfig(loc) getPathConfig) {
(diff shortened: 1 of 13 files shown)


rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/rename/Common.rsc
```diff
@@ -171,39 +194,116 @@ tuple[bool, bool](Tree) allNameSortsFilter(str name1, str name2) {
...
+                    for (qn <- qualifiedNames, qn := n || qn := [QualifiedName] normalizeEscaping("<n>")) return true;
+                }
+            }
+        }
+
+        return false;
+    };
+}
+
+bool(loc) containsQualifiedNameCheck(str _) =
```
DavyLandman, line 220: Why does this one always return true? (Add a comment on the reasoning, like fallback to parse tree or something)

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/rename/Common.rsc
```diff
@@ -171,39 +194,116 @@ tuple[bool, bool](Tree) allNameSortsFilter(str name1, str name2) {
...
+    singleNameTreeFilter2 = singleNameFilter(name2);
+
     for (loc f <- fs) {
         jobStep(j, "<f>");
-        tr = getTree(f);
-        <l, r> = treeFilter(tr);
-        if (l) fs1 += f;
-        if (r) fs2 += f;
+        if (loosePreCheck1(f)) {
+            if (loosePreCheck2(f)) {
```
DavyLandman, line 269: This potentially reads the file twice, what if we do a read, get the string, and use that for both checks?

(There is even a parse function overload where you can pass in the string and the loc).

DavyLandman (approved):
Some small comments. Looks good.

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/Rename.rsc
```diff
@@ -233,76 +233,129 @@ Tree findCursorInTree(Tree t, loc cursorLoc) {
...
-      , getPathConfig = getPathConfig
-      , debug = false
-  )
-);
+    @synopsis{
+        Augment the TModel with 'missing' use/def information.
+        Workaround until the typechecker generates this. https://github.com/usethesource/rascal/issues/2172
+    }
+    @memo{maximumSize(100), expireAfter(minutes=5)}
+    TModel augmentTModel(loc l, Renamer r) {
```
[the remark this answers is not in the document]
toinehartman: Just like the parse and type-check functions, this one is scoped to the rename. So as long as the user does not re-use the `RenameConfig` and does not change files during renaming, all is well.

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/rename/Common.rsc
toinehartman: Unused, removed

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/rename/Common.rsc
```diff
@@ -171,39 +194,116 @@ tuple[bool, bool](Tree) allNameSortsFilter(str name1, str name2) {
...
+bool(Tree) anyNameFilter(set[str] names) {
+    set[str] escNamesS = {normalizeEscaping(name) | name <- names};
+    set[Name] escNames = {*escapeSet(#Name, name) | name <- escNamesS};
+    set[Nonterminal] escNonterminals = {*escapeSet(#Nonterminal, name) | name <- escNamesS};
+    set[NonterminalLabel] escNonterminalLabels = {*escapeSet(#NonterminalLabel, name) | name <- escNamesS};
+    set[QualifiedName] qualifiedNames = {t | name <- escNamesS, just(QualifiedName t) := tryParseAs(#QualifiedName, name)};
+
+    return bool(Tree tr) {
+        visit (tr) {
+            case (Name) `<Name n>`: for (n1 <- escNames, n := n1) return true;
```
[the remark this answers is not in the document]
toinehartman: Right, fixed

rascal-lsp/src/main/rascal/lang/rascal/lsp/refactor/rename/Common.rsc
toinehartman: Changed this to read the file once for all pre-checks. The parse overload is a bit tricky to use, since it does not make use of the caching by the framework. I think the file read times are dwarfed by actual parse times, so I am not sure this is an optimization we need to do now.

PR #645, opened 2025-05-12
Insert warning and build link for pre-release builds.

toinehartman:
Clearly mark extension as pre-release and link to origin build (and therefore commit) in README.
afbeelding


toinehartman:
Note: history is a mess, so this should be squashed.

PR #651, opened 2025-05-16
Design LSP rename API

toinehartman:
# Design: support renaming in DSL LSPs

The renaming functionality consists of multiple LSP endpoints.

- Language features
  * Prepare rename
  * Rename
- Workspace features
  * Will rename files
  * Did rename files

### Prepare rename
[*Prepare rename*](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_prepareRename) selects, given a cursor position, the range in the file that should be renamed. It defaults to using word boundaries, but the user might want to override this behaviour, for example:
- When renaming qualified names like `foo::bar`. By default, VS Code would select `foo` or `bar`, but one might want to rename the complete thing.
- When renaming certain names is not supported, and this can be quickly determined (i.e. based on the parse tree). For example, the language allows renaming local variables, but not function calls.

In the Rascal LSP API, this can be modeled by the following function.
```rascal
loc (Focus _focus) prepareRenameService;
```
Unsupported renames show return `null` (in Java). The DSL can throw a specific exception or a string here, that we catch and handle in Java.

### Rename
[*Rename*](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_rename) takes a position in the document and a new name, and renames occurrences of the name at that position to the new name. It does so by compiling a list of proposed edits, that the VS Code will then apply. Besides the edits, a rename might also result in diagnostic messages. These serve to signal issues with the renaming, for example when renaming an identifier would result in duplicate definitions. In cases where the DSL is type-checked using *typepal*, a basic renaming can be quickly obtained by using the [rename framework](https://github.com/usethesource/typepal/pull/14/files#diff-331b6ee7eb6cff5a20ed758324cb50852cd52a0ef2e6c39028bbee40a9bc6963). Therefore, it would be advantageous if this endpoint integrates well with that.

In the Rascal LSP API, this can be modeled by the following function:
```rascal
tuple[list[DocumentEdit], set[Message]] (Focus _focus, str newName) renameService;
```

We group the `prepareRename` and `rename` services together under a single constructor, since they are practically married. For `prepareRename`, we can provide a sensible default.

```rascal
import Exception;

data LanguageService
  = rename(tuple[list[DocumentEdit], set[Message]] (Focus _focus, str newName) renameService
        , loc (Focus _focus) prepareRenameService = defaultPrepareRenameService)
  ;

loc defaultPrepareRenameService(Focus _:[Tree tr, *_]) = tr.src when tr.src?;
default loc defaultPrepareRenameService(Focus focus) { throw IllegalArgument(focus, "Element under cursor does not have source location"); }
```

### Will/Did rename files
The *[Will](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_willRenameFiles)/[Did rename files](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didRenameFiles)* endpoints are called by VS Code before/after files are renamed from the VS Code explorer. They allow the extension to change code references to those file names to match the renaming.

- *Will rename files* is a request, and needs to return a collection of edits (or error in some way - `null` in Java). Since this is very sensitive to evaluator locks, we cannot guarantee it will ever succeed. Therefore, we omit it, until there is a very clear use-case for it.
- *Did rename files* is a notification; the IDE does not wait for it to return and it is itself responsible for applying the edits to the workspace (e.g. via `analysis::diff::edits::ExecuteTextEdits`). We model the API so that the implementor must return edits, which we then apply from the Java wrapper.

```rascal
data LanguageService
  = didRenameFiles(void(map[loc old, loc new] fileRenames) didRenameFilesService)
  ;
```

### DIFF
--- a/rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
+++ b/rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
@@ -52,7 +52,9 @@ set[LanguageService] picoLanguageServer() = {
     execution(picoExecutionService),
     inlayHint(picoInlayHintService),
     definition(picoDefinitionService),
-    codeAction(picoCodeActionService)
+    codeAction(picoCodeActionService),
+    prepareRename(picoRenamePreparingService),
+    rename(picoRenamingService)
 };
 
 @synopsis{This set of contributions runs slower but provides more detail.}
@@ -176,6 +178,15 @@ value picoExecutionService(removeDecl(start[Program] program, IdType toBeRemoved
     applyDocumentsEdits([changed(program@\loc.top, [replace(toBeRemoved@\loc, "")])]);
     return ("result": true);
 }
+
+loc picoRenamePreparingService(Focus _:[Id id, *_]) = id.src;
+
+tuple[list[DocumentEdit], set[Message]] picoRenamingService(Focus focus, str newName) = <[changed(focus[0].src.top, [
+    replace(id.src, newName)
+    | cursor := focus[0]
+    , /Id id := focus[-1]
+    , id := cursor
+])], {}>;
 
 @synopsis{The main function registers the Pico language with the IDE}
 @description{
--- a/rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
+++ b/rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
@@ -53,8 +53,7 @@ set[LanguageService] picoLanguageServer() = {
     inlayHint(picoInlayHintService),
     definition(picoDefinitionService),
     codeAction(picoCodeActionService),
-    prepareRename(picoRenamePreparingService),
-    rename(picoRenamingService)
+    rename(picoRenamingService, prepareRenameService = picoRenamePreparingService)
 };
 
 @synopsis{This set of contributions runs slower but provides more detail.}
--- a/rascal-lsp/src/main/rascal/util/LanguageServer.rsc
+++ b/rascal-lsp/src/main/rascal/util/LanguageServer.rsc
@@ -274,7 +274,7 @@ data LanguageService
     | codeAction    (list[CodeAction] (Focus _focus) codeActionService)
     | rename        (tuple[list[DocumentEdit], set[Message]] (Focus _focus, str newName) renameService
         , loc (Focus _focus) prepareRenameService = defaultPrepareRenameService)
-    | didRenameFiles(tuple[list[DocumentEdit], set[Message]] (lrel[loc old, loc new] fileRenames) didRenameFilesService)
+    | didRenameFiles(tuple[list[DocumentEdit], set[Message]] (list[DocumentEdit] fileRenames) didRenameFilesService)
     ;
 
 loc defaultPrepareRenameService(Focus _:[Tree tr, *_]) = tr.src when tr.src?;
--- a/rascal-lsp/src/main/rascal/util/LanguageServer.rsc
+++ b/rascal-lsp/src/main/rascal/util/LanguageServer.rsc
@@ -209,7 +209,7 @@ hover documentation, definition with uses, references to declarations, implement
 * The ((execution)) service executes the commands registered by ((lenses)) and ((inlayHinter))s.
 * The ((actions)) service discovers places in the editor to add "code actions" (little hints in the margin next to where the action is relevant) and connects ((CodeAction))s to execute when the users selects the action from a menu.
 * The ((util::LanguageServer::rename)) service renames an identifier by collecting the edits required to rename all occurrences of that identifier. It might fail and report why in diagnostics.
-   * The optional ((prepareRename)) service argument discovers places in the editor where a ((util::LanguageServer::rename)) is possible. If renameing the location is not supported, it should throw an exception.
+   * The optional `prepareRename` service argument discovers places in the editor where a ((util::LanguageServer::rename)) is possible. If renameing the location is not supported, it should throw an exception.
 * The ((didRenameFiles)) service collects ((DocumentEdit))s corresponding to renamed files (e.g. to rename a class when the class file was renamed). The IDE applies the edits after moving the files. It might fail and report why in diagnostics.
 
 Many services receive a ((Focus)) parameter. The focus lists the syntactical constructs under the current cursor, from the current
--- a/rascal-lsp/src/main/rascal/util/LanguageServer.rsc
+++ b/rascal-lsp/src/main/rascal/util/LanguageServer.rsc
@@ -209,7 +209,7 @@ hover documentation, definition with uses, references to declarations, implement
 * The ((actions)) service discovers places in the editor to add "code actions" (little hints in the margin next to where the action is relevant) and connects ((CodeAction))s to execute when the users selects the action from a menu.
 * The ((prepareRename)) service discovers places in the editor where a ((util::LanguageServer::rename)) is possible.
 * The ((util::LanguageServer::rename)) service renames an identifier by collecting the edits required to rename all occurrences of that identifier. It might fail and report why in diagnostics.
-* The ((willRenameFiles)) service collects ((DocumentEdit))s corresponding to renaming files (e.g. to rename a class when the class file will be renamed). The IDE applies the edits before moving the files.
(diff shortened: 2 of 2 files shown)


rascal-lsp/src/main/rascal/demo/lang/pico/LanguageServer.rsc
```diff
@@ -177,6 +179,15 @@ value picoExecutionService(removeDecl(start[Program] program, IdType toBeRemoved
...
 }
 
+loc picoRenamePreparingService(Focus _:[Id id, *_]) = id.src;
+
+tuple[list[DocumentEdit], set[Message]] picoRenamingService(Focus focus, str newName) = <[changed(focus[0].src.top, [
+    replace(id.src, newName)
+    | cursor := focus[0]
+    , /Id id := focus[-1]
+    , id := cursor
+])], {}>;
```
toinehartman, line 189: This is just (like most of the code in this file) a very simple, syntax-based example.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -268,6 +272,10 @@ data LanguageService
     | references    (set[loc] (Focus _focus) referencesService)
     | implementation(set[loc] (Focus _focus) implementationService)
     | codeAction    (list[CodeAction] (Focus _focus) codeActionService)
+    | prepareRename (loc (Focus _focus) prepareRenameService)
```
toinehartman, line 275: How to return a failure here (for unsupported renames)?
- `Maybe[loc]`
- throw an exception
- ...

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -268,6 +272,10 @@ data LanguageService
...
+    | rename        (tuple[list[DocumentEdit], set[Message]] (Focus _focus, str newName) renameService)
```
toinehartman, line 276: Publishing the diagnostic messages resulting from the rename overlaps with `util::IDEServices::registerDiagnostics`. Can we reduce this overlap? We still want to abort a rename when an `error` was registered, even when edits were registered before the error.

*(also applies to `willRenameFiles`)*

toinehartman, line 276: Renaming might want to know about the folders/projects open in the workspace. Currently, `util::LanguageServer` does not expose those. How can we achieve that?

*(also applies to `willRenameFiles`/`didRenameFiles`)*

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
```diff
@@ -268,6 +272,10 @@ data LanguageService
...
+    | willRenameFiles(tuple[list[DocumentEdit], set[Message]] (map[loc old, loc new] fileRenames) willRenameFilesService)
+    | didRenameFiles(void(map[loc old, loc new] fileRenames) didRenameFilesService)
```
toinehartman, line 278: The `map[loc, loc]` that contains the renamed files could also be a `rel[loc, loc]`. That is closer to the LSP spec, which uses a `FileRename[]` here (`FileRename` is basically a named pair). However, the `rel` is a bit confusing; can we have multiple occurrences of original file names, renamed to different target names?

DavyLandman:
This is a close mapping of the original LSP protocol. In the other APIs we've often tried to improve it a bit where we can, here are some questions I have after reading it:

- how often is `prepareRename` called? only after someone presses f2? or much more frequent. if the latter, perhaps we can do something where a user can provide all renamable positions on a tree?
- the distinction between `didRename` and `willRename` is a bit finicky, especially since `willRename` requires us to respond quickly, and due to evaluator locks we have no guarantees a line of rascal code is even executed before the timeout passes? Perhaps we should only keep `didRename`?
- Like you pointed out, `registerDiagnostics` already exist, and a similar one for `ExecuteTextEdits`. I think it's wise to model the API to enforce the user to do the right thing. So maybe `didRenameFiles` should also return a list of text edits and messages.
- Perhaps rename should be bundled into a single constructor, with a few nested optional closures, where we provide a working default, and a user is allowed to override it. The terminology in that constructor will be the same, but at least we'll group them together?

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
DavyLandman: I think throwing an exception is the best option? and otherwise a custom ast:

```rascal
data RenameOption = fullTarget(loc l) | notPossible(Message m);
```

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
DavyLandman: that discussion on errors in a operation is broader than this api. normally we use exceptions for true errors.

DavyLandman: hmm, yes, but that should be a generic thing. This is also something that @rodinaarssen noticed when looking at the `RascalIDEServices` class, that that was missing, so we might have to add that concept of a workspace to rascal.

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
DavyLandman: I think if it's a list in LSP, it might map better onto an `lrel` to keep the order.

toinehartman:
> how often is prepareRename called? only after someone presses f2? or much more frequent. if the latter, perhaps we can do something where a user can provide all renamable positions on a tree?

`prepareRename` is called when the user presses F2, so I don't think we need a smarter solution here.

> the distinction between didRename and willRename is a bit finicky, especially since willRename requires us to respond quickly, and due to evaluator locks we have no guarantees a line of rascal code is even executed before the timeout passes? Perhaps we should only keep didRename?

I understand the point, but I'm not sure if this means we should not expose the `willRenameFiles` API at all. I prefer providing some extra documentation on this to make the distiction between both endpoints more clear, and to warn about the caveats of `willRenameFiles`?

>  So maybe didRenameFiles should also return a list of text edits and messages.
👍 

> Perhaps rename should be bundled into a single constructor, with a few nested optional closures, where we provide a working default, and a user is allowed to override it. The terminology in that constructor will be the same, but at least we'll group them together?

Making `prepareRename` an optional nested closure to the `rename` constructor makes sense to me. The only sensible default for `rename` itself is thorwing an unsupported exception, I suppose. Like @sungshik already pointed out elsewhere, having `prepareRename` without `rename` does not make any sense.

I think the `did/willRenameFiles` endpoints can also be grouped together, since typically you'll only implement one of both. I don't think they should be grouped with `(prepare)Rename`, since they are workspace endpoints (instead of document) and do no hook into the rename framework seamlessly.

toinehartman:
Btw, `prepareRename` already has a very sensible default in VS Code (word boundaries I think). Do we really need to provide a default in `util::LanguageServer`, given that that will always go via an evaluator (contrary to VS Code's own default)?

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
toinehartman: So, I am not sure if I understad - would it be fine to return a `set[Message]` here?

toinehartman: I think we want to associate the reasons for failure to locations (for example, we cannot rename because that would cause shadowing at `x`). That's where the `Message` comes in handy.

DavyLandman: I think that there is an orthogonal discussion about how we report back errors from any language contribution. That is not specific to `rename` only.

Like what if I want to report an error when someone requests a `definition` or a `codeAction`?

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
toinehartman: I agree. Just throwing a string (failure message) here might be the easiest in most cases?

rascal-lsp/src/main/rascal/util/LanguageServer.rsc
toinehartman: I see. I'll leave this how it is for now.

toinehartman:
Thanks for the reviews @DavyLandman @jurgenvinju @PaulKlint. I'll go ahead an merge this to `feature/rename-refactoring/parametric`. There, we can also develop the implementation and some examples, before merging the feature to `main`.

PR #662, opened 2025-06-12
Update Rascal and incorporate changes to Evaluator, PathConfig.

rodinaarssen:
* Bumped Rascal to RC49
* Updated TestUtils to incorporate changes to PathConfig
* Removed code in several places (mainly LSPTerminalREPL). The removed functionality has been moved to Rascal


rodinaarssen:
Rascal was bumped to an RC version that is yet to be released. It should contain the fix in https://github.com/usethesource/rascal/pull/2300, which will fix the checker and make the tests succeed.

rascal-lsp/src/main/rascal/lang/rascal/tests/rename/TestUtils.rsc
```diff
@@ -230,7 +232,7 @@ public loc calculateRascalLib() {
 public PathConfig getTestPathConfig(loc testDir) {
     return pathConfig(
         bin=testDir + "bin",
-        resources=testDir + "bin",
+        resources=[testDir + "bin"],
```
DavyLandman, line 235: `resources` should not be set anymore.

rascal-lsp/src/main/rascal/lang/rascal/tests/rename/TestUtils.rsc
```diff
@@ -88,7 +88,9 @@ private void verifyTypeCorrectRenaming(loc root, list[DocumentEdit] edits, PathC
     copy(root, backupLoc, recursive = true);
 
     executeDocumentEdits(sortEdits(groupEditsByFile(edits)));
-    remove(pcfg.resources);
+    for (res <- pcfg.resources) {
```
DavyLandman, line 91: I think we need to look at bin now

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
```diff
@@ -310,9 +310,19 @@ public static CompletableFuture<Evaluator> makeFutureEvaluator(ExecutorService e
             String jobName = "Loading " + label;
             try {
                 services.jobStart(jobName, imports.length);
-                Evaluator eval = ShellEvaluatorFactory.getDefaultEvaluator(Reader.nullReader(),
-                        IoBuilder.forLogger(customLog).setLevel(Level.INFO).buildPrintWriter(),
-                        IoBuilder.forLogger(customLog).setLevel(Level.ERROR).buildPrintWriter(), services);
+                Evaluator eval;
+                Reader nullReader = Reader.nullReader();
```
DavyLandman, line 314: nit: the left side is an clone of the right side.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
```diff
@@ -310,9 +310,19 @@ public static CompletableFuture<Evaluator> makeFutureEvaluator(ExecutorService e
...
+                PrintWriter infoWriter = IoBuilder.forLogger(customLog).setLevel(Level.INFO).buildPrintWriter();
```
DavyLandman, line 315: nit: a better name would be `var out` and `var err`. since that is what we're building, the stdout & stderr writers for the evaluator.

DavyLandman (changes requested):
some small nit's on the variable names.

PR #664, opened 2025-06-25
Fixed janky "Run in new Rascal terminal" codelens.

PieterOlivier:
Note that the current solution (catch any exception) is not ideal.
We can replace this with a more specific exception once we fix the `TreeAdapter.getArg` method in Rascal to properly handle error trees.


DavyLandman:
The original code (before #490 was merged) would not update the code lens for as long as there was an parse error. That made it a lot less jumpy.

Now this PR (which is already better than before) still might drop the code lens if the error recovery couldn't recover the `main` function definition, say for example a unaligned `}`? or an unclosed string?

Maybe there is a way to make this more stable? Maybe thats a follow up PR, but I'm just wondering if there is a way to say: ow, this tree had error recovery, so if we couldnt find a main,maybe not return an updated list of codelenses?

PR #667, opened 2025-06-26
Moved lsp internal code to specific subfolder as to hide it from downstream users

### LINKED ISSUE #663: Split up Rascal for Rascal code from Rascal for users code in rascal-lsp
**Describe the bug**

This also addresses #656 .

Let's spit up the code in `rascal-lsp/src/main/rascal`:
* `src/main/rascal/library` for `util::LanguageServer` and possible generically usable friends.
* `src/main/rascal/rascal-lsp` for the renaming framework code and `util::Util`, and also the outline and the checker wrapper code. Anything that is part of the implementation of  the Rascal IDE and _not_ a generically reusable tool (for any DSL).

This would lower the urgency of divvying up `util::Util` over the rascal standard library. We can fix #656 on our own pace, even after a release.

Watch out for the maven pom.xml editors: `src/main/rascal/rascal-lsp`'s `.rsc` and `.tpl` files should not end up at the root of the `jar`. Otherwise they will leak into user space accidentally.

DavyLandman:
In the process the evaluator construction code was also simplified, the function was getting so many parameters it was hard to understand how it was used and supposed to function. So this was rewriting to improve the readability (I hope).

Fixes #663


jurgenvinju (approved):
Looks good to me, also factoring all the parameters in LSPContext.

Should we add a pom profile for type checking the rascal code that did not go into the library?

DavyLandman:
We can't since rascal.jar doesn't contain the tpls for the compiler where much of the `/lsp/` directory depends on.

PR #673, opened 2025-07-07
Use lastTreeWithoutErrors for calculating code lenses

### LINKED ISSUE #665: Tweak rascal's code lens calculation (import & run) with relation to error recovery
> The original code (before #490 was merged) would not update the code lens for as long as there was an parse error. That made it a lot less jumpy.
> 
> Now this PR (which is already better than before) still might drop the code lens if the error recovery couldn't recover the `main` function definition, say for example a unaligned `}`? or an unclosed string?
> 
> Maybe there is a way to make this more stable? Maybe thats a follow up PR, but I'm just wondering if there is a way to say: ow, this tree had error recovery, so if we couldnt find a main,maybe not return an updated list of codelenses?

_Originally posted by @DavyLandman in https://github.com/usethesource/rascal-language-servers/issues/664#issuecomment-3003697398_

PieterOlivier:
The "Run in new Rascal terminal" code lens would disappear if you introduced certain parse errors.
With this fix, the code lenses are always calculated based on the last tree without errors
so the code lense will not disappear until the source code parses without errors and there is no longer a main function.

This PR closes #665.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
@@ -542,6 +542,8 @@ public class RascalTextDocumentService implements IBaseTextDocumentService, Lang
                 return r;
             })
             .thenApply(Versioned::get)
+            // Replace with the last tree without errors, might be the same as `tree` if the parse succeeded without any error recovery
+            .thenApply(tree -> f.getLastTreeWithoutErrors().get())
             .thenApplyAsync(rascalServices::locateCodeLenses, ownExecuter)
             .thenApply(List::stream)
             .thenApply(res -> res.map(this::makeRunCodeLens))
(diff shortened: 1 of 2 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
```diff
@@ -542,6 +542,8 @@ public CompletableFuture<List<? extends CodeLens>> codeLens(CodeLensParams param
                 return r;
             })
             .thenApply(Versioned::get)
+            // Replace with the last tree without errors, might be the same as `tree` if the parse succeeded without any error recovery
+            .thenApply(tree -> f.getLastTreeWithoutErrors().get())
```
DavyLandman, line 546: This feels very temporal coupling to the semantics of `getLastTreeWithoutErrors` (also evidenced by the fact that some logic had to be switched over there).

Are we sure this is enough of a fix? Since looking at the `TextDocumentState` it looks more like the `getLastTreeWithoutErrors` is about parse exceptions and not about trees with error recovery or not.

DavyLandman (commented):
I'm unsure if this is enough of a fix.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
PieterOlivier: `lastTreeWithoutErrors` is only updated if no parse error or exception occured and the parse tree contains no errors. 

I think updating `lastWithoutErrors` before completing the `treeAsync` future is actually a sensible fix. Otherwise the `lastTreeWithoutErrors` premise is not true in the code executed on completion of the future.

I have already tested this fix and it seems to work great!

PR #674, opened 2025-07-07
User-interruptible evaluator jobs

### LINKED ISSUE #672: Allow a user to interrupt/cancel a running typechecker
Sometimes the typechecker takes to long to run, especially for the rascal-typechecker. 

Allow the user to cancel/interrupt that checking session if they don't care for the output.

Making the typechecker faster ;)

toinehartman:
Allow the developer to interrupt evaluator tasks, for example to cancel long-running typechecker jobs. Cancelling interrupts the Rascal interpreter.

Rascal example:
afbeelding

Pico example:
afbeelding

Closes #672.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
@@ -79,6 +79,7 @@ import org.rascalmpl.vscode.lsp.BaseWorkspaceService;
 import org.rascalmpl.vscode.lsp.IBaseLanguageClient;
 import org.rascalmpl.vscode.lsp.IBaseTextDocumentService;
 import org.rascalmpl.vscode.lsp.LSPIDEServices;
+import org.rascalmpl.vscode.lsp.RascalLSPMonitor;
 import org.rascalmpl.vscode.lsp.rascal.RascalLanguageServer;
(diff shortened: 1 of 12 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
```diff
@@ -93,23 +94,42 @@ public class EvaluatorUtil {
     public static <T> InterruptibleFuture<T> runEvaluator(String task, CompletableFuture<Evaluator> eval, Function<Evaluator, T> call, T defaultResult, Executor exec, boolean isParametric, LanguageClient client) {
         AtomicBoolean interrupted = new AtomicBoolean(false);
         AtomicReference<@Nullable Evaluator> runningEvaluator = new AtomicReference<>(null);
-        return new InterruptibleFuture<>(eval.thenApplyAsync(actualEval -> {
+        AtomicReference<InterruptibleFuture<T>> future = new AtomicReference<>();
+
+        future.set(new InterruptibleFuture<>(eval.thenApplyAsync(actualEval -> {
             try {
+                while (future.get() == null) {
```
DavyLandman, line 101: to be completly threadsafe, we should store it in a local variable and check that against null.

For example:

```
InterruptibleFuture<T> ourself;
while((ourself = future.get()) == null) {
   Thread.yield();
}
```
and use `ourself`  afterwards and know for a fact it cannot be null.

DavyLandman (approved):
Thanks :+1: some small nits.

PR #677, opened 2025-07-11
Parametric: formatting

### LINKED ISSUE #130: Expose LSP formatter (pretty-printer) api via LanguageService
[spec](https://microsoft.github.io/language-server-protocol/specification.html#textDocument_formatting)

toinehartman:
* [x] Implementing the [formatting](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_formatting) and [rangeFormatting](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_rangeFormatting) LSP APIs. Both use the same contribution, where for (whole-file) formatting `_range == _input@\loc`.

Design:
```
data LanguageService = formatting(list[TextEdit](Focus _focus, FormattingOptions _opts) formattingService);

data FormattingOptions(
        int tabSize = 4
      , bool insertSpaces = true
      , bool trimTrailingWhitespace = false
      , bool insertFinalNewline = false
      , bool trimFinalNewlines = false
) = formattingOptions();
```

Closes #130.

* [x] added formatting support for Rascal as well, since that shares the essential new focus computation based on a selection range with the current PR.


rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
```diff
@@ -268,8 +268,19 @@ data LanguageService
     | references    (set[loc] (Focus _focus) referencesService)
     | implementation(set[loc] (Focus _focus) implementationService)
     | codeAction    (list[CodeAction] (Focus _focus) codeActionService)
+    | formatting    (str (Tree _input, FormattingOptions _opts) formattingService)
     ;
 
+data FormattingOptions
+    // If LSP adds more options, add them as keyword arguments for backward compatibility
```
rodinaarssen, line 275: This means that the current options are (and will remain) positional parameters, whereas future extensions will become keyword parameters. As ADTs are extensible, have you considered distinct constructors for the various formatting options and passing a `list` of options to the `formattingService`?

toinehartman: Thanks, I keep forgetting how a generator can be used to pattern match on single elements.

DavyLandman (changes requested):
I see this is a bit of a challenging API, especially if we want to make this more user-friendly for DSL users. 

I have the feeling we should involve @jurgenvinju who has been on the topic of formatting/pretty-printing for quite some time already. I suspect that we're missing some opportunities of matching this up with something already in the standard library, or something that we might should add to the standard library first.

DavyLandman:
This PR is missing the adjacent: [`rangeFormatting`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_rangeFormatting) functionality.

rascal-lsp/src/main/rascal/library/demo/lang/pico/LanguageServer.rsc
```diff
@@ -58,9 +67,93 @@ set[LanguageService] picoLanguageServer(bool allowRecovery) = {
     inlayHint(picoInlayHintService),
     definition(picoDefinitionService),
     codeAction(picoCodeActionService),
-    selectionRange(picoSelectionRangeService)
+    selectionRange(picoSelectionRangeService),
+    formatting(picoFormattingService)
 };
 
+list[TextEdit] picoFormattingService(Tree input, loc range, FormattingOptions opts) {
```
DavyLandman, line 74: I think the implmentation of this should be moved out side of the `LanguageServer` file, as it's quite big, and is just a separate formatter. It might be better if we could move this to rascal std lib, where the pico example lives?

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
```diff
@@ -270,8 +270,17 @@ data LanguageService
     | implementation(set[loc] (Focus _focus) implementationService)
     | codeAction    (list[CodeAction] (Focus _focus) codeActionService)
     | selectionRange(list[loc](Focus _focus) selectionRangeService)
+    | formatting    (list[TextEdit](Tree _input, loc range, FormattingOptions _opts) formattingService)
```
DavyLandman, line 273: I like that `range` is a parameter. Please also add a documentation section that introduces how formatting works and how this range parameter should be interpreted (like when is it the whole file? if it's equal to `.top`?)

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
```diff
@@ -270,8 +270,17 @@ data LanguageService
...
     ;
 
+data FormattingOptions(
+        int tabSize = 4
+      , bool insertSpaces = true
```
DavyLandman, line 278: we should document that if `insertSpaces` is `false` it means: use tabs for indentation.

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
```diff
@@ -270,8 +270,17 @@ data LanguageService
     | implementation(set[loc] (Focus _focus) implementationService)
     | codeAction    (list[CodeAction] (Focus _focus) codeActionService)
     | selectionRange(list[loc](Focus _focus) selectionRangeService)
+    | formatting    (list[TextEdit](Tree _input, loc range, FormattingOptions _opts) formattingService)
     ;
 
+data FormattingOptions(
```
DavyLandman, line 276: how do we know the default value? like if LSP does not send `trimTrailingWhitespace` (since it's an optional boolean) how do we know that means `true` and not `false`?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/InterpretedLanguageContributions.java
```diff
@@ -350,6 +359,16 @@ public InterruptibleFuture<IList> selectionRange(IList focus) {
         return execFunction(LanguageContributions.SELECTION_RANGE, selectionRange, VF.list(), focus);
     }
 
+    @Override
+    public InterruptibleFuture<IList> formatting(ITree input, ISet formattingOptions) {
+        debug(LanguageContributions.FORMATTING, input != null ? TreeAdapter.getLocation(input) : null, formattingOptions.size());
+        return InterruptibleFuture.flatten(formatting.thenCombine(parsing, Pair::of)
```
DavyLandman, line 365: should we make `execFunction` more generic so we can deal with this variant?

DavyLandman (commented):
I think this is going in the right direction, but I'm not sure what I should have been reviewing, the implementation or the design?

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
toinehartman: Yes, this all needs more documentation still. Although I think @jurgenvinju was not so sure about the meaning of this parameter...

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
toinehartman: I think the optional is only missing when either the server (which we control via our dependency version) or the client (which we do not control) are on an older version of the protocol.
I had a look at the protocol and LSP4J also provides some [default values](https://github.com/eclipse-lsp4j/lsp4j/blob/main/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend#L4337-L4412). We could copy those - although I don't think `tabSize = 0` is a sensible default.

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
DavyLandman: We talked about it, and look at the behavior of VS Code, and some discussion based on that.

- `insertSpaces = true` means indent with spaces (of `tabSize` width). Replace tabs used for indentation with spaces.
- `insertSpaces = false` means indent using tabs. Replace spaces used for indentation with tabs.

The realisation was that if you didn't have this behavior, you could never normalize the indentation patterns for a file with mixed tabs&spaces.

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
DavyLandman: The problem comes from typescript (and javascript) where `trimTrailingWhitespace?: boolean;` is a field that can become `undefined`, `false`, or `true`. But `undefined` is ["falsy"](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) so many places that would write `if (settings.trimTrailingWhitespace)` would mean that only if it's actually set to `true` does it have a value.

Also, more in this topic `tabSize` (and `insertSpaces`) is not optional in the [FormattingOptions](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#formattingOptions). So maybe it shouldn't be treated as a kw param?

toinehartman: Not sure. When they added some of the optional properties for 3.15, they already had this: https://github.com/eclipse-lsp4j/lsp4j/commit/21c94d0b9536a9f8a31892c4134c7ab072999dc1#diff-7db90b832b0659f70a7dd341bb1ef0d9231ca767614dd73492c3e2d22d54f968

Maybe just a very defensive implementation?

toinehartman: Discussed offline: we inspected LSP and VS Code behaviour and decided that the default values for optionals here should be `false`.

jurgenvinju:
Is there an opportunity to use the Focus abstraction for this contribution? Especially for the range alternative. 

That way language engineers don't have to go search for the right trees anymore, and they could easily support only a few top-level types for starters, and they could easily recover the required indentation level from the parent tree's layout siblings.

toinehartman:
> Is there an opportunity to use the Focus abstraction for this contribution? Especially for the range alternative.
@jurgenvinju I thought about this, but the tricky part there is that a range does not necessarily correspond to an exact tree. Even if we give a focus tree that ends at the largest tree encapsulating the range, there would still be the need to filter the edits so they are not outside of the given range.

jurgenvinju:
We talked on slack and agreed we do need the focus. 

We don't have to guarantee per SE the edits fall into the given range; but we should make it easy for engineers to format and indent the smallest syntactic unit around the given range. That way edits will also probably fall into the range and we can use the parse trees for formatting (syntax directed) like all the other features. 

Also formatters may require access to the parent of the formatted node, for proper indentation. Focus also provides that.

toinehartman:
There has been some discussion about how we should treat formatting a range that does not directly correspond to a sub-tree. Particularly interesting is the case where the selection covers multiple, but not all, elements in a syntactic list (e.g. `Statement*`).

Let's list some alternatives.

### Option 1

```rascal
@description{
Give a focus for the start and the end of the range and let the user figure out the rest
}
@benefits{
* Gives the implementer the most flexibility, since it carries the most information.
* Requires no additional data structures or computations required in the language server.
}
@pitfalls{
* In the case of partial selection of a syntactic list, leaves the (tricky) computation of the selected elements to the implementer.
}
alias StartEndFocus = tuple[Focus \start, Focus end];
data LanguageService = formatting(list[TextEdit](StartEndFocus focus, FormattingOptions opts) formattingService);
```

### Option2

```rascal
@description{
When the range covers element(s) of a syntactic list, separately pass all the selected list elements.
}
@benefits{
* All the selected elements are directly available as separate trees.
* No additional data structures required.
}
@pitfalls{
* Since formatting typically operates on a single tree, this variant is less useful and might require formatting each element separately and merging the results.
}
alias RangeFocus = tuple[list[Tree] elements, Focus parent];
data LanguageService = formatting(list[TextEdit](RangeFocus focus, FormattingOptions opts) formattingService);
```

### Option 3

```rascal
@description{
When the range covers element(s) of an `iter` list, this smallest element in focus is an artifial tree that is a list of those elements (but not the rest of the list)
[appl(regular(A*), elemsInRange), appl(regular(A*), [*_, *elemsInRange, *_]), ...] := focus
}
@benefits{
* Easiest to implement, since the selected elements are directly available in a single tree.
* No additional data structures required.
}
@pitfalls{
* Breaks the current `Focus` convention that the first element is the smallest tree around the cursor.
}
data LanguageService = formatting(list[TextEdit](Focus focus, FormattingOptions opts) formattingService);
```

@jurgenvinju @sungshik @PieterOlivier @rodinaarssen @DavyLandman What do you think?

PieterOlivier:
Option3 seems the most elegant solution to me. But I am not sure we can cover all formatting scenarios with this. Maybe we should also include the whole user selection as a location? That way all information is available so the language developer is not constraint in any way.

jurgenvinju:
I'm for option 3, because:
* I agree with @toinehartman analyses of the pitfalls and benefits of Option 1 and Option 2.
* Option 3 is what the Focus concept was for in the first place: making nesting explicit as a list. Since a sublist is also a form of nesting, `Focus` remains conceptually wholesome with Option 3. 
* The other options will require the language engineer to analyze the focus further in possibly complex and error prone ways, as @toinehartman wrote above.

By the way, the "knee-drop" of the simple Focus list is always the same (also for the other API calls): we limit ourselves to (lists of) abstract syntax tree nodes; this allows us to work in a syntax-directed way (based on the parse trees). Any selection range that skips a part of a tree is automatically enlarged to encompass the whole tree, as a result. What we gain from Focus (for each API call) is "syntax-directed" selection ranges that makes sense for programming languages, without having to search for trees with the right selection by the language engineer.

Finally: 
* Focus being able to produce elements which not only encompass the start of a selection range, but also the end, will improve the other API calls as well. It will give new meaning to the editor selection range that is now absent.
* Focus being able to embed sublists for the selected elements, is an orthogonal improvement of Focus, which will improve the behavior for formatting, but also for the other API calls.

sungshik:
Agree with option 3. But...

> But I am not sure we can cover all formatting scenarios with this.

I think @PieterOlivier has a point here. Imagine we have the following grammar:

```rascal
syntax StatementList = Statement | Statement ";" StatementList ;
```

Suppose there is a list of 100 statements. The selection covers statements 5 to 13. With the current proposals, I don't think (?) it is possible to know inside `formattingService` that the selection goes from 5 to 13:
  - If the focus ends at statement 5, information about statement 13 is lost.
  - If the focus ends at statement 13, then statement 5 is contained in the focus, too, but so are all the other statements up to 13, so it isn't distinguished.
  - (If the focus contains the whole tail after statement 13, then neither 5 nor 13 are known.)

I realize the example grammar above is not idiomatic, and one could make the case that (the usability of) the API is optimized for idiomatic Rascal. That said, I have no proof, but I expect idiomatic grammars to exist that have the same issue.

Intuitively, I feel `formattingService` should have (at least) as parameter the smallest parse tree that contains the begin and end of the selection. This doesn't seem to be so easy to capture as a focus, though.

rascal-lsp/src/main/rascal/library/demo/lang/pico/LanguageServer.rsc
```diff
@@ -61,9 +71,46 @@ set[LanguageService] picoLanguageServer(bool allowRecovery) = {
...
-    selectionRange(picoSelectionRangeService)
+    selectionRange(picoSelectionRangeService),
+    formatting(picoFormattingService)
 };
 
+list[TextEdit] picoFormattingService(Focus input, FormattingOptions opts) {
+    str original = "<input[-1]>";
+    box = toBox(input[-1]);
+    box = visit (box) { case i:I(_) => i[is=opts.tabSize] }
+    formatted = format(box);
```
jurgenvinju, line 82: this code would look better if the options were passed directly to `format` and it would implement them. 

For me the code drops under a certain level of abstraction (under the syntax-directed level) with all the string-level operations like `replaceLast`, `perLine` etc, while format and box2text which indeed take care of these details normally.

rascal-lsp/src/main/rascal/library/demo/lang/pico/LanguageServer.rsc
```diff
@@ -61,9 +71,46 @@ set[LanguageService] picoLanguageServer(bool allowRecovery) = {
...
     didRenameFiles(picoFileRenameService),
-    selectionRange(picoSelectionRangeService)
+    selectionRange(picoSelectionRangeService),
+    formatting(picoFormattingService)
 };
 
+list[TextEdit] picoFormattingService(Focus input, FormattingOptions opts) {
+    str original = "<input[-1]>";
+    box = toBox(input[-1]);
+    box = visit (box) { case i:I(_) => i[is=opts.tabSize] }
```
jurgenvinju, line 81: This is not a good example for users on how to parameterize the indentation level. The reason is that not all `I` boxes will correspond to the default indentation level in a typical language. Some are different. 

So indentation level should be a parameter to `toBox` where the spec writer can decide where to use it and where not.

jurgenvinju: It's good to discover this now and here. A few minor additions to `toBox` should help. We have to think hard how to make passing the parameters simple and easy, because toBox is highly recursive and we might skip accidentally to pass it on..

rascal-lsp/src/main/rascal/library/demo/lang/pico/LanguageServer.rsc
```diff
@@ -61,9 +71,46 @@ set[LanguageService] picoLanguageServer(bool allowRecovery) = {
...
     rename(picoRenamingService, prepareRenameService = picoRenamePreparingService),
     didRenameFiles(picoFileRenameService),
-    selectionRange(picoSelectionRangeService)
+    selectionRange(picoSelectionRangeService),
+    formatting(picoFormattingService)
 };
 
+list[TextEdit] picoFormattingService(Focus input, FormattingOptions opts) {
+    str original = "<input[-1]>";
+    box = toBox(input[-1]);
```
jurgenvinju, line 80: Why are we formatting the whole file?

rascal-lsp/src/main/rascal/library/util/Format.rsc
```diff
@@ -0,0 +1,157 @@
+module util::Format
```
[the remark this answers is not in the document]
jurgenvinju: If we want quick reuse, we could move these functions to Box2Text and call them in the top-level `format` function. Later some of the features could be fused into the layout algorithm to avoid revisiting the string again and again.

rascal-lsp/src/main/rascal/library/util/Format.rsc
```diff
@@ -0,0 +1,183 @@
...
+
+@synopsis{Comparator to sort strings by relative position in a reference list.}
+private bool(str, str) byIndex(list[str] indices) {
+    return bool(str a, str b) {
+        return indexOf(indices, a) < indexOf(indices, b);
+    };
+}
+
+@synopsis{Determine the most-used newline character in a string.}
+str mostUsedNewline(str input, list[str] lineseps = newLineCharacters, str(list[str]) tieBreaker = getFirstFrom) {
```
jurgenvinju, line 56: Can we hide this under the hood as one of the formatting options?

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
```diff
@@ -212,6 +212,7 @@ hover documentation, definition with uses, references to declarations, implement
    * The optional `prepareRename` service argument discovers places in the editor where a ((util::LanguageServer::rename)) is possible. If renameing the location is not supported, it should throw an exception.
 * The ((didRenameFiles)) service collects ((DocumentEdit))s corresponding to renamed files (e.g. to rename a class when the class file was renamed). The IDE applies the edits after moving the files. It might fail and report why in diagnostics.
 * The ((selectionRange)) service discovers selections around a cursor, that a user might want to select. It expects the list of source locations to be in ascending order of size (each location should be contained by the next) - similar to ((Focus)) trees.
+* The ((formatting)) service determines what edits to do to format (part of) a file. The `range` parameter determines what part of the file to format. For whole-file formatting, `_tree.top == range`. ((FormattingOptions)) influence how formatting treats whitespace.
```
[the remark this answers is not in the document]
jurgenvinju: There is no `range` parameter anymore since we have a Focus list now.

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
```diff
@@ -277,11 +278,30 @@ data LanguageService
...
+@description{
+Options that specify how to format contents of a file.
+* `insertSpaces`; if `true`, use spaces for indentation; if `false`, use tabs.
+* `tabSize`; if `insertSpaces == true`, use this amount of spaces for a single level of indentation.
+* `trimTrailingWhiteSpace`; if `true`, remove any whitespace (except newlines) from ends of formatted lines.
+* `insertFinalNewline`; if `true`, and the file does not end with a new line, add one.
+* `trimFinalNewlines`; if `true`, and the file ends in one or more new lines, remove them.
+  Note: formatting with `insertFinalNewline && trimFinalNewlines` is expected to return a file that ends in a single newline.
+}
+data FormattingOptions(
```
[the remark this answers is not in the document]
jurgenvinju: although we don't want people to write conditional code based on windows vs unix or something... should we hide this too under Box2Text functonality?

jurgenvinju (commented):
I think we should align a bit more with `toBox` and `format` (Box2Text) and map the LSP parameters to their implementations. This way we can hide the string manipulation details from the Pico formatter contrubuton.

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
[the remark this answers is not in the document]
toinehartman: This is indeed outdated now

rascal-lsp/src/main/rascal/library/util/Format.rsc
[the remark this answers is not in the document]
jurgenvinju: Box2Text handles this internally. I think it might be a good idea to remove this utility and steer users towards using the new module `util::Formatting` which hides all string manipulation under the Box abstraction.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -727,6 +734,67 @@ public CompletableFuture<List<Either<Command, CodeAction>>> codeAction(CodeActio
...
+            // convert the document changes
+            .thenApply(l -> DocumentChanges.translateTextEdits(this, l, Map.of())
+                .stream()
+                .filter(e -> Ranges.containsRange(range, e.getRange()))
+                .collect(Collectors.toList()));
+    }
+
+    private IConstructor getFormattingOptions(FormattingOptions options) {
+        var optionsType = tf.abstractDataType(typeStore, "FormattingOptions");
+        var consType = tf.constructor(typeStore, optionsType, "formattingOptions");
```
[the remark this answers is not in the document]
jurgenvinju: no these are dynamic per file and context (indentation level for example)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -727,6 +734,67 @@ public CompletableFuture<List<Either<Command, CodeAction>>> codeAction(CodeActio
...
+        // call the `formatting` implementation of the relevant language contribution
+        var fileState = getFile(uri);
+        return fileState
+            .getCurrentTreeAsync()
+            .thenApply(Versioned::get)
+            .thenCompose(tree -> {
+                // just a range
+                var start = Locations.toRascalPosition(uri, range.getStart(), columns);
+                var end = Locations.toRascalPosition(uri, range.getEnd(), columns);
+                // compute the focus list at the end of the range
```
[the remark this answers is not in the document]
jurgenvinju: yes, good catch

toinehartman:
@jurgenvinju and I made sure that all the changes necessary to format Rascal code are now available in #1102. This PR will be left draft for a while until we are ready to clean this up and resume work on DSL formatting.

PR #679, opened 2025-07-16
Clean up `ITerminalIDEServer`

### LINKED ISSUE #395: Cleanup `TerminalIDEServer` progress calls
They are now re-implementing progress bars a second time, while that interface is never called. We could clean that up and have less java code.

toinehartman:
- Remove unused job/monitor functions from `ITerminalIDEServer`.
- Move function that is only used once into the using class.

Closes #395.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IRascalFileSystemServices.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IRascalFileSystemServices.java
@@ -287,7 +287,7 @@ public interface IRascalFileSystemServices {
     }
 
     @JsonNotification("rascal/filesystem/onDidChangeFile")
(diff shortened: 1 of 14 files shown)


urbanfly:
I literally just wrote some code that uses `jobStart` and family. I typically run this code in a Rascal REPL within VS Code. Will this PR remove this capability or the "progress-bar like" functionality?

toinehartman:
@urbanfly I understand your code is Rascal, since you're running it in the REPL, correct? In that case, you're referring to functions from `util::Monitor`, which are not affected by this cleanup. This PR only removes stale, duplicate methods, that are not exposed to Rascal users.
Progress bars should remain functional - we use them everywhere as well!

rodinaarssen (approved):
Nice clean-up

PR #681, opened 2025-07-17
LSP: selection ranges for DSLs

### LINKED ISSUE #131: Expose selectionRange LSP feature
[spec](https://microsoft.github.io/language-server-protocol/specification.html#textDocument_selectionRange)

This feature helps IDE's with smarter text-selection suggestions.

toinehartman:
This endpoint, [`selectionRange`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_selectionRange) is called by at least the `Expand selection` command, which selects a larger area around the cursor at each invocation. This API receives a list of cursor locations (since there can be multiple cursors) and expects a list of hierarchies of selection locations, matched to the corresponding cursor by index.

After discussion with @sungshik, we landed on the following design.
```rascal
// module util::LanguageServer
data LanguageService
    = selectionRange(list[loc](Focus _focus) selectionRangeService)
    ;
```

This API receives a list as argument, and returns a list. In the spirit of our LSP API, we prefer to offer nice abstractions where possible. Since this is a one-to-one mapping from the input list to the output list, we can simplify the service by computing only the mapping per element. Internally, we map this computation over all cursors.

This design forces the DSL implementation to consider one cursor at a time. In theory this could leave performance on the table. In practice, the callback should be fast anyway, computation is often straightforward and the number of cursors is typically limited. On the parametric LSP side, if a single pass over the tree for every cursor is too slow, we could consider a variant of `computeFocusList` which takes a list of positions.

In line with many other services, this one takes a `Focus` as argument. Since the `Focus` includes all parse tree nodes that contain the cursor, and the full parse tree is included (`focus[-1]`), this typically leaves only a small amount of work for the DSL implementer. For example:
- Filtering which parse tree nodes to include (e.g. by production)
- Mapping trees to locations
- Deduplicating locations
- ...

If the DSL requires selection ranges that do not correspond to a parse tree node (which is allowed by the API), the implementation can artificially construct those locations.

### Alternatives considered
Providing a default implementation.
Example: map `Focus` to locations and deduplicate them (`dup([t@\loc | t <- focus])`).

Since this default is pretty small, and 'correctness' depends heavily on the structure of the grammar, it can be debated if it's needed in the first place. If so, the question is how to offer it.

#### Via overloaded constructors
```rascal
// module util::LanguageServer
data LanguageService
    = selectionRange(list[loc](Focus _focus) selectionRangeService)
    | selectionRange()
    ;

// module LanguageServerImplementation
import util::LanguageServer;
import List;

list[list[loc]] selectionRanges(list[Focus] positions, selectionRange())
    = [defaultSelectionRangeService(p) | p <- positions];

list[list[loc]] selectionRanges(list[Focus] positions, selectionRange(service))
    = [service(p) | p <- positions];

list[loc] defaultSelectionRangeService(Focus focus)
    = dup([t@\loc | t <- focus]);
```

#### Via a keyword argument
```rascal
// module util::LanguageServer
data LanguageService
    = selectionRange(list[loc](Focus _focus) selectionRangeService = list[loc](Focus focus) { return dup([t@\loc | t <- focus]); })
    ;
```

### Broader context

This API also aligns roughly with [`prepareRename`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_prepareRename), which, given a cursor, selects the range to be renamed. There might be opportunity for language-specific reuse.

Closes #131.

Related to Rascal counterpart #680.

### DIFF
--- a/rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
+++ b/rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
@@ -207,7 +207,7 @@ hover documentation, definition with uses, references to declarations, implement
 * The ((inlayHint)) service discovers plances to add "inlays" (little views embedded in the editor on the same line). Unlike ((lenses)) inlays do not offer command execution.
 * The ((execution)) service executes the commands registered by ((lenses)) and ((inlayHinter))s.
 * The ((actions)) service discovers places in the editor to add "code actions" (little hints in the margin next to where the action is relevant) and connects ((CodeAction))s to execute when the users selects the action from a menu.
-* The ((selectionRange)) service discovers selections around a cursor. It expects the list of source locations to be in ascending order of size (each location should be contained by the next) - similar to ((Focus)) trees.
+* The ((selectionRange)) service discovers selections around a cursor, that a user might want to select. It expects the list of source locations to be in ascending order of size (each location should be contained by the next) - similar to ((Focus)) trees.
 
 Many services receive a ((Focus)) parameter. The focus lists the syntactical constructs under the current cursor, from the current
 leaf all the way up to the root of the tree. This list helps to create functionality that is syntax-directed, and always relevant to the


PieterOlivier:
If we decide to use a default implementation (or provide a reference implementation), should we not make sure that the each selection returned is strictly larger than the previous one?

toinehartman:
> If we decide to use a default implementation (or provide a reference implementation), should we not make sure that the each selection returned is strictly larger than the previous one?

@PieterOlivier Yes, agreed. The current default takes the focus list (the spine of the parse tree that contains the cursor, in ascending order of size), takes their respective source locations, from which it removes duplicates. That should always result in a list of ranges that hold that property.

DavyLandman (changes requested):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -580,6 +584,29 @@ public CompletableFuture<List<FoldingRange>> foldingRange(FoldingRangeRequestPar
...
+        final TextDocumentState file = getFile(params.getTextDocument());
+
+        return recoverExceptions(file.getCurrentTreeAsync()
+                .thenApply(Versioned::get)
+                .thenCompose(t -> params.getPositions().stream()
+                    .map(p -> Locations.toRascalPosition(params.getTextDocument(), p, columns))
+                    .map(p -> TreeSearch.computeFocusList(t, p.getLine(), p.getCharacter()))
+                    .map(focus -> contrib.selectionRange(focus)
+                        .get()
+                        .thenApply(v -> v != null ? v : SelectionRanges.defaultImplementation(v)))
```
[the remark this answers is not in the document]
toinehartman: Agreed, this was a bit ugly. Changed this to branch on `hasSelectionRange`.

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
```diff
@@ -207,6 +207,7 @@ hover documentation, definition with uses, references to declarations, implement
 * The ((inlayHint)) service discovers plances to add "inlays" (little views embedded in the editor on the same line). Unlike ((lenses)) inlays do not offer command execution.
 * The ((execution)) service executes the commands registered by ((lenses)) and ((inlayHinter))s.
 * The ((actions)) service discovers places in the editor to add "code actions" (little hints in the margin next to where the action is relevant) and connects ((CodeAction))s to execute when the users selects the action from a menu.
+* The ((selectionRange)) service discovers selections around a cursor. It expects the list of source locations to be in ascending order of size (each location should be contained by the next) - similar to ((Focus)) trees.
```
DavyLandman, line 210: How about:

> service discovers selections around a cursor that a user might be interesting in selecting

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SelectionRanges.java
```diff
@@ -0,0 +1,106 @@
...
+                try {
+                    return list.get(current--);
+                } catch (IndexOutOfBoundsException e) {
+                    throw new NoSuchElementException();
+                }
+            }
+        };
+    }
+
+    public static IList defaultImplementation(IList focus) {
```
DavyLandman, line 97: have we checked how the default VS Code behaves? would it also select the full file as a possible selection?

DavyLandman (approved):
Two small clarification questions, apart from that, looks ready to merge.

PR #692, opened 2025-08-04
Add context menu to copy path as Rascal source location from explorer view

### LINKED ISSUE #106: Add "Copy source location" context menu item to VSCode explorer
cf. rascal-eclipse

toinehartman:
Closes #106


toinehartman:
Updated as per @DavyLandman's requests:
afbeelding

toinehartman:
The groups are document here:
https://code.visualstudio.com/api/references/contribution-points#Sorting-of-groups
I determined the position 21 is the first position that places this item
directly beneath `Copy Relative Path` (which probably has index 20). I
think `Copy Path` has index 10, and they left some room in between. I don't
think we get many guarantees about this order otherwise, since other
extensions could contribute menu items here as well.

PR #695, opened 2025-08-09
Feature/rename refactoring/parametric impl

PieterOlivier:
This PR implements rename refactoring from the design in PR #651

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseWorkspaceService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseWorkspaceService.java
@@ -31,14 +31,23 @@ import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutorService;
+import java.util.stream.Collectors;
 
 import com.google.gson.JsonPrimitive;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.eclipse.lsp4j.ClientCapabilities;
 import org.eclipse.lsp4j.DidChangeConfigurationParams;
 import org.eclipse.lsp4j.DidChangeWatchedFilesParams;
 import org.eclipse.lsp4j.DidChangeWorkspaceFoldersParams;
 import org.eclipse.lsp4j.ExecuteCommandParams;
+import org.eclipse.lsp4j.FileOperationFilter;
+import org.eclipse.lsp4j.FileOperationOptions;
+import org.eclipse.lsp4j.FileOperationPattern;
+import org.eclipse.lsp4j.FileOperationsServerCapabilities;
+import org.eclipse.lsp4j.RenameFilesParams;
 import org.eclipse.lsp4j.ServerCapabilities;
 import org.eclipse.lsp4j.WorkspaceFolder;
 import org.eclipse.lsp4j.WorkspaceFoldersOptions;
@@ -46,8 +55,11 @@ import org.eclipse.lsp4j.WorkspaceServerCapabilities;
 import org.eclipse.lsp4j.services.LanguageClient;
 import org.eclipse.lsp4j.services.LanguageClientAware;
 import org.eclipse.lsp4j.services.WorkspaceService;
+import org.rascalmpl.vscode.lsp.util.locations.Locations;
 
 public class BaseWorkspaceService implements WorkspaceService, LanguageClientAware {
+    private static final Logger logger = LogManager.getLogger(BaseWorkspaceService.class);
+
     public static final String RASCAL_LANGUAGE = "Rascal";
     public static final String RASCAL_META_COMMAND = "rascal-meta-command";
     public static final String RASCAL_COMMAND = "rascal-command";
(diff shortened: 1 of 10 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/DefaultWorkspaceService.java
```diff
@@ -72,13 +84,24 @@ public void initialize(ClientCapabilities clientCap, @Nullable List<WorkspaceFol
...
+                var folderOptions = new WorkspaceFoldersOptions();
+                folderOptions.setSupported(true);
+                folderOptions.setChangeNotifications(true);
+                workspaceCapabilities.setWorkspaceFolders(folderOptions);
+            }
+
+            if (clientWorkspaceCap.getFileOperations().getDidRename()) {
+                var fileOperationCapabilities = new FileOperationsServerCapabilities();
+                fileOperationCapabilities.setDidRename(new FileOperationOptions(
+                    List.of(new FileOperationFilter(new FileOperationPattern("**")))));
```
DavyLandman: this now let's rascal also get `didRename` events for all files, while it should only receive them for `.rsc` files. can we make that pattern a parameter for the  `DefaultWorkspaceService`?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseWorkspaceService.java
```diff
@@ -72,13 +84,24 @@ public void initialize(ClientCapabilities clientCap, @Nullable List<WorkspaceFol
...
+                var folderOptions = new WorkspaceFoldersOptions();
+                folderOptions.setSupported(true);
+                folderOptions.setChangeNotifications(true);
+                workspaceCapabilities.setWorkspaceFolders(folderOptions);
+            }
+
+            if (clientWorkspaceCap.getFileOperations().getDidRename()) {
+                var fileOperationCapabilities = new FileOperationsServerCapabilities();
+                fileOperationCapabilities.setDidRename(new FileOperationOptions(
+                    List.of(new FileOperationFilter(new FileOperationPattern("**")))));
```
DavyLandman, line 99: This is an example of the difference between rascal and parametric. We don't want rascal to receive events for all renames, only for the `**/*.rsc` changes.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseWorkspaceService.java
```diff
@@ -112,6 +135,17 @@ public void didChangeWorkspaceFolders(DidChangeWorkspaceFoldersParams params) {
...
 
+    @Override
+    public void didRenameFiles(RenameFilesParams params) {
+        logger.debug("workspace/didRenameFiles: {}", params.getFiles());
+
+        CompletableFuture.supplyAsync(() -> workspaceFolders()
+                .stream()
+                .map(f -> Locations.toLoc(f.getUri()))
+                .collect(Collectors.toSet()), getExecuter())
+                .thenAccept(folders -> documentService.didRenameFiles(params, folders));
```
DavyLandman, line 146: nit: the translation of `workspaceFolders` to a `Set<ISourceLocation>` can also be done in the same block as the `documentService.didRenameFiles`. And I would even say that just passing in the result of `workspaceFolders` to the `didRenameFiles` function would be my prefered solution, so we don't do conversion when we don't even need it (and it's all just lsp4j classes anyway).

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/InterpretedLanguageContributions.java
```diff
@@ -256,19 +267,35 @@ private CompletableFuture<IConstructor> parseCommand(String command) {
...
             logger.debug("No {} defined", cons);
             return null;
         });
     }
 
+    private static CompletableFuture<@Nullable IFunction> getFunctionFor(CompletableFuture<ISet> contributions, String cons) {
+        return getContribution(contributions, cons).thenApply(contribution -> (IFunction) contribution.get(0));
+    }
+
+    private static CompletableFuture<@Nullable IFunction> getRenameFunction(CompletableFuture<ISet> contributions) {
```
DavyLandman, line 287: why is this the only function that gets a special overload? and the other use `getFunctionFor`? should this maybe be an overload called `getParameterizedFunction`? or something like this? (the discrepancy is more clear in the constructor)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -312,6 +345,148 @@ public CompletableFuture<List<? extends CodeLens>> codeLens(CodeLensParams param
...
+                .get()
+                .exceptionally(e -> {
+                    logger.catching(Level.ERROR, e.getCause());
+                    client.showMessage(new MessageParams(MessageType.Error, e.getCause().getMessage()));
+                    return null; // Return of type `Void` is unused, but required
+                });
+        }
+    }
+
+    private Map<ILanguageContributions, List<FileRename>> bundleRenamesByContribution(List<FileRename> allRenames) {
```
DavyLandman, line 475: since `prepareRename` is called by the workspace service for all file changes, we should first make sure, the current incoming call is actually for a file extension for which we have any handlers registered.

I think this function should filter out files we do not even care about. 

And this is only since we cannot predict, for the multiplexer, which file patterns would be more precise than all of the files.

DavyLandman (changes requested):
Looking good, some small changes.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/DefaultWorkspaceService.java
PieterOlivier: This code has moved back to `RascalWorkspaceService`. If we want to improve on this it should be done in a separate PR.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseWorkspaceService.java
PieterOlivier: I agree, I have changed the implementation to use `List<WorkspaceFolder>` instead.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/InterpretedLanguageContributions.java
PieterOlivier: I have now unified the "getFunctionFor" approach.

rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
```diff
@@ -211,4 +211,59 @@ parameterizedDescribe(function (errorRecovery: boolean) {
...
     });
 
+    it("rename works", async function() {
+        if (errorRecovery) { this.skip(); }
+        const editor = await ide.openModule(TestWorkspace.picoFile);
+        await editor.moveCursor(5, 6);
+
+        let renameSuccess = false;
+        let tries = 0;
+        while (!renameSuccess && tries < 5) {
```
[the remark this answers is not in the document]
PieterOlivier: Done.

DavyLandman:
This needs to be updated wrt main, since the `old` style language server is gone now.

PieterOlivier:
> This needs to be updated wrt main, since the `old` style language server is gone now.

Yes and because of a bucketload of (trivial) merge conflicts needed to be resolved because the selection range stuff has been merged into main.

PR #700, opened 2025-08-15
Migrate DAP and consume updates in Rascal

rodinaarssen:
This PR makes `rascal-lsp` dependent on rascal 0.41.0-RC55.
* DAP has moved to Rascal
* Watch-related changes in Rascal have been adopted

After merging this, the extension will not be (fully) functional; a future release of Rascal containing https://github.com/usethesource/rascal/pull/2351 will fix that.


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/FallbackResolver.java
```diff
@@ -319,4 +324,24 @@ public TextDocumentState getDocumentState(ISourceLocation file) throws IOExcepti
...
+
+    @Override
+    public FileAttributes stat(ISourceLocation uri) throws IOException {
+        throw new UnsupportedOperationException("`stat` not supported by fallback resolver");
+    }
+
+    @Override
+    public boolean isWritable(ISourceLocation uri) throws IOException {
+        throw new UnsupportedOperationException("`isWritable` not supported by fallback resolver");
+    }
```
DavyLandman, line 346: Shouldn't we try to implement these?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileResolver.java
```diff
@@ -100,5 +101,20 @@ public String scheme() {
...
+
+    @Override
+    public boolean isReadable(ISourceLocation uri) throws IOException {
+        throw new UnsupportedOperationException("`isReadable` not supported by LSPOpenFileResolver");
+    }
+
+    @Override
+    public FileAttributes stat(ISourceLocation uri) throws IOException {
+        throw new UnsupportedOperationException("`stat` not supported by LSPOpenFileResolver");
+    }
```
DavyLandman, line 118: These should be similarly forwarded/calculated if we can.

DavyLandman (commented):
I think 2 parts are still missing an implementation?

rascal-vscode-extension/src/fs/VSCodeURIResolver.ts
```diff
@@ -368,6 +383,41 @@ class ResolverClient implements VSCodeResolverServer, Disposable  {
...
+        const fileInfo = await this.stat(req);
+        return {
+            errorCode: 0,
+            exists: (await this.exists(req)).result!,
+            type: fileInfo.type.valueOf(),
+            ctime: fileInfo.ctime,
+            mtime: fileInfo.mtime,
+            size: fileInfo.size,
+            permissions: fileInfo.permissions ? fileInfo.permissions.valueOf() : 0,
+            isWritable: (await this.isWritable(req)).result!
```
DavyLandman, line 397: shouldn't  `isReadable` also be implemented?

rodinaarssen: Good catch

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/FallbackResolver.java
```diff
@@ -265,16 +266,20 @@ public void setLastModified(ISourceLocation uri, long timestamp) throws IOExcept
...
     @Override
-    public void unwatch(ISourceLocation root, Consumer<ISourceLocationChanged> watcher) throws IOException {
-        getClient().removeWatcher(root, watcher, getServer());
+    public void unwatch(ISourceLocation root, Consumer<ISourceLocationChanged> watcher, boolean recursive) throws IOException {
+        getClient().removeWatcher(root, recursive, watcher, getServer());
+    }
 
+    @Override
+    public boolean supportsRecursiveWatch() {
+        return false;
```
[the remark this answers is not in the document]
rodinaarssen: You're right, VS Code does

PR #702, opened 2025-08-19
Made error highlighting configurable through environment variables

PieterOlivier:
You can use the following environment variables to configure highlighting:
 - RASCAL_EDITOR_ERROR_LOCATION_HIGHLIGHT
 - RASCAL_EDITOR_ERROR_WHOLE_TREE_HIGHLIGHT
 - RASCAL_EDITOR_ERROR_PREFIX_HIGHLIGHT
 - RASCAL_EDITOR_ERROR_PREFIX_START_HIGHLIGHT
 - RASCAL_EDITOR_ERROR_SKIPPED_HIGHLIGHT
 
 Each of these variables can be set to one of the following values:
 - none
 - error
 - warning
 - info
 - hint (only highlights a single character)

If you do not set a variable, its default will be used (`error` for the location highlight, `none` for the others).

For instance with these variables set:
image

Highlighting of a recovered parse error in Pico looks like this:
image

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
@@ -60,11 +60,74 @@ public class Diagnostics {
     private static final Logger logger = LogManager.getLogger(Diagnostics.class);
     private static final Map<String, DiagnosticSeverity> severityMap;
 
-    // Note: DiagnosticSeverity.Hint only highlightes a single character!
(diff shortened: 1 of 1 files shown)


DavyLandman:
This feels very much like a VS Code setting (which we discussed as the path to offer this, if we ever wanted to offer this). Especially setting custom environment variables globally is quite annoying.

So that would mean using `workspace/configuration` or the `DidChangeConfigurationNotification` event. As long as we mark these settings as experimental, it should be fine.

Note: don't spend too much time on this, if it's to painful to hook it up to the settings, just leave them as environment settings, since we don't want to expose this to the end user anyway.

DavyLandman:
I think #712 has replaced this idea?

PR #703, opened 2025-08-20
Rich VS Code logs

### LINKED ISSUE #587: Connect log4j2 to VS Code LogOutputChannel
Right now we're connected to a regular OutputChannel, but if we can become a [`LogOutputChannel`](https://code.visualstudio.com/api/references/vscode-api#LogOutputChannel) we can have users select the level of output from VS Code, and many more nice features for our logging.

There might be some intermediate framework we'll have to work around, or perhaps build a small log4j2 parser in typescript that translates our output to `LogOutputChannel` calls.

h/t to @urbanfly for pointing out the API.

toinehartman:
afbeelding

Closes #587


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/LogJsonConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+
+        ConfigurationBuilder<BuiltConfiguration> builder = ConfigurationBuilderFactory.newConfigurationBuilder();
+        builder.setConfigurationName("JsonLogger");
+        builder.setStatusLevel(targetLevel);
+
+        builder.add(builder
+            .newAppender("Console", ConsoleAppender.PLUGIN_NAME)
+            .addAttribute("target", ConsoleAppender.Target.SYSTEM_ERR)
+            .add(builder.newLayout("JsonTemplateLayout")
+                .addAttribute("maxStringLength", 4096) // never truncate JSON (which has total max size of 8192)
```
DavyLandman, line 80: I don't understand the comment? Because we do truncate at 4096? or does it mean the message gets split up? and what does that mean for a long line that is printed?

rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -38,14 +38,16 @@ import { RascalLibraryProvider } from './ux/LibraryNavigator';
 import { FileType } from 'vscode';
 import { RascalDebugViewProvider } from './dap/RascalDebugView';
 
+export const extensionLog: vscode.LogOutputChannel = vscode.window.createOutputChannel("Rascal Extension", {log: true});
```
DavyLandman, line 41: I don't think it should be this global, it should be owned by the `RascalExtension` class and disposed along side of it. Or maybe we need to have this at the root of the extension? and pass it in as a parameter to the constructor?

Roughly: I don't want our extension to register output channels if we're not active. And I want to be able to survive package updates without dangling windows etc.

rascal-vscode-extension/.eslintrc.json
```diff
@@ -28,6 +28,7 @@
         "eqeqeq": "warn",
         "no-throw-literal": "warn",
         "semi": "off",
-        "no-unused-vars": "off"
+        "no-unused-vars": "off",
+        "no-console": "error"
```
DavyLandman, line 32: nice one. :+1:

rascal-vscode-extension/src/lsp/JsonOutputChannel.ts
```diff
@@ -0,0 +1,190 @@
...
+class LogMessage {
+    constructor(
+        public readonly timestamp: Date,
+        public readonly level: LogLevel,
+        public readonly message: string,
+        public readonly threadName: string,
+        public readonly loggerName: string) {}
+
+    static cast(json: object): LogMessage | undefined {
+        if("timestamp" in json && typeof json.timestamp === "string"
```
DavyLandman, line 47: Maybe this could be nicer formulated as a [type guard](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)?

rascal-vscode-extension/src/lsp/JsonOutputChannel.ts
```diff
@@ -0,0 +1,190 @@
...
+    private formatServerTime(date: Date | string): string {
+        if (typeof date === "string") {
+            date = new Date(date);
+        }
+        return date.toISOString();
+    }
+
+    private formatMessage(thread: string, originalTime: Date | string, message: string, loggerName?: string) {
+        const loggerPart = loggerName ? ` ${loggerName}` : "";
+        return `[${thread}] (${this.formatServerTime(originalTime)})${loggerPart} ${message}`;
```
DavyLandman, line 131: it's a bit weird to print both times in a way right? could we for ourselves find a small way to print this? say epoch time?

rascal-vscode-extension/src/lsp/JsonOutputChannel.ts
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    private printNonJsonPayload(payload: string): void {
```
DavyLandman, line 134: why are we getting non json output on the stream? stuff from rascal? and could we maybe just not try to parse them using a regex? or are they that common?

rascal-vscode-extension/src/lsp/JsonOutputChannel.ts
```diff
@@ -0,0 +1,190 @@
...
+
+    append(payload: string): void {
+        this.printPayload(payload);
+    }
+
+    appendLine(payload: string): void {
+        this.printPayload(payload);
+    }
+
+    show(preserveFocus?: unknown): void {
```
DavyLandman, line 171: what if we pass a string? nothing happens? this seems a bit of a weird type? shouldn't it be `show(preserverFocus: boolean | undefined)`?

DavyLandman (commented):
Looks good, I have some initial thoughts, I want to run this for a few minutes and maybe share some extra thoughts.

rascal-vscode-extension/src/lsp/JsonOutputChannel.ts
DavyLandman: you can have multiple signatures for a function I think, i forgot the exact syntax.

```typescript
show(preserveFocus?: boolean): void
show(column?: ViewColumn, preserveFocus?: boolean): void {
}
```

but I agree, kinda weird these 2 overloads.

rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -43,9 +43,10 @@ export class RascalExtension implements vscode.Disposable {
     private readonly dsls:ParameterizedLanguageServer;
     private readonly rascal: RascalLanguageServer;
 
+    private readonly log: vscode.LogOutputChannel = vscode.window.createOutputChannel("Rascal Extension", {log: true});
```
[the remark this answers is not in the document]
toinehartman: The log that's created here is for the extension (non-language things). Logs for the Rascal and parametric servers are created here: https://github.com/usethesource/rascal-language-servers/blob/4c7171d5a5132ac33a7483408dcae3ec3efc5516/rascal-vscode-extension/src/lsp/RascalLSPConnection.ts#L42.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/LogJsonConfiguration.java
```diff
@@ -0,0 +1,97 @@
...
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.rascalmpl.vscode.lsp;
```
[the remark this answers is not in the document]
toinehartman: Sure. I can move it into a `log` subpackage, along with the existing log config.

rascal-vscode-extension/src/lsp/JsonOutputChannel.ts
```diff
@@ -0,0 +1,159 @@
...
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+import * as vscode from 'vscode';
+
+enum LogLevel {
```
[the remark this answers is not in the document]
toinehartman: Yes, will do!

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -122,11 +123,11 @@ interface BrowseParameter {
...
 }
 
-async function buildRascalServerOptions(jarPath: string, isParametricServer: boolean, dedicated: boolean, lspArg : string | undefined): Promise<ServerOptions> {
+async function buildRascalServerOptions(jarPath: string, isParametricServer: boolean, dedicated: boolean, lspArg: string | undefined, logger: vscode.LogOutputChannel): Promise<ServerOptions> {
     const classpath = buildCompilerJVMPath(jarPath);
     const commandArgs = [
-        '-Dlog4j2.configurationFactory=org.rascalmpl.vscode.lsp.LogRedirectConfiguration'
-        , '-Dlog4j2.level=DEBUG'
+        '-Dlog4j2.configurationFactory=org.rascalmpl.vscode.lsp.LogJsonConfiguration'
+        , '-Dlog4j2.level=DEBUG' // TODO Remove this minimum
```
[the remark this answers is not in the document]
toinehartman: Indeed. I think we should remove any hard-coded minimum and instead set a default minimum level for our log channel (probably `INFO`).

rascal-vscode-extension/src/RascalExtension.ts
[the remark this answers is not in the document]
toinehartman: Correct. I think those ended up somewhere else before, but it's nice to be able to also have those in a proper log channel.

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -122,11 +123,11 @@ interface BrowseParameter {
...
 }
 
-async function buildRascalServerOptions(jarPath: string, isParametricServer: boolean, dedicated: boolean, lspArg : string | undefined): Promise<ServerOptions> {
+async function buildRascalServerOptions(jarPath: string, isParametricServer: boolean, dedicated: boolean, lspArg: string | undefined, logger: vscode.LogOutputChannel): Promise<ServerOptions> {
     const classpath = buildCompilerJVMPath(jarPath);
     const commandArgs = [
-        '-Dlog4j2.configurationFactory=org.rascalmpl.vscode.lsp.LogRedirectConfiguration'
-        , '-Dlog4j2.level=DEBUG'
+        '-Dlog4j2.configurationFactory=org.rascalmpl.vscode.lsp.log.LogJsonConfiguration'
+        , '-Dlog4j2.level=TRACE'
```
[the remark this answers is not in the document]
toinehartman: https://github.com/usethesource/rascal-language-servers/issues/709

DavyLandman (approved):

PR #705, opened 2025-08-21
Cleanup old diagnostics for deleted files

### LINKED ISSUE #671: error diagnostics about a removed or renamed file are not cleaned up, and can't be removed manually
What the subject says. I renamed a Rascal module but the error about the fact that its module name did not align with its filename remains in the Problems View.

DavyLandman:
Fixes #671

I've also reduced the watch scope of rascal to only monitor rascal files, not all files. @toinehartman do you remember why initially it was for all files? is it for a rename of a file from a non `.rsc` extension?


toinehartman:
> Wouldn't the watcher trigger anyway on the new extension?

From my memory, it didn't

DavyLandman:
Okay, we've had a discussion, to respect VS Code wishes to not register to heavy of a watch, we're not going to do it. and accept the edge case that if a user started with a wrong file extension (but did try to import it) and then not only changed the file extension but also the name we do not trigger the rename code for that specific edge case.

PR #706, opened 2025-08-22
Parametric completion

PieterOlivier:
Design & implementation of [completion](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion) support for the parametric language server.

This PR also includes support for dynamically registering capabilities.


rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
```diff
@@ -279,6 +279,109 @@ data LanguageService
...
+    list[TextEdit] additionalChanges = []
+);
+
+@description{
+    Definition of a completion edit:
+    * *startColumn (required): The column where the completion edit operation will take place. Must be at or before the cursor position.
+    * *insertEndColumn* (required): End column when the user choose completion by insertion  (for instance by pressing "Enter" in VSCode).
+    * *replaceEndColumn* (required): End column when the user choose completion by replacement (for instance by pressing "Shift-Enter" in VSCode).
+        Note: *insertEndColumn* must not be larger than *replaceEndColumn* and both must be at or to the right of the cursor position.
+    * *newText* *required*: The text that will be used to perform the completion. Depending on what kind of completion (insertion
```
toinehartman, line 348:
```suggestion
    * *newText* (required): The text that will be used to perform the completion. Depending on what kind of completion (insertion
```

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
```diff
@@ -279,6 +279,109 @@ data LanguageService
...
+
+    list[str] commitCharacters = [],
+
+    list[TextEdit] additionalChanges = []
+);
+
+@description{
+    Definition of a completion edit:
+    * *startColumn (required): The column where the completion edit operation will take place. Must be at or before the cursor position.
+    * *insertEndColumn* (required): End column when the user choose completion by insertion  (for instance by pressing "Enter" in VSCode).
```
toinehartman, line 345:
```suggestion
    * *insertEndColumn* (required): End column when the user chooses completion by insertion  (for instance by pressing "Enter" in VSCode).
```

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
```diff
@@ -279,6 +279,109 @@ data LanguageService
...
+    * *replaceEndColumn* (required): End column when the user choose completion by replacement (for instance by pressing "Shift-Enter" in VSCode).
```
toinehartman, line 346:
```suggestion
    * *replaceEndColumn* (required): End column when the user chooses completion by replacement (for instance by pressing "Shift-Enter" in VSCode).
```

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
```diff
@@ -279,6 +279,109 @@ data LanguageService
...
+    * *snippet*: Can be set to true to indicate that the replacement text should be interpreted as a snippet.
+        Snippets can contain tabstops, placeholders, choices, and variables. For more information about snippets see the
+        [relevant section of the lsp specification on github](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#snippet_syntax).
+        Note that it is the responsibility of the language developer that snippets as presented to the user are syntactically correct to
+        provide a decent user experience.
+}
+data CompletionEdit = completionEdit(
+    int startColumn,
+    int insertEndColumn,
+    int replaceEndColumn,
```
toinehartman, line 360: Can we not do multi-line completions?

toinehartman: Answered by the [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem):
> *Note:* The range of the edit must be a single line range and it must contain the position at which completion has been requested.

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
```diff
@@ -279,6 +279,109 @@ data LanguageService
...
+    * *insertEndColumn* (required): End column when the user choose completion by insertion  (for instance by pressing "Enter" in VSCode).
+    * *replaceEndColumn* (required): End column when the user choose completion by replacement (for instance by pressing "Shift-Enter" in VSCode).
+        Note: *insertEndColumn* must not be larger than *replaceEndColumn* and both must be at or to the right of the cursor position.
+    * *newText* *required*: The text that will be used to perform the completion. Depending on what kind of completion (insertion
+        or replacement) selected by the user, the original text from *startColumn* to either *insertEndColumn* or *replaceEndColumn* will
+        be replaced by *newText*.
+    * *snippet*: Can be set to true to indicate that the replacement text should be interpreted as a snippet.
+        Snippets can contain tabstops, placeholders, choices, and variables. For more information about snippets see the
+        [relevant section of the lsp specification on github](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#snippet_syntax).
```
toinehartman, line 353:
```suggestion
        [LSP specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#snippet_syntax).
```

toinehartman (changes requested):
I am excited to see this in action! Since we already discussed this design together, my comments mostly consider documentation and naming.

DavyLandman (commented):
It's looking good, it's not a small API, but it's looking good.

rascal-lsp/src/main/rascal/library/demo/lang/pico/LanguageServer.rsc
```diff
@@ -227,6 +230,30 @@ tuple[list[DocumentEdit],set[Message]] picoFileRenameService(list[DocumentEdit]
...
+
+    return completion(variable(), edit, name, labelDetail="<decl.t>");
+}
+
+list[CompletionSuggestion] picoCompletionService(loc cursor, Focus focus, CompletionTrigger trigger) {
+    str prefix = "<focus[0]>";
+    return [createVarCompletion(cursor.begin.column, focus[0].src, var) | /IdType var := focus[-1], startsWith("<var.id>", "<prefix>")];
+}
+
+void testCompletion() {
```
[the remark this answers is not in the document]
PieterOlivier: This is just temporary for me to do some testing. My intention is to remove this when we actually implement this design.

rascal-lsp/src/main/rascal/library/demo/lang/pico/LanguageServer.rsc
```diff
@@ -61,7 +63,8 @@ set[LanguageService] picoLanguageServer(bool allowRecovery) = {
     codeAction(picoCodeActionService),
     rename(picoRenamingService, prepareRenameService = picoRenamePreparingService),
     didRenameFiles(picoFileRenameService),
-    selectionRange(picoSelectionRangeService)
+    selectionRange(picoSelectionRangeService),
+    completion(picoCompletionService, triggerCharacters = ["="])
```
toinehartman, line 67:
```suggestion
    completion(picoCompletionService, additionalTriggerCharacters = ["="])
```

jurgenvinju:
It would perhaps be good to test drive this contribution with a generic function that takes a reified grammar and produces a simplified syntax directed completion contribution function (based on error trees or normal trees). To be used as `completion(syntaxDirectedCompletion(#start[Program]))` for example for Pico?

That way we can see if the LSP side fits on the Rascal side with what we have in mind. Or if we still need to think more about the Rascal side.

DavyLandman:
I've discussed the design choice of merging the 2 LSP enums into one ADT with @jurgenvinju and he agrees that in the end it's the cleanest solution. It's not 100% ideal, but the alternatives are worse for the end users.

He indicated he would take a final look at the rest of the API today, as he had some concerns and wanted to verify how they were addressed.

jurgenvinju (commented):
Love it

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
```diff
@@ -279,6 +279,114 @@ data LanguageService
...
+
+We have choosen to support all features of the LSP CompletionItem except:
+* We use "DocumentSymbolKind" instead of introducing a new enum to represent CompletionKind. This is only used to show a tiny icon next to each completion alternative, and the few concept that differ can usually be mapped to some related or more general concept.
+* To keep this API simple, we have left out support for incomplete (partial) completions, so "CompletionList.isIncomplete" will always be set to false.
+* Again to keep the API simple we have not implemented support for defaults, so CompletionItem, edit range (and commitCharacters if you want them) must be set explicitly on each CompletionItem.
+
+Note: Depending on the capabilities of the client, we will generate "InsertReplaceEdit" items or "TextEdit" items.
+}
+data LanguageService
+    = completion    (list[CompletionSuggestion] (loc cursor, Focus _focus, CompletionTrigger trigger) completionService, list[str] additionalTriggerCharacters = []);
```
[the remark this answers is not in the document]
PieterOlivier: We need to point inside the lexical. I have adapted the design to use an offset inside the lexical instead of a complete loc. That is hopefully more clear.

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
```diff
@@ -279,6 +279,114 @@ data LanguageService
...
+    str newText,
+    bool snippet = false
+);
+
+@description{
+    The lsp *CompletionItemKind* and our ((DocumentSymbolKind)) describe largely the same concepts. As these are only used (in VSCode)
+    to display a tiny icon next to each completion item, the small differences remaining can be overcome by selecting a related
+    or more general concept.
+
+    *CompletionItemKind* but not a ((DocumentSymbolKind)): `Text`, `Unit`, `Value`, `Keyword`, `Snippet`, `Color`, `Reference`, `Folder`, `EnumMember`.
```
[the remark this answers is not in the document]
PieterOlivier: If we needed all the entries from `DocumentSymbolKind` that would be a good approach. But the code would require a separate list of what to include so there is imho not much value in generating it.

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
```diff
@@ -536,14 +646,49 @@ data DocumentSymbolKind
 	| \event()
 	| \operator()
 	| \typeParameter()
+    | \text()
+    | \unit()
+    | \value()
+    | \keyword()
+    | \snippet()
+    | \color()
```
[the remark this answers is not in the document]
PieterOlivier: I have not tried it but I don't think so. For completionis these types are solely used to display an icon before each completion as a hint to the user what the completion is about.

toinehartman (commented):
Commenting to get rid of my own review of the design.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/DynamicCapabilities.java
```diff
@@ -0,0 +1,237 @@
...
+                        }
+                        registrations.add(registration);
+                    }
+
+                    try {
+                        doRegistrations(registrations, unregistrations);
+                    } catch (InterruptedException e) {
+                        Thread.currentThread().interrupt();
+                    } catch (ExecutionException e) {
+                        logger.error("Doing registrations failed!", e);
```
PieterOlivier: It might be helpful to log which registration failed.

PieterOlivier (commented):
I think the dynamic contribution design is fine like this. When we gain more experience with different contributions we might find some improvements.

toinehartman:
It's probably better to redo the last commit on a separate branch, since it affects a lot of non-completion code and should be prioritised for the upcoming release.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/AbstractDynamicCapability.java
```diff
@@ -0,0 +1,112 @@
...
+ * Abstract superclass for implementations of dynamic capabilities.
+ * @param O The type of the capability's options.
+ */
+public abstract class AbstractDynamicCapability<O> {
+
+    private final String id;
+
+    protected AbstractDynamicCapability() {
+        id = UUID.randomUUID().toString();
+    }
```
DavyLandman: can we add some documentation/comment of why this requires a random string?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/AbstractDynamicCapability.java
```diff
@@ -0,0 +1,112 @@
...
+
+    protected abstract String methodName();
+
+    /**
+     * Computes the options this capability given language contributions.
+     * @param contribs The {@link ILanguageContributions} that this capability reflects.
+     * @return A future resolving to options
+     */
+    protected abstract CompletableFuture<@Nullable O> options(ILanguageContributions contribs);
```
DavyLandman: in the rest of this code we use `hasContribution` to check if you should call any option on there. That avoids the nullable return value, but makes sure we can avoid exceptions getting thrown. would that fit here?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/AbstractDynamicCapability.java
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * Checks whether the givien language contributions contain a contribution that matches this capability.
+     * @param contribs The {@link ILanguageContributions} that this capability reflects.
+     * @return A future resolving to `true` if there is such a contribution, or `false` otherwise.
+     */
+    protected abstract CompletableFuture<Boolean> hasContribution(ILanguageContributions contribs);
```
DavyLandman: the naming of this method confuses me, maybe it should be something like `matchesContribution`? as this capability has no contribution by itself?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/AbstractDynamicCapability.java
```diff
@@ -0,0 +1,112 @@
...
+
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.eclipse.lsp4j.ClientCapabilities;
+import org.eclipse.lsp4j.ServerCapabilities;
+import org.rascalmpl.vscode.lsp.parametric.ILanguageContributions;
+
+/**
+ * Abstract superclass for implementations of dynamic capabilities.
```
DavyLandman: This sentence is a wordy version of the information in `public abstract class AbstractDynamicCapability` as in, I don't learn anything new from this. Can we make this into a description that explains why this class exists/what purpose dynamic capabilities have? why we have a special class for this?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/CompletionCapability.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    @Override
+    protected @Nullable CompletionRegistrationOptions mergeOptions(@Nullable CompletionRegistrationOptions left, @Nullable CompletionRegistrationOptions right) {
+        if (left == null) {
+            return right;
+        }
+        if (right == null) {
+            return left;
+        }
```
DavyLandman: this looks like something the base/caller class should take care of, such that we're only called for the case where both are not null?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/DynamicCapabilities.java
```diff
@@ -0,0 +1,205 @@
...
+    private final Map<String, Registration> currentRegistrations = new ConcurrentHashMap<>();
+
+    private @MonotonicNonNull ClientCapabilities clientCapabilities;
+    private @MonotonicNonNull ServerCapabilities serverCapabilities;
+
+    public DynamicCapabilities(LanguageClient client, Executor exec, List<AbstractDynamicCapability<?>> supportedCapabilities) {
+        this.client = client;
+        this.exec = exec;
+        this.supportedCapabilities = supportedCapabilities;
+        this.staticCapabilities = new HashSet<>();
```
DavyLandman: not a thread safe container.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/DynamicCapabilities.java
```diff
@@ -0,0 +1,205 @@
...
+    private final LanguageClient client;
+    private final Executor exec;
+    private final List<AbstractDynamicCapability<?>> supportedCapabilities;
+    private final Set<AbstractDynamicCapability<?>> staticCapabilities;
+    private final Map<String, Registration> currentRegistrations = new ConcurrentHashMap<>();
+
+    private @MonotonicNonNull ClientCapabilities clientCapabilities;
+    private @MonotonicNonNull ServerCapabilities serverCapabilities;
+
+    public DynamicCapabilities(LanguageClient client, Executor exec, List<AbstractDynamicCapability<?>> supportedCapabilities) {
```
DavyLandman: Do we take ownership of this list? what if someone else outside of our class manipulates the list? is it a thread safe list?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/DynamicCapabilities.java
```diff
@@ -0,0 +1,205 @@
...
+        this.supportedCapabilities = supportedCapabilities;
+        this.staticCapabilities = new HashSet<>();
+    }
+
+    /**
+     * Determines whether the server should register capabiltities statically now instead of dynamically later.
+     * @param clientCapabilities The capabilities of the client, which determine whether dynamic registration is supported at all.
+     * @param serverCapabilities The server capabilities to modify.
+     */
+    public void setStaticCapabilities(ClientCapabilities clientCapabilities, final ServerCapabilities serverCapabilities) {
```
DavyLandman: I think the name is not correct, as it also sets the client dynamic capabilities?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/DynamicCapabilities.java
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        if (!registrations.isEmpty()) {
+            logger.debug("Registering dynamic capabilities: {}", registrations);
+            client.registerCapability(new RegistrationParams(registrations)).get();
+            registrations.forEach(r -> currentRegistrations.put(r.getMethod(), r));
+        }
+    }
+
+    private <T> CompletableFuture<Pair<AbstractDynamicCapability<T>, @Nullable Registration>> maybeRegistration(AbstractDynamicCapability<T> cap, Collection<ILanguageContributions> contribs) {
```
DavyLandman: I think the name is a bit mislaying I assumed it would check if they "may be  a registration" but it's more `maybeRegister` or `tryRegister`?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/DynamicCapabilities.java
```diff
@@ -0,0 +1,205 @@
...
+     */
+    CompletableFuture<Void> updateCapabilities(Collection<ILanguageContributions> contribs) {
+        // Compute registrations purely based on contributions
+        // This requires waiting for an evaluator to load, which might take long, and should not block our logbook
+        return CompletableFutureUtils.reduce(supportedCapabilities.stream().filter(cap -> !staticCapabilities.contains(cap)).map(c -> maybeRegistration(c, contribs)), exec)
+            .thenAccept(capabilities -> {
+                // Since we have some bookkeeping to do, we will now block our logbook for a moment
+                synchronized (currentRegistrations) {
+                    List<Registration> registrations = new LinkedList<>();
+                    List<Unregistration> unregistrations = new LinkedList<>();
```
DavyLandman: I'm not convinced that this code will not cause a race. The combination of all these futures with a lock on the `currentRegistrations` smells like deadlock waiting to happen. For example with `doRegistrations` that locks on the `this`.

I think it would work better if we change the code to first build the `registration` and `unregistration` work lists. and then manipulate the `currentRegistrations` based on that, without a lock, but with the remove/puts that validate the current state in the map. and based on that get a queue of things to actually send to the server.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/InterpretedLanguageContributions.java
```diff
@@ -237,8 +243,24 @@ private static boolean hasContribution(ISet contributions, String name) {
...
+                return defaultVal;
+            }
 
-        return contributions.thenApply(c -> isTrue(getContribution(c, name), parameter));
+            var val = contrib.asWithKeywordParameters().getParameter(parameter);
+            try {
+                return (T) val;
+            } catch (ClassCastException e) {
+                return defaultVal;
+            }
```
DavyLandman: Are we sure that due to type-erasure it doesn't fail if you wanted an `IList` but got an `IString` since `T` at runtime will be `Object`? So this will only fail with a `null` value, but that might not even throw a `ClassCastException`?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -288,6 +305,7 @@ private LanguageClient availableClient() {
     @Override
     public void connect(LanguageClient client) {
         this.client = client;
+        this.dynamicCapabilities = new DynamicCapabilities(client, exec, List.of(new CompletionCapability()));
```
DavyLandman: shouldn't we look if somehow we already received some registrations?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -959,6 +997,14 @@ public synchronized void registerLanguage(LanguageParameter lang) {
         multiplexer.addContributor(buildContributionKey(lang),
             new InterpretedLanguageContributions(lang, this, availableWorkspaceService(), (IBaseLanguageClient)clientCopy, exec));
 
+        try {
+            availableCapabilities().updateCapabilities(contributions).get();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        } catch (ExecutionException e) {
+            logger.error("Unexpected error while updating dynamic capabilities", e);
+        }
```
DavyLandman: This reads like we're joining on a future, why are we doing this?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Completion.java
```diff
@@ -0,0 +1,204 @@
...
+                return character.apply(VF.string(triggerChar));
+            }
+            default: throw new IllegalArgumentException("Unsupported completion trigger kind: " + context.getTriggerKind());
+        }
+    }
+
+    public static CompletableFuture<Boolean> isTriggered(IConstructor kind, CompletableFuture<IList> triggerChars) {
+        if (CompletionFields.INVOKED.equals(kind.getName())) {
+            // Manual invocation always triggers completion
+            return CompletableFuture.completedFuture(true);
```
DavyLandman: this should have an executor attached to it.

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
```diff
@@ -549,6 +650,9 @@ data DocumentSymbol
         set[DocumentSymbolTag] tags = {}
     );
 
+@description{
+    List of types that can be used for DocumentSymbol and CompletionItem.
```
DavyLandman: let's repeat the warning that some are specific to one and some are specific to the other.

DavyLandman (changes requested):
This is quite a big chunk of work. I have some small points that I pointed out, a bug in the multiplexer and some concerns about the thread-safety (and tight coupling) of the `DynamicCapabilities` class.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
DavyLandman: ah yes, connect really should have happened first, as the registerLanguage only go over an initialized lsp connection. okay, ignore my comment.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
toinehartman: Ok, then we might need to revisit some other places where we use `get`. Anyway, removed this one!

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/CompletionCapability.java
toinehartman: Sure, changed the interface so the implementer only needs to handle non-null values.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/DynamicCapabilities.java
toinehartman: Changed these functions to improve the responsibilities and timing.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/DynamicCapabilities.java
DavyLandman: execution exception is normally a completable future phase that threw an exception.

toinehartman (approved):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/DynamicCapabilities.java
```diff
@@ -0,0 +1,204 @@
...
+     * Update capabilities for language contributions.
+     * @param contribs The contributions to represent.
+     * @return A void future that completes when all capabilities are updated.
+     */
+    public CompletableFuture<Void> updateCapabilities(Collection<ILanguageContributions> contribs) {
+        // Copy the contributions so we know we are looking at a stable set of elements.
+        // If the contributions change, we expect our caller to call again.
+        var stableContribs = new LinkedHashSet<>(contribs);
+        return CompletableFutureUtils.reduce(supportedCapabilities.stream()
+            .filter(cap -> !staticCapabilities.contains(cap))
```
DavyLandman: if we're using equality, shouldn't the `Capability`  classes make sure to implement equals & hashcode? or do we really want reference equality here?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/DynamicCapabilities.java
```diff
@@ -0,0 +1,204 @@
...
+                        return;
+                    }
+                    unregistrations.forEach(u -> currentRegistrations.remove(u.getMethod()));
+                } else {
+                    logger.debug("No unregistrations to do");
+                }
+                if (!registrations.isEmpty()) {
+                    logger.debug("Registering dynamic capabilities: {}", registrations.stream().map(Registration::getMethod).collect(Collectors.toList()));
+                    try {
+                        client.registerCapability(new RegistrationParams(registrations)).get();
```
DavyLandman: same here, we shouldn't do a `.get`  on a completable future inside of another completable future's body.

DavyLandman (commented):
Looks a lot better, but I have some remaining concerns about the merging of capabilities.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/DynamicCapabilities.java
```diff
@@ -0,0 +1,204 @@
...
+     * @return A void future that completes when all capabilities are updated.
+     */
+    public CompletableFuture<Void> updateCapabilities(Collection<ILanguageContributions> contribs) {
+        // Copy the contributions so we know we are looking at a stable set of elements.
+        // If the contributions change, we expect our caller to call again.
+        var stableContribs = new LinkedHashSet<>(contribs);
+        return CompletableFutureUtils.reduce(supportedCapabilities.stream()
+            .filter(cap -> !staticCapabilities.contains(cap))
+            .map(c -> tryRegistration(c, stableContribs)), parallelExec)
+            .thenAcceptAsync(capabiltities -> {
```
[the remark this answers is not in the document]
toinehartman: Changed the names and logic. Does it read more clearly now?

DavyLandman (commented):
This is a lot more readable, I have some small concerns about the timing of updates on the map, especially if there is a window where the map might run ahead of registered reality. I don't know what the trade-off is, but at least that is not documented in the code right now.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/DynamicCapabilities.java
```diff
@@ -0,0 +1,233 @@
...
+
+        logger.trace("Registering dynamic capability {}", method);
+        return register(registration);
+    }
+
+    private CompletableFuture<Boolean> unregister(Registration reg) {
+        // If our administration contains exactly this registration, remove it and inform the client
+        if (!currentRegistrations.remove(reg.getMethod(), reg)) {
+            return falsy.get();
+        }
```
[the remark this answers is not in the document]
toinehartman: See the answer here https://github.com/usethesource/rascal-language-servers/pull/706#discussion_r2642720700

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/CapabilityRegistration.java
```diff
@@ -0,0 +1,260 @@
...
+    /**
+     * Update capabilities for language contributions.
+     * @param contribs The contributions to represent.
+     * @return A future that completes with a boolean that is false when any registration failed, and true otherwise.
+     */
+    public CompletableFuture<Void> update(Collection<ILanguageContributions> contribs) {
+        return CompletableFuture.supplyAsync(() -> {
+            logger.debug("Updating {} dynamic capabilities from {} contributions", dynamicCapabilities.size(), contribs.size());
+            // Copy the contributions so we know we are looking at a stable set of elements.
+            return lastContributions.updateAndGet(_old -> Set.copyOf(contribs));
```
DavyLandman: how do we know this is the right one? it all depends on the scheduling of `exec` if this `update` is executed later than a previous `update`?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/CapabilityRegistration.java
```diff
@@ -0,0 +1,260 @@
...
+        if (contribs == currentContribs // instance comparison is safe since we took a read-only copy
+            && Objects.equals(currentReg, r)) {
+            // Nothing changed; we're done
+            return noop;
+        }
+
+        // Something changed since we started...
+        // To be sure we arrive at a fixpoint, we need to go again
+        logger.trace("Something changed while registering {}; iterate until nothing changes", cap.methodName());
+        return updateRegistration(cap, currentContribs);
```
DavyLandman: I'm a bit confused that `lastContribution` is not changed? it looks like the loop will only end now if something we at some point was decided to be the  "last contribution" has finished, even though we have no guarantee that that last one is also scheduled/finished last?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/CapabilityRegistration.java
```diff
@@ -0,0 +1,260 @@
...
+                    return false;
+                }
+                currentRegistrations.remove(reg.getMethod(), reg);
+                return true;
+            });
+    }
+
+    /**
+     * Register this registration.
+     * Aims to be atomic, i.e. keeps local administration of registered capabilities in sync with the client.
```
DavyLandman: "aims" means it's not (always) succeeding, what are guarantees? or how is that corrected? (same for the other one)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/CapabilityRegistration.java
toinehartman: Changed when/how `lastContributions` is set: https://github.com/usethesource/rascal-language-servers/pull/706#discussion_r2689848918

Does that resolve this concern?

DavyLandman (approved):
Yes, I think this should work. could you add an entry to the release-notes?

PR #710, opened 2025-08-28
Map optional `causes` field of Message to `RelatedDiagnosticInformation` of Diagnostics

jurgenvinju:
Error causes are a list of Message connected to possibly any Message: error, warning or info.
Each cause in the list is a _conjunctive_ reason that the main message exists. A cause
is not necessarily located in the same file.

We map these causes to RelatedDiagnosticInformation, such that they appear in the Problem View 
as clickable additional information about each diagnostic.

The implementation is non-trivial because this is the first time that a DocumentService
for one LSP document, needs column/offset information on another file to properly convert
the causes locations to hyperlinks in VScode. 

See also https://github.com/usethesource/rascal/pull/2374

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
@@ -46,7 +46,6 @@ import org.rascalmpl.values.ValueFactoryFactory;
 import org.rascalmpl.values.parsetrees.ITree;
 import org.rascalmpl.values.parsetrees.TreeAdapter;
 import org.rascalmpl.vscode.lsp.util.locations.ColumnMaps;
-import org.rascalmpl.vscode.lsp.util.locations.LineColumnOffsetMap;
 import org.rascalmpl.vscode.lsp.util.locations.Locations;
 import io.usethesource.vallang.ICollection;
 import io.usethesource.vallang.IConstructor;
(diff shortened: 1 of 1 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
```diff
@@ -176,19 +176,31 @@ private static void storeFixCommands(IConstructor d, Diagnostic result) {
     }
 
     public static Diagnostic translateDiagnostic(IConstructor d, ColumnMaps cm) {
-        return translateDiagnostic(d, Locations.toRange(getMessageLocation(d), cm));
+        return translateDiagnostic(d, Locations.toRange(getMessageLocation(d), cm), cm);
     }
 
-    public static Diagnostic translateDiagnostic(IConstructor d, LineColumnOffsetMap cm) {
-        return translateDiagnostic(d, Locations.toRange(getMessageLocation(d), cm));
+    public static Diagnostic translateDiagnostic(IConstructor d, LineColumnOffsetMap thisFile, ColumnMaps otherFiles) {
```
DavyLandman, line 182: It's a bit weird that this function takes both the `ColumnMaps` as an argument, and for 1 file (the current) it already has the caller look it up. how about we change it such that we only get a `ColumnMaps` in this function? and then it's responsible for looking it up (since it already knows about `ColumnMaps`.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
```diff
@@ -176,19 +176,31 @@ private static void storeFixCommands(IConstructor d, Diagnostic result) {
...
+        result.setMessage(getMessageString(d));
         result.setRange(range);
 
+
+        if (d.asWithKeywordParameters().hasParameter("causes")) {
+            result.setRelatedInformation(
+                ((IList) d.asWithKeywordParameters().getParameter("causes")).stream()
+                .map(IConstructor.class::cast)
+                .map(c -> new DiagnosticRelatedInformation(
+                    Locations.toLSPLocation(getMessageLocation(d), otherFiles.get(getMessageLocation(d))),
```
DavyLandman, line 198: nit: `getMessageLocation` is called twice, can we avoid that?

PR #712, opened 2025-08-29
parse error message strings

jurgenvinju:
- **reusing the information in an error tree and using the concept of Related Information in LSP diagnostic messages the parse errors were simplified for beginner users, while extra information is provided for expert users**
- **finetuning messages**
- **minor fixes**

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
@@ -37,12 +37,15 @@ import java.util.stream.Stream;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.eclipse.lsp4j.Diagnostic;
+import org.eclipse.lsp4j.DiagnosticRelatedInformation;
 import org.eclipse.lsp4j.DiagnosticSeverity;
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
 import org.rascalmpl.parser.gtd.exception.ParseError;
 import org.rascalmpl.values.ValueFactoryFactory;
 import org.rascalmpl.values.parsetrees.ITree;
+import org.rascalmpl.values.parsetrees.ProductionAdapter;
+import org.rascalmpl.values.parsetrees.SymbolAdapter;
 import org.rascalmpl.values.parsetrees.TreeAdapter;
 import org.rascalmpl.vscode.lsp.IBaseTextDocumentService;
 import org.rascalmpl.vscode.lsp.util.locations.ColumnMaps;
@@ -50,6 +53,7 @@ import org.rascalmpl.vscode.lsp.util.locations.LineColumnOffsetMap;
 import org.rascalmpl.vscode.lsp.util.locations.Locations;
 import io.usethesource.vallang.ICollection;
 import io.usethesource.vallang.IConstructor;
+import io.usethesource.vallang.IInteger;
 import io.usethesource.vallang.IList;
 import io.usethesource.vallang.ISourceLocation;
 import io.usethesource.vallang.IString;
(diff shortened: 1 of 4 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
```diff
@@ -91,71 +89,120 @@ public static interface Template {
...
+            IConstructor errorProd = TreeAdapter.getProduction(errorTree);
+            int dot = ((IInteger) errorProd.get("dot")).intValue();
+            IConstructor prod = (IConstructor) errorProd.get("prod");
+
+            if (ProductionAdapter.isDefault(prod)) {
+                IList symbols = ProductionAdapter.getSymbols(prod);
+                IList prefix = symbols.sublist(0, dot);
+                IList postfix = symbols.sublist(dot, symbols.length());
+
+                related.add(related(cm, stuckLoc, "Experts: a `" + nonterminal + "` was the last but probably not the only grammar rule that was tried at this position."));
```
DavyLandman, line 157: can we move this experts information behind a settings flag? since it's a lot of information and might overwelm the user?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
```diff
@@ -91,71 +89,120 @@ public static interface Template {
...
+        diagnostics.add(cm -> {
+            var d = new Diagnostic(
+                toRange(stuckLoc, cm),
+                "Parsing got stuck here",
+                DiagnosticSeverity.Error,
+                "parser");
+
+            List<DiagnosticRelatedInformation> related = new ArrayList<>();
+
+            related.add(related(cm, stuckLoc,
```
DavyLandman, line 142: can you show a screenshot/video how this looks for users?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
```diff
@@ -91,71 +89,120 @@ public static interface Template {
     }
 
     public static Template generateParseErrorDiagnostic(ParseError e) {
-        return cm -> new Diagnostic(toRange(e, cm), e.getMessage(), DiagnosticSeverity.Error, "parser");
+        return cm -> new Diagnostic(toRange(e, cm), "Parsing got stuck.", DiagnosticSeverity.Error, "parser");
```
DavyLandman, line 92: is there something better for the user than `got stuck`? Something like `Parser could not continue?` Or maybe `Parsing could not recognize this input`?

DavyLandman (approved):
Quite some code, but I like the effort of trying to inform the user more. As long as it's not too invasive (especially with big skipped regions).

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
jurgenvinju: This is the folded diagnostic without any "related information":

image

This is with the unfolded unformation (currently only for non-experts):

image

Every line links to a different part of the source code. Even though it may start at the same position.

jurgenvinju: This is what the error hover looks like:

image

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
jurgenvinju: Decided to remove it again. The incompleteness of it does not help an expert either.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
```diff
@@ -90,76 +87,84 @@ public static interface Template {
...
+                skippedLoc.getBeginLine(),
+                completeErrorTreeLoc.getBeginColumn(),
+                skippedLoc.getBeginColumn()
+        );
+
+        final List<Template> diagnostics = new ArrayList<>();
+
+        diagnostics.add(cm -> {
+            final var d = new Diagnostic(
+                // stuckloc is here for backward compatibility (before parseError was registered on an error tree)
```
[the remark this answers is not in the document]
DavyLandman: ah right, it's also used related. my bad.

Issue #591, opened 2025-03-04
variable name "modules" is erroneously highlighted as a keyword


jurgenvinju:
**Describe the bug**

<img width="559" alt="Image" src="https://github.com/user-attachments/assets/b0bc3488-ec9f-4bb1-88a7-5e15bec2367e" />

sungshik:
Could it be there was a parse error elsewhere in the file? In that case, only TextMate highlighting is enabled and the TM grammar doesn't distinguish the usage of `module` as an identifier from its usage as keyword. (It takes identifier as default. I'd have to look into it, but I think this is a hard problem to solve in auto-generated TM grammars.)

In the absence of parse errors, I get this:

![Image](https://github.com/user-attachments/assets/f0364b07-3472-48f9-91ae-3d8f68858752)

DavyLandman:
Why does the textmate parser match on `modules` while the keyword is `module`?

sungshik:
Hm, good question... It turns out there's a [`"modules"`](https://github.com/usethesource/rascal/blob/e04a9a7cc81be6f8ad380c2fa2609a54bc49604b/src/org/rascalmpl/library/lang/rascal/syntax/Rascal.rsc#L485) literal in the Rascal grammar. And, the TM grammar generator assumes all (ci)literals that occur in the grammar are keywords. However, the production in which `"modules"` occurs doesn't seem to be reachable from `start syntax Module` (only from `start syntax Command`), so the TM grammar generator could actually be improved to only consider reachable (ci)literals as keywords.

DavyLandman:
Aha, that looks like a good feature issue for [SWAT-engineering/rascal-textmate](https://github.com/SWAT-engineering/rascal-textmate) indeed.

sungshik:
Yes, [issue created](https://github.com/SWAT-engineering/rascal-textmate/issues/26)

RedCMD:
soon™️ 
<img width="328" height="127" alt="Image" src="https://github.com/user-attachments/assets/2f8a2d88-255c-47cc-b0a1-81e73d3f35d7" />
```rascal
int main(
	list[loc] modules = [],
	list[loc] libs = []) {
}
```

DavyLandman:
Fixed in #725

Issue #432, opened 2024-08-27
Textmate grammar does not handle more than one location on a single line


rodinaarssen:
**Describe the bug**

The Textmate grammar breaks Rascal's syntax highlighting if more than one location literal with three consecutive slashes appear "consecutively" (i.e., without a location literal with just two consecutive slashes in-between).

**To Reproduce**

![image](https://github.com/user-attachments/assets/f94c9e96-99a0-4747-8ac6-43a41ac982c9)
Note that from the second slash in the second literal

**Desktop (please complete the following information):**
 - Context: VS Code extension
 - Rascal Version 0.40.8-SNAPSHOT (~current main)
 - Rascal-lsp version: 0.12.0-head (current main)

sungshik:
Thank you for reporting. Here's what happens:

  - For instance, on this line, the TextMate tokenizer assigns the following scopes:
    
    ```rascal
    if (|unknown:///| == l && |unknown:///| == l) {
    //             ^^^^^^^^^^^^^^^^^^^^^            constant.regexp
    //                                  ^^^^^^^^^^^ comment
    ```

  - The semantic highlighter subsequently tries to overwrite these scopes to `meta.embedded` to remove the formatting (using our own `uncategorized` semantic token type):
    
    ![image](https://github.com/user-attachments/assets/15b893bb-6c09-4969-af18-a13fcc1f6284)

  - **Issue 1:** However, while the overwrite does remove the color (`comment` is a different shade of gray in this theme), it doesn't remove italic, despite the ~~`fontStyle: italic`~~ in the token/scope inspector.

  - **Issue 2:** Moreover, VS Code seems to use only scopes assigned by the TextMate tokenizer to determine which brackets are included/excluded to match brackets. Brackets in a `comment` scope are excluded. As a result, the closing bracket on line 116 in the screenshot is not properly matched to the opening bracket. This is the same "related issue 2" in [this comment](https://github.com/usethesource/rascal-language-servers/issues/431#issuecomment-2324162495).

DavyLandman:
It looks like we should be opening an issue on microsoft/vscode repo about this interaction bteween the textmate & the semantic tokens. Or not @sungshik ?

sungshik:
I agree. I think the matching+highlighting of brackets in comments in the current release of the extension (i.e., before TextMate) is a symptom of the same issue. E.g., the purple brackets of `(c)` on line 2 in this screenshot:

[...]

PR #726, opened 2025-09-09
New Rascal TextMate Grammar

RedCMD:
fixes:
* #432
* #591

copied grammar from https://github.com/RedCMD/Rascal-Syntax-Highlighter/blob/master/syntaxes/rascal.tmLanguage.json

some test files are here if you want them
https://github.com/RedCMD/Rascal-Syntax-Highlighter/tree/master/tests

### DIFF
--- a/rascal-vscode-extension/package.json
+++ b/rascal-vscode-extension/package.json
@@ -181,19 +181,31 @@
         "language": "rascalmpl",
         "scopeName": "source.rascalmpl",
         "path": "./syntaxes/rascal.tmGrammar.json",
-        "tokenTypes": {
-          "string": "other",
-          "comment": "other"
-        }
+        "unbalancedBracketScopes": [
+            "punctuation.definition.section.replacement.rascal",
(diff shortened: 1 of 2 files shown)


DavyLandman:
Hi @RedCMD thank you for all the work on the Textmate grammar. 

Based on your comments and some other developments we've decided to disable our textmate grammar in VS Code. We do not want to maintaining a second copy of our grammar (especially in the limited textmate format) and have gotten quite good at error recovery such that for most inputs, the user will have syntax highlighting.

So it saddens me to say that I'm going to have to close this PR, even though I know that some VS Code features (such as the brackets) are better working in this textmate based approach.

Happy to read you've made a VS Code extension that contains the textmate grammar.

RedCMD:
I would hope you'll reconsider
I personally think the TextMate grammar is currently superior to the LSP semantic highlighting

TextMate             |  Semantic
:-------------------------:|:-------------------------:
image|image
image|image
image|image
image|image

PR #727, opened 2025-09-09
Map LSP `Range` to Rascal `ISourceLocation`

toinehartman:
The conversion this way is missing from the `Locations` library. Adding this is motivated by #688.


DavyLandman (approved):
Interesting that there is no place where we should have been calling this already. but I guess that is due to the focus lists.

DavyLandman (approved):

PR #730, opened 2025-09-09
Rascal REPLs are now transient

rodinaarssen:
Rascal REPLs were explicitly configured to be persistent


DavyLandman:
I seem to remember I flipped this boolean since it would not allow you to copy data from the repl, as it would get closed as soon as the connection was lost.

rodinaarssen:
Ah, that could be. This value prevents terminals from reappearing after a reload/restart. Let's see if we can find something else that would satisfy both concerns.

PR #744, opened 2025-09-15
Warning about deprecated rascal.mf entries and another case of the sensitive parser

### LINKED ISSUE #300: Rascal MF checker should make sure there is always a space after a `:`
```
Manifest-Version: 0.0.1
Project-Name: test-lib
Source: src/main/rascal
Require-Libraries:

```

fails because `Require-Libraries:` is missing a space after the `:`

We should add another check to the RASCAL.MF validator

DavyLandman:
image
Fixes #300 

And also a deprecation warning:
image

### DIFF
--- a/rascal-vscode-extension/src/ux/RascalMFValidator.ts
+++ b/rascal-vscode-extension/src/ux/RascalMFValidator.ts
@@ -187,7 +187,7 @@ function checkDeprecatedFeatures(mfBody: vscode.TextDocument, diagnostics: vscod
         const [key, value] = line.text.split(":");
         if (key && value && key === "Require-Libraries" && value.trim() !== "") {
             diagnostics.push(new vscode.Diagnostic(line.range,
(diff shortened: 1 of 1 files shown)


jurgenvinju:
The require libraries deprecation warning is already implemented in PathConfig.java, so we can expect duplicate errors.

DavyLandman:
True, but that is in the REPL, and we cannot link that to the IDE, as we don't have a way to invalidate/clean those messages.

So yes, there is duplication, but it does help the user see the problem.

jurgenvinju:
The createPathConfig code that is called by the IDE produces the same messages with the PathConfig as the REPL. 

So I still propose that we register those diagnostics with the IDE when the PathConfig for the IDE is re-computed. There are more interesting warnings and errors which are detected in pom.xml and RASCAL.mf which are missed if we never spin up a terminal.

Also, spinning up a terminal itself may be buggy in the presence of errors in pom.xml, let alone the checker.

PR #747, opened 2025-09-17
Improved definition overlap checks when renaming

### LINKED ISSUE #722: Rascal rename: detect ambiguous/double declarations at use sites properly
**Describe the bug**

Currently, this test fails:
```rascal
@expected{illegalRename}
test bool multiModuleAmbiguous() = testRenameOccurrences({
    byText("Foo", "public int foo = 1;", {0}),
    byText("Bar", "public int bar = 2;", {}),
    byText("Main", "import Foo;
                   'import Bar;
                   'int baz = foo + bar;", {0})
});
```

**Expected behavior**
It should succeed (a.k.a. throw illegal rename errors).

toinehartman:
Closes #722

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
@@ -44,12 +44,10 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiFunction;
 import java.util.function.Function;
-
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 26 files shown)


rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/refactor/Rename.rsc
```diff
@@ -69,10 +69,14 @@ import util::LanguageServer;
 import util::Maybe;
 import util::Reflective;
 
+private bool isQualifiedUse(loc use, Define def) = size(def.id) != use.length;
```
PieterOlivier, line 72: Is this logic sound? No way the lengths can differ without use being qualified?

toinehartman: The escape is not included in the location:
```rascal
void main() {
    int foo = 1;
    bar = \foo; // hold control and hover -> the \ is not part of the hyperlink
}
```

Added some additional tests for this nonetheless.

PR #752, opened 2025-09-17
Fix the logic to compute REPL titles

### LINKED ISSUE #751: REPL titles aren't computed right.
**Describe the bug**

REPL titles aren't computed right when setting `rascal.terminal.name.originFormat` is set to `"Module (qualified)"`.

Image

sungshik:
Fixes #751

image

Post-mortem: As part of the move to JLine 3, the command to startup REPLs was changed. Corresponding changes to extract REPL titles from such commands were made in commit d68ecad, but it didn't work exactly as intended.


DavyLandman (approved):
Could you also at least in the PR description say why this was broken? (like what happened, why did it use to work but not anymore?)

PR #774, opened 2025-09-24
Fixes #773; content panels are now reused if the same title is already present, but they are always reloaded just in case new content is available.

### LINKED ISSUE #773: IDEServices::showInteractive content opens a new tab every time
It should reuse tabs with the same "ids" or at least the same "titles" to avoid a growing number of tabs which are all showing the same content.

jurgenvinju:


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/LSPIDEServices.java
```diff
@@ -92,7 +92,7 @@ public void browse(URI uri, String title, int viewColumn) {
     public void edit(ISourceLocation path) {
         ISourceLocation physical = Locations.toClientLocation(path);
         ShowDocumentParams params = new ShowDocumentParams(physical.getURI().toASCIIString());
-        params.setTakeFocus(true);
+        params.setTakeFocus(false);
```
DavyLandman, line 95: I don't think we should switch this off, since this changes the behavior of the editor that would come forward.

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -71,13 +71,20 @@ export async function activateLanguageClient(
...
-    const allOpenTabs = vscode.window.tabGroups.all.flatMap(tg => tg.tabs);
-    const tabsForThisPanel = allOpenTabs.filter(t => t.input instanceof vscode.TabInputWebview && t.label === externalURL);
-
-    await vscode.window.tabGroups.close(tabsForThisPanel);
+    const id = title;
+    const existingPanel = contentPanels.get(id);
+
+    if (existingPanel) {
+        // reuse the tab, but reload the content
+        existingPanel.reveal(viewColumn, true);
```
DavyLandman, line 84: what if the id is the same, but not the `viewColumn`? shouldn't we then cleanup the old panel at least?

DavyLandman (changes requested):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/LSPIDEServices.java
jurgenvinju: This shouldn't be here indeed. Bad commit

jurgenvinju (commented):
Should remove that debug commit with takeFocus

DavyLandman (approved):

PR #786, opened 2025-09-26
Fix logic of when to clean outdated tpls

PieterOlivier:
The if condition was the wrong way around. We want to clean when the path config has changed.


DavyLandman:
@toinehartman this was an oversight right? I've merged the PR, but please let me know if that was wrong.

PR #787, opened 2025-09-26
editor and webview fixes

jurgenvinju:
* [x] adds viewColumn to showDocument by creating a new method editDocument that does have that parameter
   - makes sure the editor is not a preview document
   - makes sure to take the focus
   - interprets the viewColumn parameter the same as the webview creator
* [x] fixes webview creation
   - breaks a funny cache where due to an equal html document (with a nested iframe that is not equal) a webview 
     would never be updated
   - make sure the webviews don't take focus unnecessarily, such that people working in an editor can continue 
     typing

This is based on https://github.com/usethesource/rascal/pull/2453 and have to wait for an RC to be able to merge this.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/ITerminalIDEServer.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/ITerminalIDEServer.java
@@ -38,8 +38,7 @@ import java.util.Base64.Encoder;
 import java.util.concurrent.CompletableFuture;
 import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
 import org.checkerframework.checker.nullness.qual.Nullable;
-import org.eclipse.lsp4j.ShowDocumentParams;
-import org.eclipse.lsp4j.ShowDocumentResult;
+import org.eclipse.lsp4j.Range;
 import org.eclipse.lsp4j.jsonrpc.services.JsonNotification;
 import org.eclipse.lsp4j.jsonrpc.services.JsonRequest;
 import org.rascalmpl.values.IRascalValueFactory;
@@ -67,7 +66,7 @@ public interface ITerminalIDEServer {
     }
 
     @JsonRequest
-    default CompletableFuture<ShowDocumentResult> edit(ShowDocumentParams edit)  {
+    default CompletableFuture<Void> edit(EditorParameter edit)  {
         throw new UnsupportedOperationException();
     }
 
@@ -230,6 +229,30 @@ public interface ITerminalIDEServer {
         @Override
         public String toString() {
             return "editParameter: " + module;
+        }
+    }
+
+    public static class EditorParameter {
+        private String uri;
+        private int viewColumn;
+        private Range range;
+
+        public EditorParameter(String uri, Range range, int viewColumn) {
+            this.uri = uri;
+            this.range = range;
+            this.viewColumn = viewColumn;
+        }
+
+        public Range getRange() {
+            return range;
+        }
+
+        public String getUri() {
+            return uri;
+        }
+
+        public int getViewColumn() {
+            return viewColumn;
         }
     }
 
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/TerminalIDEClient.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/TerminalIDEClient.java
@@ -35,7 +35,7 @@ import java.net.URI;
 import java.util.concurrent.ExecutionException;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.eclipse.lsp4j.ShowDocumentParams;
+import org.eclipse.lsp4j.Range;
 import org.eclipse.lsp4j.jsonrpc.Launcher;
 import org.jline.terminal.Terminal;
 import org.rascalmpl.debug.IRascalMonitor;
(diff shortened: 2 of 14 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/ITerminalIDEServer.java
```diff
@@ -233,6 +232,30 @@ public String toString() {
         }
     }
 
+    public static class EditorParameter {
```
DavyLandman, line 235: shouldn't this extend `ShowDocumentParams`? to denote it's an extention of lsp feature?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/TerminalIDEClient.java
```diff
@@ -100,16 +101,15 @@ public void browse(URI uri, String title, int viewColumn) {
     }
 
     @Override
-    public void edit(ISourceLocation path) {
+    public void edit(ISourceLocation path, int viewColumn) {
         ISourceLocation physical = Locations.toClientLocation(path);
-        ShowDocumentParams params = new ShowDocumentParams(physical.getURI().toASCIIString());
-        params.setTakeFocus(true);
```
DavyLandman, line 106: we've lost the capability to set focus?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
```diff
@@ -194,7 +194,7 @@ private static void extractReasonAndStackTrace(Throwable e, String task, StringW
 
     private enum ErrorHandlingOption {
         // The order in which these values are declared is the order in which they will appear in the VS Code tooltip
-        REPORT_ON_GITHUB("Report on GitHub"),
+        REPORT_ON_GITHUB("Report on GitHub with usethesource/rascal-language-servers"),
```
DavyLandman, line 197: this does make it a very wide button, I don't think we need to be this specific?

DavyLandman (approved):
Some small comments, but the eslint has shown a few style things in the typescript that need to get fixed.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
jurgenvinju: Well I never recognized this as something that belonged to the Rascal project, because there is no hint to which extension this belongs. I thought it was some co-pilot-like extension of VScode. So I thought let's help the users see that.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/ITerminalIDEServer.java
jurgenvinju: Also it's not extending showDocument, it's replacing it. So documentation of the old feature will not remain relevant.  It's better (more stable) to let them evolve independently.

jurgenvinju:
Can't find any eslint issues. The eslint run looks clean to be. Where to look? 

I do have java lint issues because of the snapshot dependency.

rascal-vscode-extension/package.json
```diff
@@ -14,7 +14,7 @@
     "color": "#ffffff",
     "theme": "light"
   },
-  "version": "0.13.0-head",
+  "version": "0.13.0-opendag",
```
DavyLandman, line 17: this shouldn't be committed. it's something you do locally. but stuff breaks if we merge this to main.

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -104,6 +111,27 @@ async function showContentPanel(url: string, title:string, viewColumn:integer):
...
+        const uri = vscode.Uri.parse(uriString);
+
+        const doc = await vscode.workspace.openTextDocument(uri);
+
+        // Show it in an editor
+        const editor = await vscode.window.showTextDocument(doc, {
+             // make sure it's not a preview, otherwise it will dissappear with the next focus change:
+            preview: false,
+            // put it where we want: if another editor is open for the same URI _and_ the same viewColumn, that one will be reused:
+            viewColumn: viewColumn,
```
DavyLandman, line 124: this does not implement the `-1` and `-2` features of rascal.

jurgenvinju: yes it does. Why do you say that? That's copied straight from the VScode docs.

jurgenvinju:
```
export enum ViewColumn {
		/**
		 * A *symbolic* editor column representing the currently active column. This value
		 * can be used when opening editors, but the *resolved* {@link TextEditor.viewColumn viewColumn}-value
		 * of editors will always be `One`, `Two`, `Three`,... or `undefined` but never `Active`.
		 */
		Active = -1,
		/**
		 * A *symbolic* editor column representing the column to the side of the active one. This value
		 * can be used when opening editors, but the *resolved* {@link TextEditor.viewColumn viewColumn}-value
		 * of editors will always be `One`, `Two`, `Three`,... or `undefined` but never `Beside`.
		 */
		Beside = -2,
		/**
```

rascal-vscode-extension/package.json
jurgenvinju: thanks for the reminder

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -104,6 +111,27 @@ async function showContentPanel(url: string, title:string, viewColumn:integer):
...
 
+async function openEditor(uriString: string, range:vscode.Range, viewColumn: integer) {
+    const uri = vscode.Uri.parse(uriString);
+
+    const doc = await vscode.workspace.openTextDocument(uri);
+
+    // Show it in an editor
+    const editor = await vscode.window.showTextDocument(doc, {
+        // make sure it's not a preview, otherwise it will dissappear with the next focus change:
+        preview: false,
```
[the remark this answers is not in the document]
jurgenvinju: This is IDEServices::edit not IDEServices::preview; so I think we should implement what it says.

PR #800, opened 2025-09-30
Cleanup prints during lint jobs

jurgenvinju:


jurgenvinju:
Thanks @DavyLandman 
* `verify` also runs test and with `-Pcoverage` the coverage analysis is done based on those tests (but also based on what is run during `verify`
* this PR's branch name does not describe the contents anymore
* this PR now only removes 90% of the log which is due to downloading messsages by mvn
* and for the non-testing `mvn` commands it adds `-DskipTests` to make sure we don'r run them more than once.

.github/workflows/lint.yaml
```diff
@@ -85,7 +84,7 @@ jobs:
           distribution: 'temurin'
           cache: 'maven'
       - name: Run checkstyle and print errors
-        run: mvn -B checkstyle:checkstyle  checkstyle:check
+        run: mvn -B checkstyle:checkstyle  checkstyle:check -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn -Drascal.compile.skip -Drascal.tutor.skip -DskipTests
```
DavyLandman, line 87: checkstyle doesn't need the skips, it never gets to compile/test phase.

DavyLandman:
Yes, this is better, but I think we can keep it a bit cleaner (I've left 2 comments)

PR #803, opened 2025-09-30
Fixes broken base64 encoding

PieterOlivier:
The code that encoded a byte stream to a base64 string was broken.
It assumed that an incompleet read meant that the end of the stream was reached, which is not always the case.

This PR replaces the old code with a much simpler approach that uses `Base64InputStream` from apache-common-codecs.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IRascalFileSystemServices.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IRascalFileSystemServices.java
@@ -31,18 +31,18 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.URISyntaxException;
+import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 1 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IRascalFileSystemServices.java
```diff
@@ -195,28 +195,9 @@ default CompletableFuture<Void> createDirectory(URIParameter uri) {
...
-                int read;
-                while ((read = source.read(buffer, 0, BUFFER_SIZE)) == BUFFER_SIZE) {
-                    result.append(encoder.encodeToString(buffer));
-                }
-                if (read > 0) {
-                    // last part needs to be a truncated part of the buffer
-                    buffer = Arrays.copyOf(buffer, read);
-                    result.append(encoder.encodeToString(buffer));
-                }
-                return new LocationContent(result.toString());
```
DavyLandman, line 219: I would have expected some cleanup of the imports?

PR #804, opened 2025-09-30
Refactored the handling of error trees to avoid null trees and extracting common patterns

DavyLandman:
After finding some strange handling of error recovery, @toinehartman and I sat together and refactored it into some more user-friendly functions on `TextDocumentState`.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java
@@ -63,7 +63,7 @@ public class TextDocumentState {
     private final BiFunction<ISourceLocation, String, CompletableFuture<ITree>> parser;
     private final ISourceLocation location;
 
-    private final AtomicReference<@MonotonicNonNull Versioned<Update>> current;
+    private final AtomicReference<Versioned<Update>> current;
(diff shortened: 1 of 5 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java
```diff
@@ -107,14 +107,58 @@ private Update unpackCurrent() {
...
+            });
+        if (!allowRecoveredErrors) {
+            // if a parse is done, always overwrite it with the
+            // last tree that did not have any errors, also not recovered errors
+            result = result.thenApply(t -> lastWithoutErrors.get());
+        }
+
+        return result.handle((t, e) -> {
+            if (t == null) {
+                throw new RuntimeException("We've never seen a parse tree without errors for: " + getLocation());
```
toinehartman, line 155:
```suggestion
                throw new IllegalStateException("No previous parse tree without errors for: " + getLocation());
```

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/terminal/ITerminalIDEServer.java
toinehartman, line 39:
```suggestion
```

PR #812, opened 2025-10-01
Fixed small NPEs all around the code

DavyLandman:
Using checker-framework (branch : `chore/enable-checker-framework`) I found some NPEs, some were small, some were big. But we're slowly working towards reducing these errors.

The initial count of possible problems was 183, and this has been reduced to 150. Not all of them are actual errors, some of them are problems with the code & annotations.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IRascalFileSystemServices.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IRascalFileSystemServices.java
@@ -341,9 +341,9 @@ public interface IRascalFileSystemServices {
 
     public static class SourceLocation {
         private final String uri;
-        private final int[] offsetLength;
-        private final int[] beginLineColumn;
-        private final int[] endLineColumn;
+        private final int @Nullable[]  offsetLength;
+        private final int @Nullable[] beginLineColumn;
+        private final int @Nullable[] endLineColumn;
 
         public static SourceLocation fromRascalLocation(ISourceLocation loc) {
             if (loc.hasOffsetLength()) {
@@ -409,26 +409,44 @@ public interface IRascalFileSystemServices {
         }
 
         public int getOffset() {
+            if (!hasOffsetLength()) {
+                throw new IllegalStateException("This location has no offset");
+            }
             return offsetLength[0];
         }
 
         public int getLength() {
+            if (!hasOffsetLength()) {
+                throw new IllegalStateException("This location has no length");
+            }
             return offsetLength[1];
         }
 
         public int getBeginLine() {
+            if (!hasLineColumn()) {
+                throw new IllegalStateException("This location has no line and columns");
+            }
             return beginLineColumn[0];
         }
 
         public int getBeginColumn() {
+            if (!hasLineColumn()) {
+                throw new IllegalStateException("This location has no line and columns");
+            }
             return beginLineColumn[1];
         }
 
         public int getEndLine() {
+            if (!hasLineColumn()) {
+                throw new IllegalStateException("This location has no line and columns");
+            }
             return endLineColumn[0];
         }
 
         public int getEndColumn() {
+            if (!hasLineColumn()) {
+                throw new IllegalStateException("This location has no line and columns");
+            }
             return endLineColumn[1];
         }
     }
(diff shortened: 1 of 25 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
PieterOlivier, line 395: I think this explicit null check can be removed now that the `null` return value in `ParserOnlyContribution.prepareRename` is replaced by an unknown location. `null` here is now a programming error instead of a flag to indicate renaming is not possible.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParserOnlyContribution.java
```diff
@@ -150,13 +151,13 @@ public CompletableFuture<IList> parseCodeActions(String commands) {
...
     @Override
-    public InterruptibleFuture<IList> inlayHint(@Nullable ITree input) {
+    public InterruptibleFuture<IList> inlayHint(ITree input) {
         return InterruptibleFuture.completedFuture(VF.list());
     }
 
     @Override
     public InterruptibleFuture<ISourceLocation> prepareRename(IList focus) {
-        return InterruptibleFuture.completedFuture(null);
+        return InterruptibleFuture.completedFuture(URIUtil.unknownLocation());
```
PieterOlivier, line 160: In `ParamtericTextDocumentService` there is still an explicit check for null instead of a check for an unknown location.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/Locations.java
```diff
@@ -122,7 +122,7 @@ public static ISourceLocation toLoc(String uri) {
         try {
             return URIUtil.createFromURI(uri);
         } catch (UnsupportedOperationException e) {
-            if (e.getMessage().contains("Opaque URI schemes are not supported")) {
+            if (e.getMessage() != null && e.getMessage().contains("Opaque URI schemes are not supported")) {
```
PieterOlivier, line 125: Why is this code here, is it not more logical to do this check in `createFromURI`?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Lazy.java
```diff
@@ -28,9 +28,10 @@
 
 import java.util.function.Supplier;
 import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
+import org.checkerframework.checker.nullness.qual.NonNull;
 
-public interface Lazy<T> extends Supplier<T> {
-    public static <T> Lazy<T> defer(Supplier<T> generator) {
+public interface Lazy<T extends @NonNull Object> extends Supplier<T> {
```
PieterOlivier, line 33: Interesting, I did not know you could (ab)use type params to check for non-null objects!

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IRascalFileSystemServices.java
```diff
@@ -341,9 +341,9 @@ public boolean isRecursive() {
 
     public static class SourceLocation {
         private final String uri;
-        private final int[] offsetLength;
-        private final int[] beginLineColumn;
-        private final int[] endLineColumn;
+        private final int @Nullable[]  offsetLength;
```
PieterOlivier, line 344: Learned something again. I did not know you could set annotations between the element type and the array indicator.

PieterOlivier (commented):
Looks good. Some interesting stuff I did not know you could do. One small issue for purity to make the contract of `prepareRename` more clear in the code.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Lazy.java
DavyLandman: well, this is a thing, you cannot put annotations on type parameters, so if you want to say, hey the type here will not return a null value, than this is the way to mark it, but in a way, this is more of a CF thing than something else, and I'm unsure how much we want to spread this around.

toinehartman (approved):
LGTM. Note the SQ warnings (mostly unused imports).

PR #815, opened 2025-10-02
Only send changed files to rascal-check, otherwise we check more than needed

DavyLandman:
This PR makes sure that only files which have no tpl, or an outdated tpl are asked to be checked. This aligns to the rascal-maven-plugin behavior.

I think this also fixes reports of people that the checker seemed to type-check other files that aren't changed. But only if the other modules were in other projects.

### DIFF
--- a/rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/IDECheckerWrapper.rsc
+++ b/rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/IDECheckerWrapper.rsc
@@ -100,7 +100,10 @@ map[loc, set[Message]] checkFile(loc l, set[loc] workspaceFolders, start[Module]
             step3("Checked module in `<project.file>`", 1);
             pcfg = getPathConfig(project);
             checkOutdatedPathConfig(pcfg);
-            msgs += check([*modulesPerProject[project]], rascalCompilerConfig(pcfg));
+            modulesToCheck = calculateOutdated(modulesPerProject[project], pcfg);
+            if (modulesToCheck != []) {
+                msgs += check(modulesToCheck, rascalCompilerConfig(pcfg));
+            }
         }
         return true;
     }, totalWork=size(upstreamDependencies));
@@ -108,7 +111,7 @@ map[loc, set[Message]] checkFile(loc l, set[loc] workspaceFolders, start[Module]
     step("Checking module <l>", 1);
     pcfg = getPathConfig(initialProject);
     checkOutdatedPathConfig(pcfg);
-    msgs += check([l], rascalCompilerConfig(pcfg));
+    msgs += check(calculateOutdated(modulesPerProject[initialProject] + l, pcfg), rascalCompilerConfig(pcfg));
     return filterAndFix(msgs, workspaceFolders);
 }, totalWork=3);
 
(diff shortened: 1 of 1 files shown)


rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/IDECheckerWrapper.rsc
```diff
@@ -100,15 +100,18 @@ map[loc, set[Message]] checkFile(loc l, set[loc] workspaceFolders, start[Module]
...
+            }
         }
         return true;
     }, totalWork=size(upstreamDependencies));
 
     step("Checking module <l>", 1);
     pcfg = getPathConfig(initialProject);
     checkOutdatedPathConfig(pcfg);
-    msgs += check([l], rascalCompilerConfig(pcfg));
+    msgs += check(calculateOutdated(modulesPerProject[initialProject] + l, pcfg), rascalCompilerConfig(pcfg));
```
rodinaarssen, line 114: Doesn't the type checker handle files within the project already?

PaulKlint: The assumption is that the typechecker only gets a list of *changed* modules to check.

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/IDECheckerWrapper.rsc
```diff
@@ -174,7 +183,7 @@ loc targetToProject(loc l) {
 loc inferProjectRoot(loc member) {
     current = targetToProject(member);
     if (!isDirectory(current)) {
-        current = current.parent;
+         current = current.parent;
```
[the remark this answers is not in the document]
DavyLandman: :D I almost rewrote that function to use recursion and get better hits on the memo table, but that line 194 thing makes it a lot harder.

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/IDECheckerWrapper.rsc
DavyLandman: But if I would pass in the request to check `A` and it imports `B`, would it not figure out that `B` is newer than its tpl?

DavyLandman: Okay, I'll keep the behavior as here in the PR, just so that it more closely follows the maven plugin. With the understanding that the typechecker should not only lean on that, but keep its checking logic for tpl invalidation.

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/IDECheckerWrapper.rsc
```diff
@@ -143,10 +155,7 @@ void checkOutdatedPathConfig(PathConfig pcfg) {
     }
 }
 
-bool tplInputsChanged(PathConfig old, PathConfig new)
-    = old.srcs != new.srcs
-    || old.libs != new.libs
-    ;
+bool tplInputsChanged(PathConfig old, PathConfig new) = old[messages=[]] != new[messages=[]];
```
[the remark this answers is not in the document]
DavyLandman: True, but I wanted to be a bit more conservative, so that if something changes there, we won't have to remember that here we encoded a specific "equality" in the path configs

PR #816, opened 2025-10-02
removed the pom dependencies from the JVM classpath for the terminal. Also we do not have to ask them anymore from the Rascal LSP server

jurgenvinju:


rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -146,13 +146,11 @@ export class RascalExtension implements vscode.Disposable {
                 }
                 progress.report({increment: 20, message: "Requesting IDE configuration"});
                 const serverConfig = await rascal.sendRequest<IDEServicesConfiguration>("rascal/supplyIDEServicesConfiguration");
-                progress.report({increment: 25, message: "Calculating project class path"});
-                const compilationPath = await rascal.sendRequest<string[]>("rascal/supplyProjectCompilationClasspath", { uri: uri?.toString() });
```
DavyLandman, line 150: Are there other parts where we use this? if not, let's remove our extension of the lsp protocol.

PR #817, opened 2025-10-03
Turn PathConfig messages into diagnostics

PieterOlivier:
During PathConfig building messages about pom.xml and RASCAL.MF might be generated.
This PR translates these messages into lsp diagnostics and keeps them synchronized when files are saved or projects are closed.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IBaseTextDocumentService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IBaseTextDocumentService.java
@@ -54,6 +54,10 @@ public interface IBaseTextDocumentService extends TextDocumentService {
     void initialized();
     void registerLanguage(LanguageParameter lang);
     void unregisterLanguage(LanguageParameter lang);
+
+    void projectAdded(String name, String uri);
+    void projectRemoved(String name, String uri);
+
     CompletableFuture<IValue> executeCommand(String languageName, String command);
     LineColumnOffsetMap getColumnMap(ISourceLocation file);
     ColumnMaps getColumnMaps();
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IBaseTextDocumentService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IBaseTextDocumentService.java
@@ -55,8 +55,8 @@ public interface IBaseTextDocumentService extends TextDocumentService {
     void registerLanguage(LanguageParameter lang);
     void unregisterLanguage(LanguageParameter lang);
 
-    void projectAdded(String name, String uri);
-    void projectRemoved(String name, String uri);
+    void projectAdded(String name, ISourceLocation projectRoot);
+    void projectRemoved(String name, ISourceLocation projectRoot);
 
     CompletableFuture<@Nullable IValue> executeCommand(String languageName, String command);
     LineColumnOffsetMap getColumnMap(ISourceLocation file);
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
@@ -213,8 +213,7 @@ public class RascalTextDocumentService implements IBaseTextDocumentService, Lang
     public void connect(LanguageClient client) {
         this.client = client;
         this.rascalServices = new RascalLanguageServices(this, workspaceService, (IBaseLanguageClient) client, ownExecuter);
-        this.facts = new FileFacts(ownExecuter, rascalServices, columns);
-        facts.setClient(client);
+        this.facts = new FileFacts(ownExecuter, rascalServices, client, columns);
     }
 
     @Override
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
@@ -572,12 +572,12 @@ public class RascalTextDocumentService implements IBaseTextDocumentService, Lang
     }
 
     @Override
-    public void projectAdded(String name, String uri) {
+    public void projectAdded(String name, ISourceLocation projectRoot) {
     }
 
     @Override
-    public void projectRemoved(String name, String uri) {
-        facts.projectRemoved(URIUtil.assumeCorrectLocation(uri));
+    public void projectRemoved(String name, ISourceLocation projectRoot) {
+        facts.projectRemoved(projectRoot);
     }
 
     @Override
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
@@ -103,6 +103,7 @@ import org.eclipse.lsp4j.services.LanguageClientAware;
 import org.rascalmpl.library.Prelude;
 import org.rascalmpl.library.util.PathConfig;
 import org.rascalmpl.uri.URIResolverRegistry;
+import org.rascalmpl.uri.URIUtil;
 import org.rascalmpl.values.parsetrees.ITree;
 import org.rascalmpl.values.parsetrees.ProductionAdapter;
 import org.rascalmpl.values.parsetrees.TreeAdapter;
@@ -563,6 +564,15 @@ public class RascalTextDocumentService implements IBaseTextDocumentService, Lang
     @Override
     public void unregisterLanguage(LanguageParameter lang) {
         throw new UnsupportedOperationException("registering language is a feature of the language parametric server, not of the Rascal server");
+    }
+
+    @Override
+    public void projectAdded(String name, String uri) {
+    }
+
+    @Override
+    public void projectRemoved(String name, String uri) {
+        facts.projectRemoved(URIUtil.assumeCorrectLocation(uri));
     }
 
     @Override
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
@@ -102,7 +102,6 @@ import org.eclipse.lsp4j.services.LanguageClientAware;
 import org.rascalmpl.library.Prelude;
 import org.rascalmpl.library.util.PathConfig;
 import org.rascalmpl.uri.URIResolverRegistry;
-import org.rascalmpl.uri.URIUtil;
 import org.rascalmpl.values.parsetrees.ITree;
 import org.rascalmpl.values.parsetrees.ProductionAdapter;
 import org.rascalmpl.values.parsetrees.TreeAdapter;
@@ -573,6 +572,7 @@ public class RascalTextDocumentService implements IBaseTextDocumentService, Lang
 
     @Override
     public void projectAdded(String name, ISourceLocation projectRoot) {
+        // No need to do anything
     }
 
     @Override
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/FileFacts.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/FileFacts.java
@@ -69,6 +69,10 @@ public class FileFacts {
         this.confs = new PathConfigs(exec, new PathConfigDiagnostics(client, cm));
     }
 
+    public void projectRemoved(ISourceLocation projectLocation) {
+        confs.expungePathConfig(projectLocation);
+    }
+
     public void invalidate(ISourceLocation file) {
         getFile(file).invalidate();
     }
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/FileFacts.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/FileFacts.java
@@ -61,15 +61,12 @@ public class FileFacts {
     private final ColumnMaps cm;
     private final PathConfigs confs;
 
-    public FileFacts(Executor exec, RascalLanguageServices rascal, ColumnMaps cm) {
+    public FileFacts(Executor exec, RascalLanguageServices rascal, LanguageClient client, ColumnMaps cm) {
         this.exec = exec;
         this.rascal = rascal;
-        this.cm = cm;
-        this.confs = new PathConfigs();
-    }
-
-    public void setClient(LanguageClient client) {
         this.client = client;
+        this.cm = cm;
+        this.confs = new PathConfigs(new PathConfigDiagnostics(client, cm));
     }
 
     public void invalidate(ISourceLocation file) {
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/FileFacts.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/FileFacts.java
@@ -35,7 +35,6 @@ import java.util.concurrent.Executor;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.PublishDiagnosticsParams;
@@ -56,7 +55,7 @@ public class FileFacts {
     private static final Logger logger = LogManager.getLogger(FileFacts.class);
     private final Executor exec;
     private final RascalLanguageServices rascal;
-    private volatile @MonotonicNonNull LanguageClient client;
+    private final LanguageClient client;
     private final Map<ISourceLocation, FileFact> files = new ConcurrentHashMap<>();
     private final ColumnMaps cm;
     private final PathConfigs confs;
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/FileFacts.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/FileFacts.java
@@ -66,7 +66,7 @@ public class FileFacts {
         this.rascal = rascal;
         this.client = client;
         this.cm = cm;
-        this.confs = new PathConfigs(new PathConfigDiagnostics(client, cm));
+        this.confs = new PathConfigs(exec, new PathConfigDiagnostics(client, cm));
     }
 
     public void invalidate(ISourceLocation file) {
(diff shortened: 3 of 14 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IBaseTextDocumentService.java
```diff
@@ -54,6 +54,10 @@ public interface IBaseTextDocumentService extends TextDocumentService {
     void initialized();
     void registerLanguage(LanguageParameter lang);
     void unregisterLanguage(LanguageParameter lang);
+
+    void projectAdded(String name, String uri);
+    void projectRemoved(String name, String uri);
```
DavyLandman, line 59: how about we turn the `String uri` into an `URI` or a `ISourceLocation`? to avoid people getting creative with them in ways we don't want?

rascal-lsp/src/test/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigDiagnosticsTest.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    @Before
+    public void before() throws URISyntaxException {
+        mockedClient = mock(LanguageClient.class);
+        columnMaps = new ColumnMaps(this::getContents);
+        sut = new PathConfigDiagnostics(mockedClient, columnMaps);
+    }
+
+    private String resourceUrl(String res) {
```
DavyLandman, line 90: nit: I think this should return a URI, leave it to the consumers to downgrade it to a string.

rascal-lsp/src/test/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigDiagnosticsTest.java
```diff
@@ -0,0 +1,171 @@
...
+        ClassLoader classLoader = getClass().getClassLoader();
+        return classLoader.getResource(res).toString();
+    }
+
+    private ISourceLocation resourceLocation(String res) {
+        try {
+            return URIUtil.createFromURI(resourceUrl(res));
```
DavyLandman, line 97: there is also `IValueFactory::sourcelocation` that accepts a URI as paramter.

rascal-lsp/src/test/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigDiagnosticsTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+    @Test
+    public void testClearDiagnostics() {
+        sut.publishDiagnostics(projectA, VF.list(msg(POM_A, MSG1, 4, 7)));
+        sut.publishDiagnostics(projectB, VF.list(msg(POM_A, MSG2, 5, 7)));
+
+        reset(mockedClient);
+        sut.clearDiagnostics(projectA);
+        verify(mockedClient).publishDiagnostics(params(pomAUrl, diag(4, 7, MSG2)));
+    }
```
DavyLandman, line 169: Nice that we have detailed tests. We don't have them for any other interaction in the system, and many places it's hard to mock, as a lot of context needs to be setup.

So nice, but a set of UI test would have been enough I think, at least to match the rest of the tests.

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
```diff
@@ -28,10 +28,9 @@
 import { expect } from 'chai';
 import * as fs from 'fs/promises';
 import * as path from 'path';
-import { TextEditor, ViewSection, VSBrowser, WebDriver, Workbench } from 'vscode-extension-tester';
+import { TextEditor, ViewSection, VSBrowser, WebDriver, Workbench, until } from 'vscode-extension-tester';
 import { Delays, IDEOperations, ignoreFails, printRascalOutputOnFailure, sleep, TestWorkspace } from './utils';
 
-
 const protectFiles = [TestWorkspace.mainFile, TestWorkspace.libFile, TestWorkspace.libCallFile];
```
DavyLandman, line 34: add the manifest to the protected files.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigDiagnostics.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+    }
+
+    private void publishFileDiagnostics(ISourceLocation file) {
+        List<Diagnostic> fileDiagnostics = new ArrayList<>();
+        for (List<Diagnostic> diags :  publishedProjectDiagsPerFile.getOrDefault(file, Collections.emptyMap()).values()) {
+            fileDiagnostics.addAll(diags);
+        }
+        client.publishDiagnostics(new PublishDiagnosticsParams(file.getURI().toString(), fileDiagnostics));
+    }
```
DavyLandman, line 115: the interaction between this function and the work in `publishDiagnostics` via the common fields is what makes it harder for me to review what is happening.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigDiagnostics.java
```diff
@@ -0,0 +1,116 @@
...
+/*
+ * This class updates VSCode diagnostics based on PathConfig messages.
+ * Messages are tracked by project and by file. So changes in one project will not affect
+ * messages genrated by other projects, even if files with messages belong to multiple projects.
+ */
+/*package*/ class PathConfigDiagnostics {
+    private final LanguageClient client;
+    private final ColumnMaps cm;
+    private final Map<ISourceLocation, Map<ISourceLocation, List<Diagnostic>>> publishedProjectDiagsPerFile = new HashMap<>();
+    private final Map<ISourceLocation, Set<ISourceLocation>> filesWithDiagsPerProject = new HashMap<>();
```
DavyLandman, line 58: the field names and later variable names based on these are confusing, because they mix file & project a lot. 

We have to think of better names, or perhaps a intermediate class.


I think it might be conceptually easier to group diagnostics based on the target file, but then track the origin of them?

so something like:

```
class TracedDiagnostic extends Diagnostic {
    final ISourceLocation originProject;
}

private final Map<ISourceLocation /* file */, List<TracedDiagnostics> /*all messages> currentDiagnostics;
```

and then every `publishDiagnostics` we update these lists, and we keep track of both what we changed (added/removed) and then send this worklist to the function that publishes them.

DavyLandman (changes requested):
Yes, this looks quite nice. But I think the `PathConfigDiagnostics` class could be improved, it's a bit complex to follow all the maps & lists and how they interact.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/FileFacts.java
```diff
@@ -61,15 +61,16 @@ public class FileFacts {
...
+    public FileFacts(Executor exec, RascalLanguageServices rascal, LanguageClient client, ColumnMaps cm) {
         this.exec = exec;
         this.rascal = rascal;
+        this.client = client;
         this.cm = cm;
-        this.confs = new PathConfigs();
+        this.confs = new PathConfigs(exec, new PathConfigDiagnostics(client, cm));
     }
 
-    public void setClient(LanguageClient client) {
```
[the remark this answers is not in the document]
PieterOlivier: It is not caused by something in this PR, `setClient` was called directly after the constructor anyway. I just folded them into one as the body of the constructor need changing anyway.
I think this code has evolved up to a point that the constructor and `setClient` drifted together.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/FileFacts.java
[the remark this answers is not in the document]
PieterOlivier: Are we going to use `@NonNull` on all fields that cannot hold `null`? I understand the checker benefits, but that does not improve readability imho.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigDiagnostics.java
```diff
@@ -0,0 +1,116 @@
...
+            List<Diagnostic> newDiagnostics = entry.getValue();
+
+            Map<ISourceLocation, List<Diagnostic>> fileDiagsPerProject = publishedProjectDiagsPerFile.computeIfAbsent(file, (f) -> new LinkedHashMap<>());
+            @Nullable List<Diagnostic> publishedDiags = fileDiagsPerProject.get(file);
+
+            // Only publish diagnostics for a file if the diagnostics for that file have changed in the current project
+            if (!newDiagnostics.equals(publishedDiags)) {
+                // The diagnostics for this file related to the current project have changed.
+                // We need to re-publish all diagnostics for this file (including the ones from other projects)
+                fileDiagsPerProject.put(project, newDiagnostics);
```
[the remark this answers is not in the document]
PieterOlivier: That does not fly in this case: we need to know if the old list has changed. If the merge function is called we could check it, but then we would mis the case where the merge function is not called (the list is not in the map yet).
Having an extra check beforehand defeats the purpose of using merge.
The thread-safeness must be guaranteed on a higher level of abstraction here.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigDiagnostics.java
```diff
@@ -0,0 +1,116 @@
...
+    public void publishDiagnostics(ISourceLocation project, IList messages) {
+        // Gather messages per file
+        Map<ISourceLocation, List<Diagnostic>> diagnosticsPerFile = Diagnostics.translateMessages(messages, cm);
+
+        // Publish diagnostics, but only for files for which diagnostics have changed
+        for (var entry : diagnosticsPerFile.entrySet()) {
+            ISourceLocation file = entry.getKey().top();
+            List<Diagnostic> newDiagnostics = entry.getValue();
+
+            Map<ISourceLocation, List<Diagnostic>> fileDiagsPerProject = publishedProjectDiagsPerFile.computeIfAbsent(file, (f) -> new LinkedHashMap<>());
```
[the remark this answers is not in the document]
PieterOlivier: Also the reason for using a `LinkedHashMap` was that I wanted to make sure the order of diagnostics for the same line was relatively stable. Maybe we should order diagnostics alphabetically to make absolutely sure.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IBaseTextDocumentService.java
PieterOlivier: I totally agree. `ISourceLocations` is fine here and matches some of the other methods in this class.

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
PieterOlivier: Awesome, I never noticed these!

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigDiagnostics.java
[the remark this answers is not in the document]
PieterOlivier: I have now synchronized the code at a higher level without introducing the risk of deadlocks.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigDiagnostics.java
```diff
@@ -0,0 +1,141 @@
...
+                    // The diagnostics for this file related to the current project have changed.
+                    // We need to re-publish all diagnostics for this file (including the ones from other projects)
+                    diagsPerProject.put(project, newDiagnostics);
+                    filesToRepublish.add(file);
+                }
+            }
+
+            // Now we have published the diagnostics of all files that occur in the list of new messages
+            // Time to remove diagnostics from files that our project previously published diagnostics for but no longer have diagnostics
+            Set<ISourceLocation> newFiles = diagnostics.keySet().stream().map(file -> file.top()).collect(Collectors.toSet());
```
DavyLandman, line 97: nit: `file.top()` -> `ISourceLocation::top`

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigDiagnostics.java
```diff
@@ -0,0 +1,141 @@
...
+
+import io.usethesource.vallang.IList;
+import io.usethesource.vallang.ISourceLocation;
+
+/*
+ * This class updates VSCode diagnostics based on PathConfig messages.
+ * Messages are tracked by project and by file. So changes in one project will not affect
+ * messages genrated by other projects, even if files have messages belonging to different projects.
+ */
+/*package*/ class PathConfigDiagnostics {
```
DavyLandman, line 54: To avoid retaining memory for long runs of VS Code, we have to cleanup entries in `projectDiagsPerFile`. for example if no project has a diagnostic for a file anymore.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigDiagnostics.java
```diff
@@ -0,0 +1,141 @@
...
+        Set<ISourceLocation> affectedFiles;
+
+        synchronized (this) {
+            affectedFiles = filesWithDiagsPerProject.remove(project);
+            if (affectedFiles == null) {
+                return;
+            }
+
+            for (ISourceLocation file : affectedFiles) {
+                projectDiagsPerFile.get(file).remove(project);
```
DavyLandman, line 122: if we start cleaning up the `projectDiagsPerFile` map, we have to become more defensive as `get` might return a null.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigs.java
```diff
@@ -99,28 +122,54 @@ private PathConfig buildPathConfig(ISourceLocation projectRoot) {
...
             this.currentPathConfigs = currentPathConfigs;
+            projectWatches = new ConcurrentHashMap<>();
         }
 
         // we detect changes to roots, and keep track of the last changed time
         // the thread will clear them if the time is longer than the timeout
         private final Map<ISourceLocation, Long> changedRoots = new ConcurrentHashMap<>();
+
+        // Watch a single file. We keep track of the watch registrations so we can unwatch
+        // these files when the project is closed.
```
DavyLandman, line 151: nit: if we turn these in `/** ... */` java docs, the comment is also visible in the UI for down-stream users.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigs.java
```diff
@@ -99,28 +122,54 @@ private PathConfig buildPathConfig(ISourceLocation projectRoot) {
...
         }
 
-        private static final long UPDATE_DELAY = TimeUnit.SECONDS.toNanos(5);
+        public void unregisterProject(ISourceLocation projectRoot) throws IOException {
+            List<WatchRegistration> registrations = projectWatches.remove(projectRoot);
+            if (registrations != null) {
+                for (WatchRegistration registration : registrations) {
+                    reg.unwatch(registration.file, false, registration.callback);
+                }
+            }
```
DavyLandman, line 170: yes, nice and clean.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigs.java
```diff
@@ -143,7 +192,8 @@ public void run() {
                         // we clear it from the list, as the path config calculation
                         // can take some time
                         changedRoots.remove(root);
-                        currentPathConfigs.replace(root, actualBuild(root));
+                        var pathConfig = actualBuild(root);
+                        currentPathConfigs.replace(root, pathConfig);
```
DavyLandman, line 196: why do we need this change? I guess it's an incomplete reverse of something in the PR that wasn't needed anymore?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
```diff
@@ -211,8 +212,7 @@ public void pair(BaseWorkspaceService workspaceService) {
     public void connect(LanguageClient client) {
         this.client = client;
         this.rascalServices = new RascalLanguageServices(this, workspaceService, (IBaseLanguageClient) client, ownExecuter);
-        this.facts = new FileFacts(ownExecuter, rascalServices, columns);
-        facts.setClient(client);
+        this.facts = new FileFacts(ownExecuter, rascalServices, client, columns);
```
DavyLandman, line 215: yes :+1: much better.

DavyLandman (changes requested):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigs.java
PieterOlivier: You are right, I will revert this change.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigDiagnostics.java
```diff
@@ -0,0 +1,141 @@
...
+ */
+/*package*/ class PathConfigDiagnostics {
+    private final LanguageClient client;
+    private final ColumnMaps cm;
+
+    // For each file, keep track of the list of diagnostics per project.
+    private final Map<ISourceLocation, Map<ISourceLocation, List<Diagnostic>>> projectDiagsPerFile = new HashMap<>();
+
+    // For each project, keep track of the set of files that currently have diagnostics for that project
+    private final Map<ISourceLocation, Set<ISourceLocation>> filesWithDiagsPerProject = new HashMap<>();
```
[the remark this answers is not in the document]
PieterOlivier: I did not succeed in implementing this without locks. I was able to bring down the number of map fields to one.

DavyLandman (approved):
This got a lot better 2/3 small comments left.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigDiagnostics.java
```diff
@@ -0,0 +1,145 @@
...
+
+    private void publishDiagnostics(ISourceLocation project, Map<ISourceLocation, List<Diagnostic>> diagnostics) {
+        for (PublishDiagnosticsParams params : updateDiagnostics(project, diagnostics)) {
+            client.publishDiagnostics(params);
+        }
+    }
+
+    private synchronized List<PublishDiagnosticsParams> updateDiagnostics(ISourceLocation project, Map<ISourceLocation, List<Diagnostic>> diagnostics) {
+        // Assume all files that currently have diagnostics for our project need to be republished unless proven otherwise
+        List<ISourceLocation> filesToRepublish = new ArrayList<>();
```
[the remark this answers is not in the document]
PieterOlivier: Yes this was a misguided attempt to keep the order of file diagnostic updates consistent. Initially I had tests that depended on this but that was more a shortcoming of the tests than of the code. This is no longer an issue so I have replaced the list with a set.

PR #820, opened 2025-10-08
Remove nullables from selection range

toinehartman:
Remove `@Nullable` annotations and possible `null` values from selection range computations.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SelectionRanges.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SelectionRanges.java
@@ -32,7 +32,7 @@ import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.stream.Collectors;
 
-import org.checkerframework.checker.nullness.qual.Nullable;
+import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
 import org.eclipse.lsp4j.SelectionRange;
 import org.rascalmpl.values.IRascalValueFactory;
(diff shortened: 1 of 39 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SelectionRanges.java
```diff
@@ -53,23 +53,39 @@ private SelectionRanges() { /* hide implicit constructor */ }
      * @param columns The editor's column map.
      * @return A range with optional parent ranges.
      */
-    public static @Nullable SelectionRange toSelectionRange(IList ranges, ColumnMaps columns) {
+    public static SelectionRange toSelectionRange(IList ranges, ColumnMaps columns) {
```
DavyLandman, line 56: Should we keep this overload?

DavyLandman (approved):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/SelectionRanges.java
toinehartman: I merged the overloads. This also reduces the need to throw exceptions, since we can always build an empty range.

DavyLandman (changes requested):
Good to factor this out! Some questoins regarding naming.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/concurrent/CompletableFutureUtils.java
```diff
@@ -0,0 +1,90 @@
...
+            IList::concat
+        );
+    }
+
+    /**
+     * Combines a {@link Stream} of {@link CompletableFuture} into a single future that yields a {@link C}.
+     * @param <I> The type of the results of the input futures.
+     * @param <C> The type of the result of the combined future.
+     * @param futures A {@link Stream} of futures to combine.
+     * @param identity The identity function of {@link C}.
```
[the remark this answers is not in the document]
DavyLandman: okay. fair.

DavyLandman (approved):

PR #826, opened 2025-10-13
Use the location of the cause and not the original location

### LINKED ISSUE #825: Causes in double declaration error message have wrong location
**Describe the bug**

_Causes_ in double declaration error message have wrong location

**To Reproduce**

Create the following module which contains a double declaration error that is properly flagged:

Image

Here is how the error message for line 6 looks like:

Image

Observe that the causes field points to the declaration on line 6 instead of line 3 (and vice versa)

If we run the type checker on the command line we see that the correct locations are generated,
i.e., the error on line 3 points in its causes to line 6 and vice versa:

```
list[ModuleMessages]: [program(
    |file:///Users/paulklint/git/rascal/src/org/rascalmpl/compiler/lang/rascalcore/compile/Examples/D.rsc|(0,73,<1,0>,<6,12>),
    {
      error(
        "Double declaration of `n`",
        |file:///Users/paulklint/git/rascal/src/org/rascalmpl/compiler/lang/rascalcore/compile/Examples/D.rsc|(52,1,<3,4>,<3,5>),
        causes=[info(
            "Other declaration of `n`",
            |file:///Users/paulklint/git/rascal/src/org/rascalmpl/compiler/lang/rascalcore/compile/Examples/D.rsc|(65,1,<6,4>,<6,5>))]),
      error(
        "Double declaration of `n`",
        |file:///Users/paulklint/git/rascal/src/org/rascalmpl/compiler/lang/rascalcore/compile/Examples/D.rsc|(65,1,<6,4>,<6,5>),
        causes=[info(
            "Other declaration of `n`",
            |file:///Users/paulklint/git/rascal/src/org/rascalmpl/compiler/lang/rascalcore/compile/Examples/D.rsc|(52,1,<3,4>,<3,5>))])
    })]
```

**Expected behavior**
Causes in line 3 point to line 3; causes in line 6 points to line 3.

 - Context: VS Code extension 0.13.0-head2768

DavyLandman:
Fixes #825

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Diagnostics.java
@@ -220,7 +220,7 @@ public class Diagnostics {
                 ((IList) dKW.getParameter("causes")).stream()
                 .map(IConstructor.class::cast)
                 .map(c -> new DiagnosticRelatedInformation(
-                    Locations.toLSPLocation(getMessageLocation(d), otherFiles.get(getMessageLocation(d))),
+                    Locations.toLSPLocation(getMessageLocation(c), otherFiles),
                     getMessageString(c)))
                 .collect(Collectors.toList())
             );


jurgenvinju (approved):
looks like I copied some code and failed to adapt to the current context. good catch!

PR #846, opened 2025-10-16
Use native jsonrpc error codes instead of our own IOErrors

DavyLandman:
This also reduces all the nulls in our VFS client side.


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/jsonrpc/messages/BooleanResult.java
```diff
@@ -30,17 +30,16 @@
...
 import org.eclipse.lsp4j.jsonrpc.validation.NonNull;
 
-public class BooleanResult extends IOResult {
-    private @Nullable Boolean result;
+public class BooleanResult {
+    private Boolean result;
 
-    public BooleanResult(@NonNull int errorCode, @Nullable String errorMessage, @Nullable Boolean result) {
-        super(errorCode, errorMessage);
+    public BooleanResult(@NonNull int errorCode, String errorMessage, Boolean result) {
```
toinehartman: I think the `error...` params can be removed here

PieterOlivier (approved):
Looks good to me!

PR #849, opened 2025-10-20
Abort type-check when dependencies have parse errors.

### LINKED ISSUE #832: duplicate parse error reported
**Describe the bug**

Image

The generic "Parse error in" error which is always reported on line 1 and column 1 is confusing me. I want to look only at the real parse errors with the accurate locations.

I suspect this is produced by a downstream tree processor in the LSP server? We can/should assume that all parse errors have been reported by the parser process, or will be reported shortly asynchronously.

toinehartman:
**Recovered parse error**
_Shows 1 cause per disambiguated error tree_
afbeelding

**Fatal parse error**
_Shows 1 cause for the file with an error_
afbeelding

Closes #832

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
@@ -167,7 +167,8 @@ public class RascalLanguageServices {
         return e.getFunctionValueFactory().function(getParseTreeType, (t, u) -> {
             ISourceLocation resolvedLocation = Locations.toClientLocation((ISourceLocation) t[0]);
             try {
-                var tree = rascalTextDocumentService.getFile(resolvedLocation).getCurrentTreeAsync(false).get();
+                // although we cannot type-check modules with errors, we prefer to get the errors here instead of retrying the parse and still failing after this try-block
+                var tree = rascalTextDocumentService.getFile(resolvedLocation).getCurrentTreeAsync(true).get();
                 if (tree != null) {
                     return tree.get();
                 }
--- a/rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/IDECheckerWrapper.rsc
+++ b/rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/IDECheckerWrapper.rsc
@@ -85,7 +85,7 @@ map[loc, set[Message]] checkFile(loc l, set[loc] workspaceFolders, start[Module]
                         ml = locateRascalModule(modName, getPathConfig(currentProject), getPathConfig, workspaceFolders);
                         if (ml.extension == "rsc", mlpt := getParseTree(ml), mlpt.src.top notin checkedForImports) {
                             if (hasParseErrors(mlpt)) {
-                                return (l: {error("Cannot typecheck this module, since a dependency has parse error(s).", openFileHeader.src,
+                                return (l: {error("Cannot typecheck this module, since dependency `<modName>` has parse error(s).", openFileHeader.src,
                                     causes=[error("Has a parse error around this position.", e.src) | Tree e <- findBestParseErrors(mlpt)])
                                 });
                             }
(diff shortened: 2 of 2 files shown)


rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/IDECheckerWrapper.rsc
```diff
@@ -82,9 +97,13 @@ map[loc, set[Message]] checkFile(loc l, set[loc] workspaceFolders, start[Module]
             checkedForImports += currentSrc;
             checkForImports -= tree;
         }
-        return true;
+        return {};
     }, totalWork=1);
 
+    if ({} !:= parseMsgs) {
+        return toMap(parseMsgs);
```
DavyLandman: why not make `parseMsgs` a map? it looks like the code is only building a rel with a single entry?

DavyLandman (changes requested):
Looks quit small, which is great. I have some questions while reading the source code.

DavyLandman (approved):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -166,12 +167,15 @@ IFunction makeParseTreeGetter(Evaluator e) {
         return e.getFunctionValueFactory().function(getParseTreeType, (t, u) -> {
             ISourceLocation resolvedLocation = Locations.toClientLocation((ISourceLocation) t[0]);
             try {
-                var tree = rascalTextDocumentService.getFile(resolvedLocation).getLastTreeWithoutErrors();
+                var tree = rascalTextDocumentService.getFile(resolvedLocation).getCurrentTreeAsync(true).get();
```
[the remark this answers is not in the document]
DavyLandman: why is it now `getCurrentTreeAsync(true)`  again when it used to be `false`?

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/IDECheckerWrapper.rsc
```diff
@@ -53,7 +53,29 @@ import lang::rascalcore::check::ModuleLocations;
...
     = job("Rascal check", map[loc, set[Message]](void(str, int) step) {
-    checkForImports = [getParseTree(l)];
+
+    tuple[start[Module], set[Message]] getParseTreeOrErrors(loc l, str name, loc errorLocation) {
+        try {
+            t = getParseTree(l);
+            errors = hasParseErrors(t)
+                ? {error("Cannot typecheck this module, since dependency `<name>` has parse error(s).", errorLocation,
+                        causes=[error("Parse error around this position.", e.src) | e <- findBestParseErrors(t)])}
+                : {};
```
DavyLandman: it we throw a `ParseError` here, we can reuse the same code of the `catch` a few lines lower and reduce the duplication.

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/IDECheckerWrapper.rsc
```diff
@@ -53,7 +53,29 @@ import lang::rascalcore::check::ModuleLocations;
...
+            return <t, errors>;
+        } catch ParseError(loc err): {
+            return <(start[Module]) `module Placeholder`, {error("Cannot typecheck this module, since dependency `<name>` has parse error(s).", errorLocation, causes=[error("Parse error(s).", err)])}>;
+        }
+    }
+
+    // Note: check further down parses again, possibly leading to a different tree if the contents changed in the meantime
+    <openFile, parseErrors> = getParseTreeOrErrors(l, "unknown", |unknown:///|);
```
DavyLandman: why pass in `|unknown:///|`? `l` would also be fine?

DavyLandman (changes requested):
Yes, this is looking good. Some small tweaks of the code and I think it's done.

DavyLandman (approved):

PR #855, opened 2025-10-30
Migrate to Rascal `ColumnMaps` port

toinehartman:
This PR migrates to Rascal's implementation of `ColumnMaps` and friends. Since this was ported from this repository, this PR also removes the code clone.


DavyLandman:
I thought this was cloned to the rascal project? So shouldn't we fix that and remove this clone? paging @rodinaarssen

PR #856, opened 2025-10-31
Remote IDE services

rodinaarssen:


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IBaseLanguageServerExtensions.java
```diff
@@ -59,6 +55,11 @@ default CompletableFuture<Two<String, URI[]>[]> supplyPathConfig(PathConfigParam
     @JsonNotification("rascal/vfs/register")
     void registerVFS(VFSRegister registration);
 
+    @JsonNotification("rascal/supplyRemoteIDEServicesConfiguration")
+    default CompletableFuture<IDEServicesConfiguration> supplyRemoteIDEServicesConfiguration() {
+        throw new UnsupportedOperationException();
+    }
```
DavyLandman: why did this move?

rodinaarssen: No particular reason; that's an artefact of how this code evolved. I'll move it up

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IBaseLanguageClient.java
```diff
@@ -58,4 +63,44 @@ public interface IBaseLanguageClient extends LanguageClient {
...
 
+    public static class BrowseParameter {
+        private String uri;
+        private String title;
+        private int viewColumn;
+
+        public BrowseParameter(URI uri, IString title, IInteger viewColumn) {
+            this.uri = uri.toString();
+            this.title = title.getValue();
+            this.viewColumn = viewColumn.intValue();
```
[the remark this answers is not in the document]
rodinaarssen: That's right!

DavyLandman:
> @rodinaarssen @DavyLandman What is this PR? "Remote IDE services" _sounds_ interesting, but I can't tell if it _really is_ interesting 😝

This is part of a large project where we want to disconnect the version of the Rascasl REPL from the rascal version in the  VS Code extension. Right now you always get the packaged rascal version. But we're working towards a setup where the extension respects your pom.xml. Read more about it here: https://github.com/usethesource/rascal/issues/2171

But to enable that we've had to disconnect some hard dependencies in the REPL. One of the last ones (only VFS is remaining) was this IDE Service, that now instead of being a java connection between the REPL & rascal-lsp server, it's now a jsonrcp connection.


I think in summary: not so interesting, mainly internal engineering to prepare for a future where we can do more interesting stuff such as upgrading the vs code extension without forcing the users to also upgrade their rascal version.

PR #857, opened 2025-10-31
Fix flaky UI tests caused by unexpected edits during type-check

toinehartman:
Fixes flaky tests that showed up while developing #849. Sometimes (for unclear reasons, but probably due to [one](https://issues.chromium.org/issues/42320821) of [various](https://issues.chromium.org/issues/42320393) [issues](https://selenium-users.narkive.com/l4cqRtlX/issue-sendkeys-command-eats-up-last-characters) with `sendKeys`), `TextEditor::getText` inserts the letter 'A' in an open editor, introducing parse errors. Before this PR, type-checks would (erroneously) use the last parse tree without errors, incidentally hiding this introduced parse error.

This PR supplies the current parse tree to type-checks instead, and works around the functions from [redhat-developer/vscode-extension-tester](https://github.com/redhat-developer/vscode-extension-tester) that cause these parse errors. It also changes how failed tests are handled, by taking an extra screenshot at that time.

Joint effort with @rodinaarssen.


rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
```diff
@@ -103,6 +101,10 @@ parameterizedDescribe(function (errorRecovery: boolean) {
         await fs.writeFile(TestWorkspace.picoFile, picoFileBackup);
     });
 
+    afterEach("print output in case of test failure",
+        async function () { await printRascalOutputOnFailure(this, ide, "Language Parametric Rascal"); }
```
DavyLandman: `async` and a single `await`  can also just be `function(){ return ...; }` which is the same, with less keywords.

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
```diff
@@ -157,11 +161,6 @@ describe('IDE', function () {
     });
 
     it("go to definition works across projects", async () => {
-        // due to a current bug, we have to make sure that the lib in the other project is correctly resolved
-        const libEditor = await ide.openModule(TestWorkspace.libFile);
-        await triggerTypeChecker(libEditor, "", true);
-        await bench.getEditorView().closeAllEditors();
```
DavyLandman: which bug? is this the fix that we're not triggering upstream checks anymore?

rascal-vscode-extension/src/test/vscode-suite/utils.ts
```diff
@@ -464,42 +470,40 @@ async function assureDebugLevelLoggingIsEnabled() {
     await prompt.confirm();
 }
 
-export function printRascalOutputOnFailure(channel: 'Language Parametric Rascal' | 'Rascal MPL') {
-
+export async function printRascalOutputOnFailure(context: Mocha.Context, ide: IDEOperations, channel: 'Language Parametric Rascal' | 'Rascal MPL') {
```
DavyLandman: can this function now become like this again? to do an after each?

DavyLandman (changes requested):
What a fix this has been. Damn, I have some small comments

rascal-vscode-extension/src/test/vscode-suite/utils.ts
toinehartman: Sure, I reverted it. That also adresses your other comments regarding duplication. It means we lose the screenshot at failure time, but that was not that useful anyway.

DavyLandman (approved):

PR #876, opened 2025-11-13
Add change annotations to LSP library

### LINKED ISSUE #872: Allow for annotating text edits with `needsConfirmation` such that code actions can be reviewed by users
Since LSP 3.16 we have to option to have text edits be reviewed by users (but not all of them). We're using that already in the rename code, but we would like to expose this for all `util::LanguageServer` places where we allow for `DocumentEdits`.

In essence copying this design: https://github.com/usethesource/rascal-language-servers/blob/d8b75ee8bab28bf0538f52c12c1425ebccfcef02/rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/refactor/rename/Common.rsc#L59-L60 over to `util::LanguageServer`.

toinehartman:
Port `ChangeAnnotation` from rename-only to `util::LanguageServer`, exposing it to other Rascal contributions and parametric language servers.

Closes #872

### DIFF
--- a/rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
+++ b/rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
@@ -577,8 +577,9 @@ In LSP, text edits can contain extra information w.r.t. ((analysis::diff::edits:
 * needsConfirmation: Flags whether the user should confirm this change. By default, this is false, which means that ((util::LanguageServer::TextEdit))s are applied without user information.
 
 Typically, clients provide options to group edits by label/description when showing them to the user.
-
 See the [LSP documentation](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#changeAnnotation) for more details.
+
+Note: to easily annotate all text edits in a ((analysis::diff::edits::TextEdits::FileSystemChange)), use the convenience keywords on ((util::LanguageServer::FileSystemChange)).
 }
 @pitfalls{
 When `needsConfirmation = false` for all edits, the client will typically apply them without showing any information from the annotations to the user.
--- a/rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
+++ b/rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
@@ -569,22 +569,27 @@ the right ((DocumentEdit))s immediately.
 }
 data Message(list[CodeAction] fixes = []);
 
-@synopsis{Annotate any text edit with a label and description and require user confirmation.}
+@synopsis{A ((analysis::diff::edits::TextEdits::TextEdit)) with additional context for LSP.}
 @description{
-Annotations contain extra information about text edits in LSP.
+In LSP, text edits can contain extra information w.r.t. ((analysis::diff::edits::TextEdits::TextEdit)).
 * label: Human-readable string that describes the change.
 * description: Human-readable string that additionally describes the change, rendered less prominently.
 * needsConfirmation: Flags whether the user should confirm this change. By default, this is false, which means that ((util::LanguageServer::TextEdit))s are applied without user information.
+
+Typically, clients group edits by label when showing them to the user.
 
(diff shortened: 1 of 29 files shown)


DavyLandman (approved):
I'm missing changes for this on the java side? or was is already there in our translations from Rascal to LSP?

jurgenvinju (changes requested):
Love this feature! The fact that the grouping feature is on the client side (by label), makes this extra nice.

The OO-ness of the LSP leaks through too much to Rascal programmers imho. And also the names are inconsistent as well as pleonasms;
1. We are annotating TextEdits but the annotation is called a Change annotation. Does not compute 😉
2. We are annotating an edit using a keyword field, which is an annotation. It's like calling the Maybe ADT "MaybeADT" instead of just "Maybe". Or a change keyword field is called "changekeyword field". Etc. 
3. We add superfluous nesting structure without adding real structural information like this.

It feels like this OO design is useful for abstraction and future backward compatible extensions, but which would work counter productive on the Rascal side. Here we have other language features for extensibility without adding more structure.

Propose to bring this down to the minimum without losing structure or making the mapping to JSON harder:

```
data TextEdit(str label="", str description=label, bool needsConfirmation=false);
```

Same information, less OO-ish, fewer names/misnomers. Same functionality, same extensibility. Wdyt?

Next to this it seems natural to add the same fields to FileChanges, and project the two fields down to every individual TextEdit while converting to JSON-LSP objects. This will reduce the memory footprint of a large list of changes as well as their cache locality (since most edits won't need their own map of keyword parameters anymore), as well as save the time for building all these tables. I'm guessing most labels and descriptions will be constant within one file change list.

Groetjes!

jurgenvinju:
I forgot `bool needsConfirmation=false`. Added above

jurgenvinju (approved):
Great and clean code. Only that one doc comment still talks about ChangeAnnotation

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
```diff
@@ -604,6 +604,29 @@ the right ((DocumentEdit))s immediately.
 }
 data Message(list[CodeAction] fixes = []);
 
+@synopsis{A ((analysis::diff::edits::TextEdits::TextEdit)) with additional context for LSP.}
+@description{
+In LSP, text edits can contain extra information w.r.t. ((analysis::diff::edits::TextEdits::TextEdit)).
+* label: Human-readable string that describes the change.
+* description: Human-readable string that additionally describes the change, rendered less prominently.
+* needsConfirmation: Flags whether the user should confirm this change. By default, this is false, which means that ((util::LanguageServer::TextEdit))s are applied without user information.
```
DavyLandman: User information -> user interaction/confirmation

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
```diff
@@ -604,6 +604,29 @@ the right ((DocumentEdit))s immediately.
...
+
+Typically, clients provide options to group edits by label/description when showing them to the user.
+See the [LSP documentation](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#changeAnnotation) for more details.
+
+Note: to easily annotate all text edits in a ((analysis::diff::edits::TextEdits::FileSystemChange)), use the convenience keywords on ((util::LanguageServer::FileSystemChange)).
+}
+@pitfalls{
+When `needsConfirmation = false` for all edits, the client will typically apply them without showing any information from the annotations to the user.
```
DavyLandman: I think it has an influence on labels in undo stack?

DavyLandman (changes requested):
Two small text changes

rascal-lsp/src/main/rascal/library/util/LanguageServer.rsc
toinehartman: It does!
afbeelding

Will add that as well.

PR #877, opened 2025-11-17
Add evaluate, setVariable and completions to DAP implementation

jbdoderlein:
Hi,

This PR add a possible implementation of 3 new command to the Rascal DAP implementation : `evaluate`, `setVariable` and `completions`. 
In VSCode, these new commands allow interacting with the Rascal Interpreters inside the Debug Console, and allow changing values of locals variable in the debug panel.

https://github.com/user-attachments/assets/9c41516b-d973-4417-888f-686d50790309

I implemented the evaluation so that it runs in the same interpreter as the debugger, which requires disabling the EventTrigger during evaluation to avoid triggering a breakpoint.

For now, completion is minimal and only considers variables in the current frame and functions in the module and imported modules.

Does this approach seem sound?


DavyLandman:
@jbdoderlein this is cool! 

You'll have to retarget the PR to the rascal repo, as we're in the process of moving DAP towards rascal project.

So this code is quite outdated. We're working towards a 0.13.0 release and won't be releasing a 0.12.x.

PR #894, opened 2025-11-20
Add support for extensionless documents to the parametric language server

### LINKED ISSUE #862: Unable to use the parametric language server for documents without an extension
**Describe the bug**

VS Code opens various documents in normal use that cause the language server to fail with error popups in the corner and errors logged in the language-specific Output channel.

Some examples are:
1. Untitled documents manually set to the Rascal-backed language
   > Request textDocument/inlayHint failed.
   > Message: Unknown file: |untitled:///:Untitled-1|
2. Comparing unsaved content with the saved file via the `workbench.files.action.compareWithSaved` command
   > Request textDocument/foldingRange failed.
   > Message: Unknown file: |showModifications:///c:/Source/path/to/file.lang-ext?%7B%22scheme%22:%22file%22,%22query%22:%22%22%7D|
   - I don't think this is relevant, but to save you some time decoding, the above URI ends with `file.lang-ext?{"scheme":"file","query":""}`
3. Hovering over text in the GitHub Copilot chat window
   > Request textDocument/hover failed.
   > Message: Unknown file: |vscode-chat-code-block://8bd8f3ec-f0be-4107-906e-1a245c268bb0/

In all cases, the failing LSP request was preceded with a `textDocument/didOpen` message.

**To Reproduce**

1. Open VS Code where a Rascal-based DSL extension is installed
2. Do any of these (for example)
   1. Open a new file and manually set the language to `mylang` or whatever your language is called
   2. Open an existing file and execute the `workbench.files.action.compareWithSaved` command
   3. Ask GitHub Copilot to generate code in your language (this could be more tricky)
3. See errors in UI and Output window

**Expected behavior**

I expect Rascal to provide language services for any document - whether it is a physical file or a virtual document from a `TextDocumentContentProvider`.

**Stack traces**

Similar stack for all examples given:

```log
2025-11-03 13:59:04,459 [pool-3-thread-1] DEBUG - org.rascalmpl.vscode.lsp.parametric.ParametricTextDocumentService Did Open file: TextDocumentItem [
  uri = "vscode-chat-code-block://8bd8f3ec-f0be-4107-906e-1a245c268bb0/response_34e8482e-0df2-4c33-882d-4bd8fb1902c2/0#%7B%22references%22%3A%5B%5D%7D"
  languageId = "mylang"
  version = 1
  text = ""
]
2025-11-03 13:59:04,460 [pool-3-thread-1] WARN - org.eclipse.lsp4j.jsonrpc.RemoteEndpoint Notification threw an exception: {
  "jsonrpc": "2.0",
  "method": "textDocument/didOpen",
  "params": {
    "textDocument": {
      "uri": "vscode-chat-code-block://8bd8f3ec-f0be-4107-906e-1a245c268bb0/response_34e8482e-0df2-4c33-882d-4bd8fb1902c2/0#%7B%22references%22%3A%5B%5D%7D",
      "languageId": "mylang",
      "version": 1,
      "text": ""
    }
  }
}
java.lang.UnsupportedOperationException: Rascal Parametric LSP has no support for this file: vscode-chat-code-block://8bd8f3ec-f0be-4107-906e-1a245c268bb0/response_34e8482e-0df2-4c33-882d-4bd8fb1902c2/0#%7B%22references%22%3A%5B%5D%7D
	at org.rascalmpl.vscode.lsp.parametric.ParametricTextDocumentService.contributions(ParametricTextDocumentService.java:412) ~[rascal-lsp.jar:?]
	at org.rascalmpl.vscode.lsp.parametric.ParametricTextDocumentService.contributions(ParametricTextDocumentService.java:399) ~[rascal-lsp.jar:?]
	at org.rascalmpl.vscode.lsp.parametric.ParametricTextDocumentService.lambda$open$8(ParametricTextDocumentService.java:444) ~[rascal-lsp.jar:?]
	at java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1705) ~[?:?]
	at org.rascalmpl.vscode.lsp.parametric.ParametricTextDocumentService.open(ParametricTextDocumentService.java:443) ~[rascal-lsp.jar:?]
	at org.rascalmpl.vscode.lsp.parametric.ParametricTextDocumentService.didOpen(ParametricTextDocumentService.java:230) ~[rascal-lsp.jar:?]
```

 - Context: VS Code w/ Rascal Metaprogramming Language 0.12.2
 - DSL Extension using
    - rascal v0.40.17
    - rascal-lsp v2.21.1

### LINKED ISSUE #868: Add basic support for files without extension
If a file without extension is opened in VS Code and the language is set to the language-parametric server, the server cannot determine the language.

Proposal: if only a single language has been registered, treat it as this language.

urbanfly:
When there is only one language contributed and the document has no extension, use the single language.

Fixes #862 
Fixes #868

### DIFF
--- a/rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
+++ b/rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
@@ -138,14 +138,11 @@ parameterizedDescribe(function (errorRecovery: boolean) {
 
     it("has syntax highlighting in documents without extension", async function () {
         await bench.executeCommand("workbench.action.files.newUntitledFile");
-        console.log("I opened a file");
         await bench.executeCommand("workbench.action.editor.changeLanguageMode");
-        console.log("I triggered change language mode");
 
         const inputBox = new InputBox();
         await inputBox.setText("parametric-rascalmpl");
         await inputBox.confirm();
-        console.log("I changed language mode");
 
         const file = "Untitled-1";
         const editor = await driver.wait(async () => {
@@ -157,22 +154,17 @@ parameterizedDescribe(function (errorRecovery: boolean) {
         }, Delays.normal, "Could not open file");
         expect(editor).to.not.be.undefined;
 
-        console.log("I found an editor");
         await editor.setText(`begin
   declare
      a : natural;
   a := 2
 end
 `);
-        console.log("I updated the text");
         await ide.hasSyntaxHighlighting(editor, Delays.slow);
-        console.log("I found highlighting");
 
         try {
             await editor.setTextAtLine(4, "  a := ");
-            console.log("I updated text");
             await ide.hasErrorSquiggly(editor, Delays.slow);
-            console.log("I found errors");
         } finally {
             await ide.revertOpenChanges();
         }


rascal-vscode-extension/package.json
```diff
@@ -266,7 +266,7 @@
     }
   },
   "scripts": {
-    "lsp4j:package": "cp ../rascal-lsp/target/rascal-lsp*.jar assets/jars/rascal-lsp.jar && cp ../rascal-lsp/target/lib/*.jar assets/jars/",
+    "lsp4j:package": "cp ../rascal-lsp/target/rascal-lsp*.jar assets/jars/ && cp ../rascal-lsp/target/lib/*.jar assets/jars/",
```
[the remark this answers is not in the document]
DavyLandman: This copies the jar to the wrong name. So this can't work.

I think you might need to remove the jars from your target folder, it might have more than one rascal-lsp jar in there?

rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
```diff
@@ -103,14 +110,10 @@ parameterizedDescribe(function (errorRecovery: boolean) {
...
-    it("have highlighting and parse errors", async function () {
+    it("has highlighting and parse errors", async function () {
         await ignoreFails(new Workbench().getEditorView().closeAllEditors());
         const editor = await ide.openModule(TestWorkspace.picoFile);
-        const isPicoLoading = ide.statusContains("Pico");
-        // we might miss this event, but we wait for it to show up
-        await ignoreFails(driver.wait(isPicoLoading, Delays.normal, "Pico parser generator should have started"));
-        // now wait for the Pico parser generator to disappear
-        await driver.wait(async () => !(await isPicoLoading()), Delays.verySlow, "Pico parser generator should have finished", 100);
+        await waitForPico();
```
DavyLandman: this waiting should only be in the first test of the file.

So I don't think we should be extracting this into a separate function and calling again a bit later, as that would be hiding a different bug.

(the whole reason it's a copy of `loadPico` is that sometimes we've had issues where the original loader missed a step in waiting, and then while we're already looking for syntax errors, the VS Code extension on a slow CI server is still loading the language).

rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
```diff
@@ -137,6 +140,33 @@ parameterizedDescribe(function (errorRecovery: boolean) {
...
+        }, Delays.normal, "Could not open file");
+        expect(editor).to.not.be.undefined;
+
+        await editor!.setText(`begin
+  declare
+     a : natural;
+  a := 2;
+end
+`, true);
+        await ide.hasSyntaxHighlighting(editor!, Delays.slow);
```
DavyLandman: Maybe we should also test that if we add invalid syntax, we get parsing errors reported? so that the parser works, but also the diagnostics propogation.

rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
```diff
@@ -137,6 +140,33 @@ parameterizedDescribe(function (errorRecovery: boolean) {
         }
     });
 
+    it("has syntax highlighting in documents without extension", async function () {
```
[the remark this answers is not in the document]
urbanfly: Both. It had syntax highlighting, but would not get red squiggles in an untitled file.

DavyLandman (approved):
If the test succeed, I think this is ready to merge.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -646,7 +646,17 @@ private boolean isLanguageRegistered(ISourceLocation loc) {
     }
 
     private Optional<String> safeLanguage(ISourceLocation loc) {
-        return Optional.ofNullable(registeredExtensions.get(extension(loc)));
+        var ext = extension(loc);
+        if ("".equals(ext)) {
+            if (contributions.size() == 1) {
+                logger.trace("file was openeed without an extension; falling back to the single registered language for: {}", loc);
```
rodinaarssen:
```suggestion
                logger.trace("file was opened without an extension; falling back to the single registered language for: {}", loc);
```

urbanfly:
Thanks for the help, @rodinaarssen. Can you help me understand how "not clearing diagnostics" was causing my test to fail on MacOS?

---

Also, why is the change to `isLanguageRegistered` necessary? I don't like making it "lie" about whether a language is registered. It is used:

1. when a file is deleted, to publish empty diags
   - But why even check `isLanguageRegistered` here? If the client is informing the language server that a file was deleted, then that document was already (manually?) associated with that language. What would be the harm in just clearing diags without the check?

2. to skip `triggerAnalyzer`, when files are opened or changed (or when a lang. is registered "late")
   - The check here appears to be trying to avoid an `UnsupportedOperationException` when `facts(loc)` calls `language(loc)`.
   - Wouldn't it be better to make `facts(loc)` use `safeLanguage(loc)` and fail gracefully if no language is provided? maybe adding `safeFacts(loc)`?

3. to skip certain renamed files
   - Used within `bundleRenamesByContribution`, but wouldn't it be better to use `safeLanguage` and `continue;` if there is no language?

urbanfly:
> uncleared diagnostics were causing UX issues

The test doesn't assert "no red squiggles", though.

> isn't it, in essence, the same lie as in `safeLanguage`

One is asking an explicit question "Is this language registered?"; the other is saying "give me the language for this document - if you can figure it out". I don't see the change to `safeLanguage` as a "lie", just that it can provide a language in more cases now.

> I don't think there's any guarantee that the argument of `didDeleteFiles` only contains files that we're managing.

@DavyLandman and I experimented with this a bit and I think we agreed that removing the condition is fine. It will be called for unmanaged files, but there should be no harm in telling VS Code "we don't have any diags for this file"

rodinaarssen:
I updated the code after discussing with @DavyLandman and made this branch up-to-date.

`isLanguageRegistered` has been removed

DavyLandman (approved):

PR #902, opened 2025-11-26
Clearing diagnostics of transient files after a close event by mimicking a delete

rodinaarssen:
Our TextDocumentService implementations now explicitly check, after receiving and handling a `DidCloseTextDocument` event, whether the closed file still exists. For transient files, such as unsaved untitled files, this will not be the case. In such cases, we now mimic a file deletion event in order to clear out any remaining diagnostics coupled to the closed file. This solves an issue where non-clearable (or, at least, not trivially clearable) diagnostics were kept around in the IDE.

This is a partial fix for @urbanfly's test issues #894.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
@@ -45,7 +45,6 @@ import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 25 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -335,14 +336,21 @@ public void didClose(DidCloseTextDocumentParams params) {
...
+        });
     }
 
     @Override
     public void didDeleteFiles(DeleteFilesParams params) {
         exec.submit(() -> {
             // if a file is deleted, and we were tracking it, we remove our diagnostics
             for (var f : params.getFiles()) {
-                if (isLanguageRegistered(URIUtil.assumeCorrectLocation(f.getUri()))) {
+                if (isLanguageRegistered(Locations.toLoc(f.getUri()))) {
```
toinehartman: Nice catch👍. I think #897 was missing more conversions. Fixing those in #903.

PR #903, opened 2025-11-26
Fix URI conversions

toinehartman:
In the spirit of #897, URI conversions should be applied in many more places. This PR does that.


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
rodinaarssen: Is this one safe as is?

rodinaarssen (approved):
Nice! Just a small question

DavyLandman (approved):
I think we should update the developer.md to state:

- `URI`s are in the VS Code domain (possible as a string from lsp)
- `ISourceLocation` are in the rascal domain
- Conversion between URIs and ISourcelocation should happen via de `Locations`  class, and only at the first/last possible time, just when crossing the barrier.

PR #904, opened 2025-11-27
Improved diagnostics clearing

rodinaarssen:


DavyLandman (approved):
I agree with the SQ warnings/errors in the file, if we take care of those, it's ready to merge.

PR #958, opened 2026-01-27
Quick fix: hide `target` from search results

### LINKED ISSUE #941: Quick-fix settings of rascal projects to hide `target` folder from search
Since we're using maven in rascal, rascal files get copied to the `target/classes` folder. But that copy should not be edited by anyone.

Change the template but also for existing projects, add a small quick-fix to add the [`.vscode/settings.json`](https://code.visualstudio.com/docs/configure/settings#_workspace-settingsjson-location)  entry that hides the target folder from search.

Note we shouldn't hide the target folder, but only exclude them from search:

```json
{
    "search.exclude": {
        "/target/": true,
    },
}
```

toinehartman:
This adds a couple of aids to quickly build a useful `.vscode/settings.json`.

When no `search.exclude` entry exists in the settings, we offer a quick-fix for excluding the `target` folder from VS Code search results. 
afbeelding

Closes #941.

### DIFF
--- a/rascal-vscode-extension/src/ux/VsCodeSettingsFixer.ts
+++ b/rascal-vscode-extension/src/ux/VsCodeSettingsFixer.ts
@@ -31,6 +31,7 @@ import { buildMFChildPath, isRascalProject } from './RascalMFValidator';
 
 const VSCODE_DIR = ".vscode";
 const SETTINGS_FILE = "settings.json";
+const SETTINGS_GLOB = posix.join("**", VSCODE_DIR, SETTINGS_FILE);
 
 const SEARCH_EXCLUDE_SETTING_KEY = "search.exclude";
 const EXCLUDE_ENTRY = `"/target/": true,`;
@@ -52,6 +53,9 @@ export class VsCodeSettingsFixer implements vscode.Disposable {
     private readonly disposables: vscode.Disposable[] = [];
 
     constructor () {
+        // Register code actions
+        this.disposables.push(vscode.languages.registerCodeActionsProvider({ pattern: SETTINGS_GLOB }, new FixSettingsActions()));
+
         this.diagnostics = vscode.languages.createDiagnosticCollection("Rascal project settings");
         this.disposables.push(this.diagnostics);
 
@@ -66,10 +70,9 @@ export class VsCodeSettingsFixer implements vscode.Disposable {
             }
         });
 
-        const settingsGlob = posix.join("**", VSCODE_DIR, SETTINGS_FILE);
-        const watcher = vscode.workspace.createFileSystemWatcher(settingsGlob);
-        watcher.onDidCreate(f => this.fixSettings(this.projectRootFromSettings(f)), this, this.disposables);
-        watcher.onDidChange(f => this.fixSettings(this.projectRootFromSettings(f)), this, this.disposables);
+        const watcher = vscode.workspace.createFileSystemWatcher(SETTINGS_GLOB);
+        watcher.onDidCreate(f => this.fixSettings(projectRoot(f)), this, this.disposables);
+        watcher.onDidChange(f => this.fixSettings(projectRoot(f)), this, this.disposables);
         watcher.onDidDelete(this.clearFileDiagnostics, this, this.disposables);
         this.disposables.push(watcher);
 
@@ -77,13 +80,6 @@ export class VsCodeSettingsFixer implements vscode.Disposable {
         for (const projectRoot of vscode.workspace.workspaceFolders || []) {
             this.fixSettings(projectRoot.uri); // Do not await; process workspaces in parallel
         }
-
-        // Register code actions
-        this.disposables.push(vscode.languages.registerCodeActionsProvider({ pattern: settingsGlob }, new FixSettingsActions()));
-    }
-
-    private projectRootFromSettings(uri: vscode.Uri): vscode.Uri {
-        return uri.with({ path: posix.dirname(posix.dirname(uri.path)) });
     }
 
     private clearWorkspaceDiagnostics(uri: vscode.Uri) {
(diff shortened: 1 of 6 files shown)


rascal-vscode-extension/src/ux/VsCodeSettingsFixer.ts
```diff
@@ -0,0 +1,189 @@
...
+        await this.fixWorkspaceSettings(projectRoot);
+    }
+
+    private async createWorkspaceSettings(projectRoot: vscode.Uri) {
+        const projectName = posix.basename(projectRoot.path);
+        const settingsFile = vscode.Uri.joinPath(projectRoot, VSCODE_DIR, SETTINGS_FILE);
+
+        if (fs.existsSync(settingsFile.fsPath)) {
+            return;
+        }
```
DavyLandman: we should only use `vscode.workspace.fs` and never the `fs` of node. (this is true for the rest of this file too)

rascal-vscode-extension/src/ux/VsCodeSettingsFixer.ts
```diff
@@ -0,0 +1,189 @@
...
+        for (const d of context.diagnostics) {
+            switch (d.code) {
+                case FixKind.insertEmptySettings: {
+                    const insertEmptyBlock = new vscode.CodeAction("Insert blank settings", vscode.CodeActionKind.QuickFix);
+                    insertEmptyBlock.diagnostics = [d];
+                    insertEmptyBlock.edit = new vscode.WorkspaceEdit();
+                    insertEmptyBlock.edit.insert(document.uri, document.positionAt(0), EMPTY_SETTINGS);
+                    result.push(insertEmptyBlock);
+                    break;
+                }
```
DavyLandman: why is this a quick-fix? if there are no settings, that's not something that's a problem that needs fixing. since adding an empty block of code has the same semantics as a non-existing file. so this fixes nothing.

DavyLandman (changes requested):
I think the UX of this quick-fix is suboptimal. It requires extra steps that the user shouldn't be bothered with, and it warns for things that are not problem.

I think we should say: hey this project does not have a  `target` entry for `search.exclude`, shall we fix that for you? and then we fix it. No other messages, no other steps involved, we get it done.

rascal-vscode-extension/src/ux/VsCodeSettingsFixer.ts
```diff
@@ -0,0 +1,189 @@
...
+        for (const projectRoot of vscode.workspace.workspaceFolders || []) {
+            this.createAndFixWorkspaceSettings(projectRoot.uri); // Do not await; process workspaces in parallel
+        }
+
+        // Register code actions
+        this.disposables.push(vscode.languages.registerCodeActionsProvider({ pattern: posix.join("**", VSCODE_DIR, "**") }, new FixSettingsActions()));
+    }
+
+    private projectRootFromSettings(uri: vscode.Uri): vscode.Uri {
+        return vscode.Uri.joinPath(uri, "..", "..");
```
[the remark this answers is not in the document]
toinehartman: I wasn't aware of that one. However, it does not have a [`dirname`](https://github.com/microsoft/vscode-uri?tab=readme-ov-file#usage-util) method. I'll use `vscode.Uri.with` to change the path instead.

rascal-vscode-extension/src/ux/VsCodeSettingsFixer.ts
```diff
@@ -0,0 +1,202 @@
...
+`;
+const DEFAULT_SETTINGS = `{
+    ${SEARCH_EXCLUDE_SETTING}
+}`;
+
+const TARGET_IN_SEARCH_WARNING = "The projects target folder appears in search results. Editing those files breaks the project build.";
+const EXCLUDE_TARGET_TITLE = "Exclude target from search";
+
+const Y = "Create settings.json and ignore target folder";
+const N = "Do not update settings";
```
DavyLandman: is there a reason the UX messages are all as constants at the top of the file?

rascal-vscode-extension/src/ux/VsCodeSettingsFixer.ts
```diff
@@ -0,0 +1,202 @@
...
+/**
+ * Create/fix the VS Code settings file and suggest appropriate settings for Rascal projects.
+ */
+export class VsCodeSettingsFixer implements vscode.Disposable {
+
+    private readonly diagnostics: vscode.DiagnosticCollection;
+    private readonly disposables: vscode.Disposable[] = [];
+
+    constructor () {
+        this.diagnostics = vscode.languages.createDiagnosticCollection("VSCode workspace settings diagnostics");
```
DavyLandman: let's use a less generic string that tells a user it came from the rascal extension.

rascal-vscode-extension/src/ux/VsCodeSettingsFixer.ts
```diff
@@ -0,0 +1,206 @@
...
+        const d = new vscode.CodeAction("Exclude target from search", vscode.CodeActionKind.QuickFix);
+        d.edit = new vscode.WorkspaceEdit();
+        d.edit.createFile(settingsUri, { contents: Buffer.from(DEFAULT_SETTINGS) });
+        d.diagnostics = [diag];
+        return d;
+    }
+
+    private fixTargetExclude(settingsDoc: vscode.TextDocument, diag: vscode.Diagnostic): vscode.CodeAction {
+        const d = new vscode.CodeAction("Exclude target from search", vscode.CodeActionKind.QuickFix);
+        const edits = jsonc.modify(settingsDoc.getText(), ["search.exclude", "/target/"], true, {});
```
DavyLandman: Does this handle the case of other entries in `search.exclude`?

DavyLandman (changes requested):

PR #967, opened 2026-02-11
Unify VFS interfaces

rodinaarssen:


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IRascalFileSystemServices.java
```diff
@@ -161,463 +159,103 @@ default CompletableFuture<FileStat> stat(URIParameter uri) {
...
-        public boolean isCreate() {
-            return create;
-        }
-
-        public boolean isOverwrite() {
-            return overwrite;
-        }
-    }
+    @Override
+    public void onDidChangeFile(FileChangeEvent event) { }
```
toinehartman: Is this supposed to have an emtpy body? Shouldn't we push this down to the implementing class?

rascal-vscode-extension/src/fs/VSCodeURIResolver.ts
```diff
@@ -272,51 +250,42 @@ class ResolverClient implements VSCodeResolverServer, Disposable  {
...
+    async readFile(loc: ISourceLocation): Promise<string> {
+        this.logger.trace("[VFS] readFile: ", loc);
+        return asyncCatcher(async () => Buffer.from(await this.fs.readFile(this.toUri(loc))).toString("base64"));
     }
 
-    isRascalNative(req: ISourceLocationRequest | vscode.Uri) : boolean {
+    isRascalNative(loc: ISourceLocation | vscode.Uri) : boolean {
         //this.rascalNativeSchemes.has(uri.scheme)
-        const scheme = "scheme" in req ? req.scheme : req.uri.substring(0, req.uri.indexOf(":"));
+        const scheme = typeof(loc) === "string" ? loc.substring(0, loc.indexOf(":")) : loc.scheme;
```
toinehartman: I think this needs a comment that `ISourceLocation == URI == string`

toinehartman (approved):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -307,15 +311,11 @@ public void connect(LanguageClient client) {
             var actualClient = (IBaseLanguageClient) client;
             lspDocumentService.connect(actualClient);
             lspWorkspaceService.connect(actualClient);
+            provideClient(actualClient);
```
toinehartman: This is called `connect` in some other places (`IBaseTextDocumentService` and `BaseWorkspaceService` for example.)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IBaseLanguageServerExtensions.java
```diff
@@ -32,35 +32,36 @@
...
     default CompletableFuture<Two<String, URI[]>[]> supplyPathConfig(PathConfigParameter projectFolder) {
         throw new UnsupportedOperationException();
     }
 
-    @JsonNotification("rascal/vfs/register")
-    void registerVFS(VFSRegister registration);
-
-    @JsonNotification("rascal/logLevel")
+    @JsonNotification("logLevel")
     void setMinimumLogLevel(String level);
```
toinehartman, line 63: Not part of your work but perhaps this should just become `logLevel`?

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -0,0 +1,193 @@
...
+    sendRequest<R, A>(uri: vscode.Uri | string, method: string, param: A): Promise<R>;
+    sendRequest<R, A>(uri: vscode.Uri | string, method: string, param?: A): Promise<R> {
+        return this.client.sendRequest<R>(method, param ?? <ISourceLocationRequest>{ loc: this.toRascalUri(uri) } )
+            .catch((r: ResponseError) => {
+                throw RemoteIOError.translateResponseError(r, uri, this.logger);
+            });
+    }
+
+    /**
+     * Attemptes to register all schemes.
```
toinehartman:
```suggestion
     * Attempts to register all schemes.
```

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -0,0 +1,193 @@
...
+     * @param schemes The list of schemes to register for this provider
+     */
+    tryRegisterSchemes(schemes: string[]) {
+        schemes
+            .filter(s => !this.protectedSchemes.includes(s))
+            // we add support for schemes that look inside a jar
+            .concat(schemes
+                .filter(s => s !== "jar" && s !== "zip" && s !== "compressed")
```
toinehartman, line 162: `compressed:`? Never seen that before. How can that be used?

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -0,0 +1,193 @@
...
+                size: a.size,
+                permissions: a.isWritable ? undefined : vscode.FilePermission.Readonly
+            }
+        );
+    }
+
+    readDirectory(uri: vscode.Uri): [string, vscode.FileType][] | Thenable<[string, vscode.FileType][]> {
+        this.logger.trace("[RascalFileSystemInVSCode] readDirectory: ", uri);
+        return this.sendRequest<DirectoryListingResponse>(uri, "rascal/vfs/input/list")
+            .then(c => c.entries.map(ft => [ft.name, ft.types.reduce((a, i) => a | i)]));
```
toinehartman: Why the `reduce`?

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -0,0 +1,514 @@
...
+        toIgnore.forEach(v => this.rascalNativeSchemes.add(v));
+    }
+}
+
+class ResolverClient implements VSCodeResolverServer, Disposable  {
+    private readonly connection: rpc.MessageConnection;
+    private readonly watchListener: WatchEventReceiver;
+    private readonly fs: vscode.FileSystem;
+    private readonly rascalNativeSchemes: Set<string>;
+    private toClear: Disposable[] = [];
```
toinehartman: nit: This is typically called `disposables`.

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -0,0 +1,514 @@
...
+    }
+    async mkDirectory(req: ISourceLocationRequest): Promise<void> {
+        this.logger.trace("[VSCodeFileSystemInRascal] mkDirectory: ", req.loc);
+        return this.asyncVoidCatcher(this.fs.createDirectory(this.toUri(req.loc)));
+    }
+    async remove(req: RemoveRequest): Promise<void> {
+        this.logger.trace("[VSCodeFileSystemInRascal] remove: ", req.loc);
+        return this.asyncVoidCatcher(this.fs.delete(this.toUri(req.loc), { recursive: req.recursive }));
+    }
+    async rename(req: RenameRequest): Promise<void> {
```
toinehartman, line 322: nit: Can we a white line in between the couple of functions above this one?

toinehartman (approved):
Impressive amount of work again! Although I trust your local testing, I think it would be nice to see an all-green CI by using a test-only snapshot release of Rascal if possible.

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
rodinaarssen: `types` is an (admittedly mostly singleton) array of file types (e.g., file and symbolic link), which are encoded as a single bit. The `reduce` converts the array into a bitmap.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/jsonrpc/RascalFileSystemInVSCode.java
```diff
@@ -0,0 +1,130 @@
...
+ * In particular, locations from LSP are mapped to Rascal-friendly locations, and Rascal exceptions are mapped to exceptions LSP expects.
+ */
+public class RascalFileSystemInVSCode extends RascalFileSystemServices {
+    private static final Logger logger = LogManager.getLogger(RascalFileSystemServices.class);
+    private static final URIResolverRegistry reg = URIResolverRegistry.getInstance();
+    
+    @Override
+    public CompletableFuture<SourceLocationResponse> resolveLocation(ISourceLocationRequest req) {
+        logger.trace("resolveLocation: {}", req.getLocation());
+        return super.resolveLocation(new ISourceLocationRequest(Locations.toClientLocation(req.getLocation())));
```
DavyLandman: how about we introduce a small utilty function that can reconstruct a `ISourceLocationRequest` as right now the `Locations.toClientLocation` is repeated in quite some places.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/jsonrpc/RascalFileSystemInVSCode.java
```diff
@@ -0,0 +1,130 @@
...
+import org.rascalmpl.uri.remote.jsonrpc.SourceLocationResponse;
+import org.rascalmpl.uri.remote.jsonrpc.WatchRequest;
+import org.rascalmpl.uri.remote.jsonrpc.WriteFileRequest;
+import org.rascalmpl.vscode.lsp.util.locations.Locations;
+
+/**
+ * Wrapper for RascalFileSystemServices handling LSP-specifics.
+ * In particular, locations from LSP are mapped to Rascal-friendly locations, and Rascal exceptions are mapped to exceptions LSP expects.
+ */
+public class RascalFileSystemInVSCode extends RascalFileSystemServices {
```
DavyLandman: can we add a comment that explains why not all calls over overriden? it looks to me like there should be more?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileRedirector.java
```diff
@@ -0,0 +1,93 @@
...
+
+import org.checkerframework.checker.initialization.qual.UnderInitialization;
+import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
+import org.rascalmpl.uri.URIUtil;
+import org.rascalmpl.vscode.lsp.IBaseTextDocumentService;
+import org.rascalmpl.vscode.lsp.TextDocumentState;
+
+import io.usethesource.vallang.ISourceLocation;
+
+public class LSPOpenFileRedirector {
```
DavyLandman, line 47: Let's add a small javadoc that explains the class.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -119,6 +122,7 @@ private static Launcher<IBaseLanguageClient> constructLSPClient(InputStream in,
             .setOutput(out)
             .configureGson(GsonUtils.complexAsJsonObject())
             .setExecutorService(threadPool)
+            .setExceptionHandler(t -> RemoteIOError.translate((Exception) t).getResponseError())
```
DavyLandman: we should only call this if it's an `IOException` as it might already be a `RespondseException` or something else. 


I think we already have something like this to deal with exceptions that should be send to the VS Code UI for error reporting.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -326,5 +326,14 @@ public void setMinimumLogLevel(String level) {
...
             Configurator.setRootLevel(l);
         }
+
+        @Override
+        public CompletableFuture<String[]> fileSystemSchemes() {
+            var reg = URIResolverRegistry.getInstance();
+            var inputs = reg.getRegisteredInputSchemes();
+            var logicals = reg.getRegisteredLogicalSchemes();
+
+            return CompletableFuture.completedFuture(Sets.union(inputs, logicals).toArray(String[]::new));
```
DavyLandman: we should put this on the `executor`, there are some `CompletableFutureUtils` functions specifically for this. as all downstream `.map` etc calls will inherit the executor unless specified.

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -0,0 +1,193 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+import path from 'path';
+import * as vscode from 'vscode';
+import { BaseLanguageClient, ResponseError } from 'vscode-languageclient';
+import { CopyRequest, DirectoryListingResponse, FileAttributes, ISourceLocationRequest, LocationContentResponse, RemoveRequest, RenameRequest, WatchRequest, WriteFileRequest } from './VSCodeFileSystemInRascal';
+import { RemoteIOError } from './RemoteIOError';
+
+export class RascalFileSystemInVSCode implements vscode.FileSystemProvider {
+    readonly client: BaseLanguageClient;
```
DavyLandman: can this be private?

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -0,0 +1,193 @@
...
+    readonly onDidChangeFile: vscode.Event<vscode.FileChangeEvent[]> = this._emitter.event;
+    private readonly protectedSchemes: string[] = ["file", "http", "https", "unknown"];
+
+    /**
+     * Registers a single FileSystemProvider for every URI scheme that Rascal supports, except
+     * for file, http and https.
+     *
+     * @param client to use as a server for the file system provider methods
+     */
+    constructor (client: BaseLanguageClient, private readonly logger: vscode.LogOutputChannel) {
```
DavyLandman, line 53: nit: use `private readonly client` in parameter, that way it becomes a class property, without needing the annoying `this.client = client` java style thing.

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -0,0 +1,193 @@
...
+        return uriString;
+    }
+
+    sendRequest<R>(uri: vscode.Uri | string, method: string): Promise<R>;
+    sendRequest<R, A>(uri: vscode.Uri | string, method: string, param: A): Promise<R>;
+    sendRequest<R, A>(uri: vscode.Uri | string, method: string, param?: A): Promise<R> {
+        return this.client.sendRequest<R>(method, param ?? <ISourceLocationRequest>{ loc: this.toRascalUri(uri) } )
+            .catch((r: ResponseError) => {
+                throw RemoteIOError.translateResponseError(r, uri, this.logger);
+            });
```
DavyLandman: nit: can we do  `=> throw..` instead of the curly brace?

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
DavyLandman: It's for dynamically decompressing a file. It's used for large data sets for example: `readFile(|compressed+file:///big-data-set.csv.gz|)` or `readCSV(|compressed+home:///big-data-set.csv.gz|`);

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -0,0 +1,193 @@
...
+                    else {
+                        this.logger.debug(`Rascal VFS lost the race to register scheme: ${s}, which in most cases is fine`);
+                    }
+                }
+            });
+    }
+
+    watch(uri: vscode.Uri, options: { recursive: boolean; excludes: string[]; }): vscode.Disposable {
+        this.logger.trace("[RascalFileSystemInVSCode] watch: ", uri);
+        this.sendRequest(uri, "rascal/vfs/watcher/watch", <WatchRequest>{
```
DavyLandman: this promise is never awaited for, so the exception is ignored. we should turn this function into a `async` one and then do `await` on the send request.

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -0,0 +1,193 @@
...
+            const colon = uri.scheme.length + 1;
+            return `${uriString.slice(0, colon)}//${uriString.slice(colon)}`;
+        }
+        return uriString;
+    }
+
+    sendRequest<R>(uri: vscode.Uri | string, method: string): Promise<R>;
+    sendRequest<R, A>(uri: vscode.Uri | string, method: string, param: A): Promise<R>;
+    sendRequest<R, A>(uri: vscode.Uri | string, method: string, param?: A): Promise<R> {
+        return this.client.sendRequest<R>(method, param ?? <ISourceLocationRequest>{ loc: this.toRascalUri(uri) } )
```
DavyLandman: if we turn the `sendRequest` function into an `async` one, we can use `try+catch` instead of the `.catch` callback.

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -0,0 +1,193 @@
...
+            this.sendRequest(uri, "rascal/vfs/watcher/unwatch", {
+                loc: this.toRascalUri(uri),
+                recursive: options.recursive
+            });
+        });
+    }
+
+    stat(uri: vscode.Uri): vscode.FileStat | Thenable<vscode.FileStat> {
+        this.logger.trace("[RascalFileSystemInVSCode] stat: ", uri);
+        return this.sendRequest<FileAttributes>(uri, "rascal/vfs/input/stat").then(a =>
```
DavyLandman: I'll don't repeat this for all the other cases in this function, but in general it's nicer to stop using the `.then` and use the `async` functions with `await`.

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -0,0 +1,193 @@
...
+            .then(r => Buffer.from(r.content, "base64"));
+    }
+
+    writeFile(uri: vscode.Uri, content: Uint8Array, options: { create: boolean; overwrite: boolean; }): void | Thenable<void> {
+        // The `create` and `overwrite` options are handled on this side
+        this.logger.trace("[RascalFileSystemInVSCode] writeFile: ", uri);
+        const parentUri = uri.with({ path: path.dirname(uri.path) });
+        Promise.all([
+            this.sendRequest<FileAttributes>(uri, "rascal/vfs/input/stat"),
+            this.sendRequest<FileAttributes>(parentUri, "rascal/vfs/input/stat")]
```
DavyLandman: good use of `Promise.all` 👍🏼

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    readFile(uri: vscode.Uri): Uint8Array | Thenable<Uint8Array> {
+        this.logger.trace("[RascalFileSystemInVSCode] readFile: ", uri);
+        return this.sendRequest<LocationContentResponse>(uri, "rascal/vfs/input/readFile")
+            .then(r => Buffer.from(r.content, "base64"));
+    }
+
+    writeFile(uri: vscode.Uri, content: Uint8Array, options: { create: boolean; overwrite: boolean; }): void | Thenable<void> {
+        // The `create` and `overwrite` options are handled on this side
```
DavyLandman: Let's explain why.

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -0,0 +1,193 @@
...
+        this.logger.trace("[RascalFileSystemInVSCode] writeFile: ", uri);
+        const parentUri = uri.with({ path: path.dirname(uri.path) });
+        Promise.all([
+            this.sendRequest<FileAttributes>(uri, "rascal/vfs/input/stat"),
+            this.sendRequest<FileAttributes>(parentUri, "rascal/vfs/input/stat")]
+        ).then(r => {
+            const fileStat = r[0];
+            const parentStat = r[1];
```
DavyLandman: use destructuring bind `[fileState, parentState] =>` or after making this async `const [fileState, parentStat] = await Promise.all(...`

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -0,0 +1,193 @@
...
+        if (uri.authority === "" && uri.scheme !== "file") {
+            const colon = uri.scheme.length + 1;
+            return `${uriString.slice(0, colon)}//${uriString.slice(colon)}`;
+        }
+        return uriString;
+    }
+
+    sendRequest<R>(uri: vscode.Uri | string, method: string): Promise<R>;
+    sendRequest<R, A>(uri: vscode.Uri | string, method: string, param: A): Promise<R>;
+    sendRequest<R, A>(uri: vscode.Uri | string, method: string, param?: A): Promise<R> {
```
DavyLandman: why the uri as first param if there is an argument?

can we not do:

```typescript
sendRequest<R>(method: string, param: vscode.URI): Promise<R>;
sendRequest<R,A>(method: string, param: A): Promise<R>;
sendRequest<R,A>(method: string: param: vscode.URI | A): Promise<R> {
...
}
```

such that a user does not have to repeat the URI if they're providing a custom parameter object?

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -0,0 +1,193 @@
...
+            if (!fileStat.exists && !options.create) {
+                throw vscode.FileSystemError.FileNotFound(`File ${uri} does not exist and \`create\` was not set`);
+            }
+            if (!parentStat.exists && options.create) {
+                throw vscode.FileSystemError.FileNotFound(`Parent of ${uri} does not exist but \`create\` was set`);
+            }
+            if (fileStat.exists && options.create && !options.overwrite) {
+                throw vscode.FileSystemError.FileExists(`File ${uri} exists and \`create\` was set, but \`override\` was not set`);
+            }
+            return this.sendRequest(uri, "rascal/vfs/output/writeFile", <WriteFileRequest>{
```
DavyLandman: i think there is also a case for a writeFile on a directory that we could now throw. (like path is a directory or something is the exception).

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    rename(oldUri: vscode.Uri, newUri: vscode.Uri, options: { overwrite: boolean; }): void | Thenable<void> {
+        this.logger.trace("[RascalFileSystemInVSCode] rename: ", oldUri, newUri);
+        return this.sendRequest(oldUri, "rascal/vfs/output/rename", <RenameRequest>{ from: this.toRascalUri(oldUri), to: this.toRascalUri(newUri), overwrite: options.overwrite });
+    }
+
+    copy(source: vscode.Uri, target: vscode.Uri, options?: { overwrite?: boolean; }): Thenable<void> {
+        this.logger.trace("[RascalFileSystemInVSCode] copy: ", source, target);
+        return this.sendRequest(source, "rascal/vfs/output/copy", <CopyRequest>{ from: this.toRascalUri(source), to: this.toRascalUri(target), recursive: true, overwrite: (options && options.overwrite) ?? false });
```
DavyLandman: I think we might be able to use `overwrite: options?.overwrite ?? false` as the `.?` just forwards the undefined

rascal-vscode-extension/src/fs/RemoteIOError.ts
```diff
@@ -0,0 +1,93 @@
...
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+import * as vscode from 'vscode';
+import * as rpc from 'vscode-jsonrpc/node';
+import { ResponseError } from "vscode-languageclient";
+
+export class RemoteIOError {
```
DavyLandman, line 35: add a comment that this mirrors the definitions in the rascal enum (and mention it)

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -0,0 +1,516 @@
...
+ */
+interface VSCodeResolverServer extends ISourceLocationInput, ISourceLocationOutput, ISourceLocationWatcher { }
+
+/**
+ * Rascal side should implement this on the other side of the stream
+ */
+//interface VSCodeResolverClient extends WatchEventReceiver {}
+
+
+// Rascal's interface reduced to a subset we can support
```
DavyLandman: is this still true? this comment I mean, can we explain why we cannot support something?

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -0,0 +1,516 @@
...
+
+function buildWatchReceiver(connection: rpc.MessageConnection): WatchEventReceiver {
+    return {
+        emitWatch: (e) => {
+            connection.sendNotification(new rpc.NotificationType1<ISourceLocationChanged>("rascal/vfs/watcher/sourceLocationChanged"), e);
+        }
+    };
+}
+
+// Messages (requests and responses)
```
DavyLandman: I think these message classes should be moved to a seperate file (that contains them all) such that both VFS implementations can import that file.

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -0,0 +1,516 @@
...
+    entries: DirectoryEntry[]
+}
+
+export interface DirectoryEntry {
+    name: string;
+    types: vscode.FileType[]
+}
+
+export class VSCodeFileSystemInRascal extends JsonRpcServer {
+    private rascalNativeSchemes: Set<string> = new Set();
```
DavyLandman, line 132: can be readonly.

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -0,0 +1,516 @@
...
+}
+
+class ResolverClient implements VSCodeResolverServer, Disposable  {
+    private readonly connection: rpc.MessageConnection;
+    private readonly watchListener: WatchEventReceiver;
+    private readonly fs: vscode.FileSystem;
+    private readonly rascalNativeSchemes: Set<string>;
+    private disposables: Disposable[] = [];
+    constructor(connection: rpc.MessageConnection, debug: boolean, rascalNativeSchemes: Set<string>, private readonly logger: vscode.LogOutputChannel){
+        this.rascalNativeSchemes = rascalNativeSchemes;
```
DavyLandman: move these copies (same for `connection`) to the parameter definition list of the constructor.

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -0,0 +1,516 @@
...
+            else {
+                return run;
+            }
+        });
+    }
+
+    toUri(loc: ISourceLocation): vscode.Uri {
+        const uri = vscode.Uri.parse(loc);
+        if (this.isRascalNative(uri)) {
+            throw new rpc.ResponseError(RemoteIOError.isRascalNative, `Cannot request VS jobs on native Rascal locations: ${loc}`);
```
DavyLandman: `VS jobs` is a strange term?

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -0,0 +1,516 @@
...
+    async fileStat(req: ISourceLocationRequest): Promise<FileAttributes> {
+        return this.asyncCatcher(async () => {
+            const fileInfo = await this.stat(req.loc);
+            return {
+                exists: true,
+                isFile: (fileInfo.type | vscode.FileType.File) > 0,
+                created: fileInfo.ctime,
+                lastModified: fileInfo.mtime,
+                isReadable: true,
+                isWritable: true,
```
DavyLandman: `isWriteable` should depend on the `fileInfo.permissions`

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -0,0 +1,516 @@
...
+        this.logger.trace("[VSCodeFileSystemInRascal] isWritable: ", req.loc);
+        const scheme = this.toUri(req.loc).scheme;
+        const writable = this.fs.isWritableFileSystem(scheme);
+        if (writable === undefined) {
+            throw new rpc.ResponseError(RemoteIOError.unsupportedScheme, `Unsupported scheme: ${scheme}`);
+        }
+        if (!writable) {
+            // not a writable file system, so no need to check the uri
+            return { value: false };
+        }
```
DavyLandman, line 281: well done 👍🏼

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -0,0 +1,516 @@
...
+
+    decodeFileTypeBitmask(input: number): vscode.FileType[] {
+        return input === 0 ? [vscode.FileType.Unknown] : [vscode.FileType.File, vscode.FileType.Directory, vscode.FileType.SymbolicLink].filter(t => t === (t & input));
+    }
+
+    async writeFile(req: WriteFileRequest): Promise<void> {
+        this.logger.trace("[VSCodeFileSystemInRascal] writeFile: ", req.loc);
+        const loc = this.toUri(req.loc);
+        let prefix: Buffer<ArrayBuffer> = Buffer.of();
+        if (await this.exists(req) && req.append) {
```
DavyLandman: flip the arguments, such that we check `req.append` first, that way we skip an extra `stat`

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -0,0 +1,516 @@
...
+        this.logger.trace("[VSCodeFileSystemInRascal] rename: ", req.from, req.to);
+        return this.asyncVoidCatcher(this.fs.rename(this.toUri(req.from), this.toUri(req.to), { overwrite: req.overwrite }));
+    }
+
+    async copy(req: CopyRequest): Promise<void> {
+        this.logger.trace("[VSCodeFileSystemInRascal] copy: ", req.from, req.to);
+        if (req.recursive && await this.isDirectory({ loc: req.from })) {
+            throw new rpc.ResponseError(RemoteIOError.isADirectory, 'Non-recursive copy requested on a directory', req);
+        }
+        return this.asyncVoidCatcher(this.fs.copy(this.toUri(req.from), this.toUri(req.to), { overwrite: req.overwrite }));
```
DavyLandman, line 332: shouldn't the `recursive` flag be passed along, or something done with it?

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -0,0 +1,516 @@
...
+    }
+
+    private readonly activeWatches = new Map<string, WatcherCallbacks>();
+
+    async watch(newWatch: WatchRequest): Promise<void> {
+        this.logger.trace("[VSCodeFileSystemInRascal] watch: ", newWatch.loc);
+        const watchKey = newWatch.loc + newWatch.recursive;
```
DavyLandman: we should a charcter between the loc and the boolean, such that we never get colissions. for example:

```typescript
const watchKey = `${newWatch.loc}##${newWatch.recursive}`;
```

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -0,0 +1,516 @@
...
+        if (!this.activeWatches.has(watchKey)) {
+            const watcher = new WatcherCallbacks(this.toUri(newWatch.loc), newWatch.recursive, this.watchListener, newWatch.watchId);
+            this.activeWatches.set(watchKey, watcher);
+            this.disposables.push(watcher);
+            return;
+        }
+        throw new rpc.ResponseError(RemoteIOError.watchAlreadyDefined, `Watch already defined: ${newWatch.loc}`);
```
DavyLandman: this only works because we provide some guarantees of never being called twice. I think we should document some of that on the java classes.

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -0,0 +1,516 @@
...
+        }
+        return this.asyncVoidCatcher(this.fs.copy(this.toUri(req.from), this.toUri(req.to), { overwrite: req.overwrite }));
+    }
+
+    private readonly activeWatches = new Map<string, WatcherCallbacks>();
+
+    async watch(newWatch: WatchRequest): Promise<void> {
+        this.logger.trace("[VSCodeFileSystemInRascal] watch: ", newWatch.loc);
+        const watchKey = newWatch.loc + newWatch.recursive;
+        if (!this.activeWatches.has(watchKey)) {
```
DavyLandman: nit: move the throw inside the if, and the normal flow outside of it (same for the `unwatch` function).

rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -266,15 +266,14 @@ export class RascalExtension implements vscode.Disposable {
...
         }
         shellArgs.push();
+        const vfsServerPort = await this.vfsServer.serverPort;
         shellArgs.push(
             '-Dfile.encoding=UTF8'
-            , '-Drascal.fallbackResolver=org.rascalmpl.vscode.lsp.uri.FallbackResolver'
             , `-Drascal.languageRegistryPort=${await this.rascal.languageRegistry.serverPort}`
-            , 'org.rascalmpl.vscode.lsp.terminal.LSPTerminalREPL'
+            , `-Drascal.remoteResolverRegistryPort=${vfsServerPort}`
+            , 'org.rascalmpl.shell.RascalShell'
```
DavyLandman, line 274: 🎉

rascal-vscode-extension/src/RascalTerminalLinkProvider.ts
```diff
@@ -89,6 +89,27 @@ export class RascalTerminalLinkProvider implements TerminalLinkProvider<Extended
             );
         }
     }
+
+    toRascalLocationString(sloc: SourceLocation): string {
+        let ret = "|" + sloc.uri + "|";
```
DavyLandman: nit: use typescript/js string templating

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
rodinaarssen: In VS Code, `copy` is always recursive

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
rodinaarssen: Yes, that works as well

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
rodinaarssen: Shouldn't that be handled by Rascal? The fiddling around with `create` and `overwrite` is to comply with what VS Code expects us to do, not to eagerly handle potential error situations

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
rodinaarssen: We cannot, that's not syntactically valid

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/jsonrpc/RascalFileSystemInVSCode.java
rodinaarssen: I've updated this

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
DavyLandman: Okay, fair.

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
DavyLandman: So what if we pass recursieve as false? We have to honor that right?

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
DavyLandman: Ah okay, but making this into an async function would also make this nicer to read, and this closure goes away.

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
rodinaarssen: NB: the fragment that's shown here in the web interface is outdated somehow: I've flipped `req.recursive`.

We in fact do: if `!req.recursive` there are two cases: either it's a directory and we throw an error, or it's a file and it's fine.

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
rodinaarssen: How could there a collision here?

DavyLandman: Hmmmm, I was thinking about a file named true or something, but that's not right, it'll still het the boolean suffix.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/jsonrpc/RascalFileSystemInVSCode.java
rodinaarssen: The other methods were simply missing. Thanks

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
rodinaarssen: `watch` cannot be made `async` as it inherits its signature from `vscode.FileSystemProvider`. Exceptions are now caught.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
rodinaarssen: I've special-cased `ResponseErrorException`; everything else is considered a remote IO error

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/jsonrpc/RascalFileSystemInVSCode.java
```diff
@@ -0,0 +1,215 @@
...
+import org.rascalmpl.uri.remote.jsonrpc.WriteFileRequest;
+import org.rascalmpl.vscode.lsp.util.locations.Locations;
+
+import io.usethesource.vallang.ISourceLocation;
+
+/**
+ * Wrapper for RascalFileSystemServices handling LSP-specifics.
+ * In particular, locations from LSP are mapped to Rascal-friendly locations.
+ */
+public class RascalFileSystemInVSCode extends RascalFileSystemServices {
```
DavyLandman: Since we don't want to mis an override, what if we let this implement the interface, and then have the `RascalFileSystemServices` as a private field?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -122,6 +126,12 @@ private static Launcher<IBaseLanguageClient> constructLSPClient(InputStream in,
             .setOutput(out)
             .configureGson(GsonUtils.complexAsJsonObject())
             .setExecutorService(threadPool)
+            .setExceptionHandler(t -> {
+                if (t instanceof ResponseErrorException) {
+                    return ((ResponseErrorException) t).getResponseError();
+                }
+                return RemoteIOError.translate(t).getResponseError();
```
DavyLandman, line 133: shouldn't we make sure to only do this for an `t instanceof IOException`?

rascal-vscode-extension/src/fs/JsonRpcMessages.ts
```diff
@@ -0,0 +1,129 @@
...
+export declare type ISourceLocation = URI;
+
+export interface JsonRpcRequest { }
+
+export interface ISourceLocationRequest extends JsonRpcRequest {
+    loc: ISourceLocation;
+}
+
+export interface WriteFileRequest extends ISourceLocationRequest {
+    content: string;
```
DavyLandman, line 42: maybe this should share something with `LocationContentResponse`? Otherwise repeat the base64 comment?

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -0,0 +1,277 @@
...
+            });
+    }
+
+    watch(uri: vscode.Uri, options: { recursive: boolean; excludes: string[]; }): vscode.Disposable {
+        this.logger.trace("[RascalFileSystemInVSCode] watch: ", uri);
+        const watchKey: [vscode.Uri, boolean] = [uri, options.recursive];
+        void this.sendRequest("rascal/vfs/watcher/watch", <WatchRequest>{
+            loc: RascalFileSystemInVSCode.toRascalUri(uri),
+            recursive: options.recursive,
+            watchId: this.watchId
```
DavyLandman: I was thinking, we should add at least the loc to the watchid. since we shouldn't depend on the semantics of it being fine to pass the same watchid for all the calls.

so something like `${uri}-${recursive}-${watchId}`

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -0,0 +1,277 @@
...
+                }
+
+                // A non-recursive watch applies to a file, or a directory's direct children
+                if (!recursive && uri.path !== eventUri.path && this.ensureTrailingSlash(uri.path) !== this.ensureTrailingSlash(path.dirname(eventUri.path))) {
+                    // Current watch does not apply to the event uri
+                    continue;
+                }
+
+                // A recursive watch applies to a directory's children at arbitrary depth
+                if (recursive && !eventUri.path.startsWith(this.ensureTrailingSlash(uri.path))) {
```
DavyLandman: well done with the ensuring of trailing slash!

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -0,0 +1,277 @@
...
+                    // Current watch does not apply to the event uri
+                    continue;
+                }
+
+                // Current watch does apply to the event uri; checking whether it is excluded in this watch
+                for (const exclude of excludes) {
+                    const isAbsolute = path.isAbsolute(exclude);
+                    const isGlob = exclude.indexOf("*") + exclude.indexOf("?") + exclude.indexOf("[") + exclude.indexOf("{") !== -4;
```
DavyLandman: since the excludes are constant, let's calculate these properties once, instead of on every watch event.

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -0,0 +1,277 @@
...
+                    if (isAbsolute && this.excludeMatchesUri(eventUri.path, exclude, isGlob)
+                        || !isAbsolute && this.excludeMatchesUri(eventUri.path, path.join(uri.path, exclude), isGlob)) {
+                        // Event uri was excluded in current watch
+                        continue watches;
+                    }
+                }
```
DavyLandman: this loop never seems to set some kind of state that says: hey this event should be excluded?

rascal-vscode-extension/src/fs/TestVirtualFileSystem.ts
```diff
@@ -90,9 +144,62 @@ export class TestVirtualFileSystem implements vscode.FileSystemProvider, vscode.
...
+                this.logger.info(`Exclude ${exclude}`);
+                const isAbsolute = path.isAbsolute(exclude);
+                const isGlob = exclude.indexOf("*") + exclude.indexOf("?") + exclude.indexOf("[") + exclude.indexOf("{") !== -4;
+                if (isAbsolute && this.excludeMatchesUri(targetUri.path, exclude, isGlob)
+                    || !isAbsolute && this.excludeMatchesUri(targetUri.path, path.join(uri.path, exclude), isGlob)) {
+                    // Event uri was excluded in current watch
+                    this.logger.info(`Excluded, continue to next watch`);
+                    continue watches;
+                }
+            }
```
DavyLandman: same here, no side-effect of the watches?

Also can we not extract this watch logic into class we can share between the 2 VFS we have?

rascal-vscode-extension/src/fs/TestVirtualFileSystem.ts
```diff
@@ -90,9 +144,62 @@ export class TestVirtualFileSystem implements vscode.FileSystemProvider, vscode.
...
+        this.activeWatches.set(watchKey, options.excludes);
+
+        return new vscode.Disposable(async () => {
+            this.logger.info("[TVFS] Watch disposed");
+            this.activeWatches.delete(watchKey);
+        });
+    }
+
+    notifyWatchers(targetUri: vscode.Uri, type: vscode.FileChangeType) {
+        this.logger.info("[TVFS] notifyWatchers ", targetUri);
```
DavyLandman: should be debug (and more in this class)

rascal-vscode-extension/src/fs/TestVirtualFileSystem.ts
```diff
@@ -113,6 +220,7 @@ export class TestVirtualFileSystem implements vscode.FileSystemProvider, vscode.
 
     createDirectory(uri: vscode.Uri) {
         this.logger.debug("[TVFS] createDirectory: ", uri);
+        this.notifyWatchers(uri, vscode.FileChangeType.Created);
```
DavyLandman: I think the watch should fire after it has been done (same for the other functions in this class)

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -0,0 +1,456 @@
...
+    getCharset(req: ISourceLocationRequest): Promise<StringResponse>;
+}
+
+function connectInputHandler(connection: rpc.MessageConnection, handler: ISourceLocationInput, toClear: Disposable[]) {
+    function req<T> (method: string, h: rpc.RequestHandler1<ISourceLocationRequest, T, void>) {
+        toClear.push(connection.onRequest(
+            new rpc.RequestType1<ISourceLocationRequest, T, void>("rascal/vfs/input/" + method),
+            h.bind(handler)));
+    }
+    req<LocationContentResponse>("readFile", handler.readFile);
```
DavyLandman: why can we not use these nice definitions of the API for both sides? (I mean `RascalFileSystemInVSCode` )

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -0,0 +1,456 @@
...
+        this.logger.trace("[VSCodeFileSystemInRascal] created: ", req.loc);
+        return this.numberResult(req.loc, f => f.ctime);
+    }
+
+    size(req: ISourceLocationRequest): Promise<NumberResponse> {
+        this.logger.trace("[VSCodeFileSystemInRascal] size: ", req.loc);
+        return this.numberResult(req.loc, f => f.size);
+    }
+
+    private async boolResult(loc: ISourceLocation, mapper: (s: vscode.FileStat) => boolean): Promise<BooleanResponse> {
```
DavyLandman: I got confused by the `boolResult`  name, as it actually runs `stat`

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -0,0 +1,456 @@
...
+        this.activeWatches.set(req.watchId, watcher);
+        this.disposables.push(watcher);
+    }
+
+    async unwatch(req: WatchRequest): Promise<void> {
+        this.logger.info("[VSCodeFileSystemInRascal] unwatch: ", req.loc);
+        this.logger.info("Active watches:");
+        for (const watch of this.activeWatches) {
+            this.logger.info(`${watch}`);
+        }
```
DavyLandman: looks like a debug print

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -0,0 +1,456 @@
...
+        const watcher = this.activeWatches.get(req.watchId);
+        this.logger.info(`watcher: ${watcher}`);
+        if (watcher === undefined) {
+            this.logger.info("UNWATCH: undefined watcher");
+            throw new rpc.ResponseError(RemoteIOError.watchNotDefined, `Watch not defined: ${req.loc}`);
+        }
+        this.activeWatches.delete(req.watchId);
+        this.logger.info("Deleted");
+        this.logger.info("Active watches:");
```
DavyLandman: more debug prints

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -0,0 +1,456 @@
...
+        watcher.dispose();
+        const index = this.disposables.indexOf(watcher);
+        if (index >= 0) {
+            this.disposables.splice(index, 1);
+        }
+    }
+
+    async resolve(req: ISourceLocationRequest): Promise<SourceLocationResponse> {
+        this.logger.trace("[VSCodeFileSystemInRascal] resolve: ", req.loc);
+        return <SourceLocationResponse>{ loc: undefined };
```
DavyLandman: this is not implemented? maybe add a comment to why?

DavyLandman (approved):
This has been quite a journey. But yes, only a few small remarks.

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
rodinaarssen: Oops.

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
rodinaarssen: Oops

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
rodinaarssen: Maybe I'm not understanding your comment correctly, but this `continue` does exactly that: the file the event is reporting on matches the current active watch, but it's excluded. As there may be another registered watch that matches this file, the iteration over the registered watches is continued.

DavyLandman: I misread the indentation/bracket of the `.fire` code.

rascal-vscode-extension/src/fs/TestVirtualFileSystem.ts
rodinaarssen: The log levels have been unified

rascal-vscode-extension/src/fs/TestVirtualFileSystem.ts
rodinaarssen: That is a good idea, thanks.

rascal-vscode-extension/src/fs/TestVirtualFileSystem.ts
rodinaarssen: The Test VFS now reuses the code from RascalFileSystemInVSCode

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
rodinaarssen: Done

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
rodinaarssen: The exception handler has to return a `ResponseError`. The `translate` method differentiates between different types of exceptions, of which not everything even inherits from `IOException`.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/jsonrpc/RascalFileSystemInVSCode.java
rodinaarssen: Good idea. In fact, I missed one.

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
rodinaarssen: In `RascalFileSystemInVSCode` the relevant API is `vscode.FileSystemProvider`, whereas here the API is modeled after the resolvers in Rascal.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/jsonrpc/RascalFileSystemInVSCode.java
```diff
@@ -0,0 +1,240 @@
...
+import org.rascalmpl.vscode.lsp.util.locations.Locations;
+
+import io.usethesource.vallang.ISourceLocation;
+
+/**
+ * Wrapper for RascalFileSystemServices handling LSP-specifics.
+ * In particular, locations from LSP are mapped to Rascal-friendly locations.
+ */
+public class RascalFileSystemInVSCode implements IRemoteResolverRegistryServer {
+    private static final Logger logger = LogManager.getLogger(RascalFileSystemServices.class);
```
DavyLandman: wrong class for the `getLogger` argument.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
DavyLandman: ah, right. okay.

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    static isGlob(exclude: string): boolean {
+        return exclude.indexOf("*") + exclude.indexOf("?") + exclude.indexOf("[") + exclude.indexOf("{") !== -4;
+    }
+
+    async stat(uri: vscode.Uri): Promise<vscode.FileStat> {
+        this.logger.debug("[RascalFileSystemInVSCode] stat: ", uri);
+        const stat = await this.inputClient.stat(uriToRequest(uri));
+        return <vscode.FileStat>{
```
DavyLandman: nit: no need to repeat the return type

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -0,0 +1,412 @@
...
+        return this.numberResult(req.loc, f => f.ctime);
+    }
+
+    size(req: ISourceLocationRequest): Promise<NumberResponse> {
+        this.logger.debug("[VSCodeFileSystemInRascal] size: ", req.loc);
+        return this.numberResult(req.loc, f => f.size);
+    }
+
+    private async boolResultFromStat(loc: ISourceLocation, mapper: (s: vscode.FileStat) => boolean): Promise<BooleanResponse> {
+        return this.asyncCatcher(async () => <BooleanResponse>{ value: mapper((await this.internalStat(loc))) });
```
DavyLandman: nit: I think the `<BooleanResponse>` can be dropped.

DavyLandman (approved):

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
rodinaarssen: This has been redesigned. Good thinking

PR #969, opened 2026-02-13
Fix go-to-def test

toinehartman:
1. Use 1-based `selectText` index (instead of 0-based).
2. Use name on multiple lines to work around https://github.com/redhat-developer/vscode-extension-tester/issues/2254.
3. Add assertion to verify jumping to use before jumping back to def.


rodinaarssen (approved):
Perhaps we should add a comment that the test might start failing if `IO.rsc` is changed. Other than that: LGTM

PR #989, opened 2026-02-28
Neovim plugin

qu1ncyk:
Resolves #988.

This PR adds a Neovim plugin that makes it easy to use the Rascal LSP in Neovim. To make the LSP work in Neovim, this PR includes fixes for a NullPointerException.
<details>
<summary>NullPointerException stack trace</summary>

```
java.util.concurrent.CompletionException: java.lang.NullPointerException: Cannot invoke "org.eclipse.lsp4j.FileOperationsWorkspaceCapabilities.getDidCreate()" because the return value of "org.eclipse.lsp4j.WorkspaceClientCapabilities.getFileOperations()" is null
    at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315)
    at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:320)
    at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1770)
    at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
    at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
    at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: java.lang.NullPointerException: Cannot invoke "org.eclipse.lsp4j.FileOperationsWorkspaceCapabilities.getDidCreate()" because the return value of "org.eclipse.lsp4j.WorkspaceClientCapabilities.getFileOperations()" is null
    at org.rascalmpl.vscode.lsp.rascal.RascalWorkspaceService.initialize(RascalWorkspaceService.java:61)
    at org.rascalmpl.vscode.lsp.BaseLanguageServer$ActualLanguageServer.lambda$initialize$4(BaseLanguageServer.java:260)
    at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1768)
    ... 3 more
```

</details>

To be honest, I'm not sure what `getDidCreate`, `getDidRename` and `getDidDelete` do and whether it is necessary to invoke `getFileOperations` 3 times, but at least the error is gone.

This plugin only adds support for the Rascal LSP, not for the parametric LSP.

PR #996, opened 2026-03-01
Fixed a `NullPointerException` that would occur in Neovim

qu1ncyk:
See #989.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
@@ -62,11 +62,9 @@ public class RascalWorkspaceService extends BaseWorkspaceService {
         if (operations != null && operations.getDidCreate().booleanValue()) {
             fileOperationCapabilities.setDidCreate(whichFiles);
         }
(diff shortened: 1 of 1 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
```diff
@@ -58,13 +58,16 @@ public void initialize(ClientCapabilities clientCap, @Nullable List<WorkspaceFol
         }
         var fileOperationCapabilities = capabilities.getWorkspace().getFileOperations();
         var whichFiles = new FileOperationOptions(fileFilters);
-        if (clientCapabilities.getFileOperations().getDidCreate().booleanValue()) {
+        var operations = clientCapabilities.getFileOperations();
+        if (operations != null && operations.getDidCreate().booleanValue()) {
             fileOperationCapabilities.setDidCreate(whichFiles);
         }
-        if (clientCapabilities.getFileOperations().getDidRename().booleanValue()) {
+        operations = clientCapabilities.getFileOperations();
```
DavyLandman: I think we don't need to re-request the `getFileOperations`, yes it's the more correct refactoring, but it doesn't make sense.

DavyLandman (approved):

PR #1003, opened 2026-03-04
Fix TModel race between type-checker and renaming

toinehartman:
This PR changes the distribution of tasks over evaluators, moving renaming to the compiler evaluator, since renaming and compilation might otherwise both run the type-checker in parallel on the same module.

**Pro** No more races for type-checking a certain module.
**Con** Renaming might need to wait for an unrelated, long-running, type-check.


DavyLandman:
how bad is this race? they would be sharing the same bin folder, so they should see each-other can catch up?

PR #1009, opened 2026-03-09
Nullability fixes

toinehartman:
This PR fixes nullability bugs by using CF [`3.54.0`](https://github.com/typetools/checker-framework/releases/tag/checker-framework-3.54.0). However, since it contains some bugs as well (inference crashes on some of our code), this PR does not actually update the dependency.


toinehartman:
These changes are mostly annotation changes, to check against newer CF versions. Let's reconsider this once there is a stable version that we'll use.

PR #1012, opened 2026-03-10
Rename independent of `util::LanguageServer`

toinehartman:
Move away from needing any `@javaClass`-annotated functions from LSP during renaming, simplifying testing in https://github.com/usethesource/rascal-core-big-tests/pull/5. This also helps with separating the renaming implementation from the language server interface and cleaning up `util::LanguageServer`.


DavyLandman (commented):
I like the idea, but have some small remarks.

PR #1026, opened 2026-03-17
Refactor computation of facts for readability

toinehartman:
Semantics-preserving refactoring of the code in #904.


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/FileFacts.java
```diff
@@ -60,6 +60,7 @@ public class FileFacts {
     private final Map<ISourceLocation, FileFact> files = new ConcurrentHashMap<>();
     private final ColumnMaps cm;
     private final PathConfigs confs;
+    private final NopFileFact nopFact = new NopFileFact();
```
DavyLandman: maybe instead the `NopFileFact` class should be replaced by an anonymous class that get's assigned to the static field `NOPE_FACT` or something?

PR #1032, opened 2026-03-19
Test IValue serialization as JSON

### LINKED ISSUE #1029: Test missing for typescript calling rascal actions
We should make a test that calls a pico action is called from typescript, and that the json result doesn't change, even for complexer ADTs that are going over the line.

A recent regression on this (fixed in  #1028) would have been caught if we tested this interaction.

toinehartman:
This is a little nasty to test, since we need to get hold of values returned from LSP commands. VS Code Extension Tester only allows us to access commands from the command palette, for which we need to add them to the `package.json`. This command calls the LSP command and is only available during testing, but unfortunately will be shipped in production code.

Additionally, we cannot access command return values. Hence, we open an editor with the returned result, which we can then access from the test. Since the expected value is pretty big, we store that in a file in the test directory.

Fixes #1029.


rodinaarssen (approved):

DavyLandman (changes requested):
This is great! I have a few small remarks.

DavyLandman (approved):

PR #1040, opened 2026-03-31
Use slightly longer delays for Ubuntu tests

toinehartman:
After changing Ubuntu machine images in #1034 and #1037, we are facing flaky UI tests. This PR attempts to fix this.


toinehartman:
This experiment led to the following fixes:
- #1055 
- #1056

Issue #968, opened 2026-02-11
Path config computation fails for modules in `std://`


toinehartman:
**Describe the bug**

Sometimes, the path config is computed with a files (instead of folder) as the project root. That leads to bogus configs and subsequent failures in computations that use the configs. Example:

```
2026-02-10 13:48:07.358 [debug] [rascal-lsp-1] org.rascalmpl.vscode.lsp.rascal.model.PathConfigs Building pcfg from: |std:///IO.rsc| (@1770731287350 ms)
2026-02-10 13:48:07.358 [debug] [rascal-lsp-1] org.rascalmpl.vscode.lsp.rascal.model.PathConfigs New path config: Path configuration items:
projectRoot:|jar+file:///C:/Users/runneradmin/.vscode/extensions/usethesource.rascalmpl-0.13.1-head/assets/jars/rascal.jar!/org/rascalmpl/library/IO.rsc|
srcs:       [|jar+file:///C:/Users/runneradmin/.vscode/extensions/usethesource.rascalmpl-0.13.1-head/assets/jars/rascal.jar!/org/rascalmpl/library/IO.rsc/src|]
ignores:    []
libs:       [|jar+file:///C:/Users/runneradmin/.vscode/extensions/usethesource.rascalmpl-0.13.1-head/assets/jars/rascal.jar!/|,|file:///C:/Users/runneradmin/.vscode/extensions/usethesource.rascalmpl-0.13.1-head/assets/jars/rascal-lsp.jar|,|jar+file:///C:/Users/runneradmin/.vscode/
extensions/usethesource.rascalmpl-0.13.1-head/assets/jars/rascal-lsp.jar!/|]
bin:        |jar+file:///C:/Users/runneradmin/.vscode/extensions/usethesource.rascalmpl-0.13.1-head/assets/jars/rascal.jar!/org/rascalmpl/library/IO.rsc/target/classes|
resources:  []
messages:   [error("Source folder |jar+file:///C:/Users/runneradmin/.vscode/extensions/usethesource.rascalmpl-0.13.1-head/assets/jars/rascal.jar!/org/rascalmpl/library/IO.rsc/src| does not exist.",|jar+file:///C:/Users/runneradmin/.vscode/extensions/usethesource.rascalmpl-0.13.1-head/
assets/jars/rascal.jar!/org/rascalmpl/library/IO.rsc/META-INF/RASCAL.MF|)]
 (@1770731287356 ms)
```

**To Reproduce**

[...]

PR #1041, opened 2026-04-01
Fix project root computation

toinehartman:
Rewrite the path config computation to consider several possible scenarios.

Compute the correctly root for
- Typical source files
- Source or TPL files in the target directory
- Nested projects (like example/test projects)

Triggered by, but not a fix for, #968.

### DIFF
--- a/rascal-lsp/src/test/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigsTest.java
+++ b/rascal-lsp/src/test/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigsTest.java
@@ -87,16 +87,6 @@ public class PathConfigsTest {
     }
 
     @Test
-    public void standardRoot() {
-        checkRoot(URIUtil.rootLocation("std"), "IO.rsc");
-    }
-
-    @Test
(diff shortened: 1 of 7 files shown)


rascal-lsp/src/test/java/org/rascalmpl/vscode/lsp/rascal/model/PathConfigsTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private final Projects projects = new Projects();
+
+    private void checkRoot(ISourceLocation project, String modulePath) throws URISyntaxException {
+        var m = VF.sourceLocation(project.getScheme(), project.getAuthority(), project.getPath() + "/" + modulePath);
+        var root = projects.inferRoot(m);
+        assertEquals(project, root);
+    }
+
+    @Mock PathConfigDiagnostics diagnostics;
```
DavyLandman: where is `PathConfigDiagnostics` defined?

DavyLandman (approved):
I think this looks okay.

toinehartman:
Superseded by https://github.com/usethesource/rascal-language-servers/pull/1053

PR #1053, opened 2026-04-13
Path config-less summaries for stdlib

### LINKED ISSUE #1048: Path config-less summary computation for library modules
**Describe the bug**

Currently, summary computation requires a path config. Path configs are only applicable to source projects, so for library projects, this is confusing and might introduce errors.

Change summary computation for library modules to not compute path configs.

Preparing work on Rascal: https://github.com/usethesource/rascal/pull/2748
Related issues:
- https://github.com/usethesource/rascal/issues/2743
- https://github.com/usethesource/rascal-language-servers/issues/968

**To Reproduce**

1. Open a module with stdlib imports.
2. Trigger a type-check.
3. Go-to-def on `import IO` (or similar)
4. In the opened standrad library module, go-to-def on some reference again.
5. Observe the path config in the debug console; it has a project root in `std://`, `jar+file://` or `mvn://`

**Expected behavior**
Do not compute a path config for binary modules.

 - Context: VS Code extension
 - Rascal Version: LSP `2.22.3`

toinehartman:
Closes #1048

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
@@ -196,7 +196,7 @@ public class RascalLanguageServices {
     }
 
     public InterruptibleFuture<@Nullable IConstructor> getSummary(ISourceLocation occ, Function<ISourceLocation, PathConfig> computePathConfig) {
-        Function<Evaluator, IConstructor> computeSummary;
+        Function<Evaluator, @Nullable IConstructor> computeSummary;
         var tplLoc = libraryTplLocation(occ);
         if (tplLoc != null) {
             computeSummary = eval -> {
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
@@ -137,7 +137,7 @@ public class RascalLanguageServices {
         return p.substring(1, p.lastIndexOf('.')).replaceAll("/", "::");
     }
 
-    private @Nullable Pair<ISourceLocation, String> libraryModule(ISourceLocation l) {
+    private @Nullable Pair<ISourceLocation, String> resolveLibraryModule(ISourceLocation l) {
         try {
             switch (l.getScheme()) {
                 case "mvn": {
@@ -161,7 +161,7 @@ public class RascalLanguageServices {
 
     public InterruptibleFuture<@Nullable IConstructor> getSummary(ISourceLocation occ, Function<ISourceLocation, PathConfig> computePathConfig) {
         Function<Evaluator, IConstructor> computeSummary;
-        var lib = libraryModule(occ);
+        var lib = resolveLibraryModule(occ);
         if (lib != null) {
             computeSummary = eval -> {
                 eval.doImport(eval, "util::PathConfig");
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
@@ -43,7 +43,6 @@ import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.function.Function;
 import java.util.stream.Collectors;
-import org.apache.commons.lang3.tuple.Pair;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.checkerframework.checker.nullness.qual.Nullable;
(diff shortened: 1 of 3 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -130,19 +132,60 @@ public RascalLanguageServices(RascalTextDocumentService docService, BaseWorkspac
         this.workspaceService = workspaceService;
     }
 
-    public InterruptibleFuture<@Nullable IConstructor> getSummary(ISourceLocation occ, PathConfig pcfg) {
+    private String pathToModuleName(ISourceLocation l) {
+        var p = l.getPath();
+        return p.substring(1, p.lastIndexOf('.')).replaceAll("/", "::");
```
DavyLandman: I agree, use `replace` that already does all, but doesn't parse the search string as a regex.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -130,19 +132,60 @@ public RascalLanguageServices(RascalTextDocumentService docService, BaseWorkspac
...
+                    var modTplDir = URIUtil.getChildLocation(URIUtil.changePath(modPath, "rascal"), modPrefix);
+                    return URIUtil.getChildLocation(modTplDir, tplFileName);
+                }
+                case "std": {
+                    var stdResolved = Locations.toPhysicalIfPossible(modPath);
+                    if ("jar+file".equals(stdResolved.getScheme())) {
+                        var stdJarRoot = URIUtil.changePath(stdResolved, stdResolved.getPath().substring(0, stdResolved.getPath().lastIndexOf('!') + 1));
+                        var modTplDir = URIUtil.getChildLocation(URIUtil.getChildLocation(stdJarRoot, "rascal"), modPrefix);
+                        return URIUtil.getChildLocation(modTplDir, tplFileName);
+                    }
```
DavyLandman: add comment for intended fall-through

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -130,19 +132,60 @@ public RascalLanguageServices(RascalTextDocumentService docService, BaseWorkspac
...
-            IString moduleName = VF.string(pcfg.getModuleName(occ));
-            return runEvaluator("Rascal makeSummary", semanticEvaluator, eval -> {
-                IConstructor result = (IConstructor) eval.call("makeSummary", moduleName, pcfg.asConstructor());
-                return result != null && result.asWithKeywordParameters().hasParameters() ? result : null;
-            }, null, exec, false, client);
-        } catch (IOException e) {
-            logger.error("Error looking up module name from source location {}", occ, e);
-            return InterruptibleFuture.completedFuture(null, exec);
+            var tplFileName = "$" + URIUtil.getLocationName(URIUtil.changeExtension(modPath, "tpl"));
+            var modPrefix = URIUtil.getParentLocation(modPath).getPath();
```
DavyLandman: these two can be merged, it avoids the double `getChildLocation` on both branches

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -130,19 +132,60 @@ public RascalLanguageServices(RascalTextDocumentService docService, BaseWorkspac
...
+            switch (modPath.getScheme()) {
+                case "jar+file": // intentional fall-through
+                case "mvn": {
+                    var modTplDir = URIUtil.getChildLocation(URIUtil.changePath(modPath, "rascal"), modPrefix);
+                    return URIUtil.getChildLocation(modTplDir, tplFileName);
+                }
+                case "std": {
+                    var stdResolved = Locations.toPhysicalIfPossible(modPath);
+                    if ("jar+file".equals(stdResolved.getScheme())) {
```
DavyLandman: what if std resolved to `mvn`?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -130,19 +132,60 @@ public RascalLanguageServices(RascalTextDocumentService docService, BaseWorkspac
...
-            }, null, exec, false, client);
-        } catch (IOException e) {
-            logger.error("Error looking up module name from source location {}", occ, e);
-            return InterruptibleFuture.completedFuture(null, exec);
+            var tplFileName = "$" + URIUtil.getLocationName(URIUtil.changeExtension(modPath, "tpl"));
+            var modPrefix = URIUtil.getParentLocation(modPath).getPath();
+            switch (modPath.getScheme()) {
+                case "jar+file": // intentional fall-through
+                case "mvn": {
+                    var modTplDir = URIUtil.getChildLocation(URIUtil.changePath(modPath, "rascal"), modPrefix);
```
DavyLandman: is this correct for `jar+file`? doesn't this remove the `!` part?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -130,19 +132,60 @@ public RascalLanguageServices(RascalTextDocumentService docService, BaseWorkspac
...
+                    return URIUtil.getChildLocation(modTplDir, tplFileName);
+                }
+                case "std": {
+                    var stdResolved = Locations.toPhysicalIfPossible(modPath);
+                    if ("jar+file".equals(stdResolved.getScheme())) {
+                        var stdJarRoot = URIUtil.changePath(stdResolved, stdResolved.getPath().substring(0, stdResolved.getPath().lastIndexOf('!') + 1));
+                        var modTplDir = URIUtil.getChildLocation(URIUtil.getChildLocation(stdJarRoot, "rascal"), modPrefix);
+                        return URIUtil.getChildLocation(modTplDir, tplFileName);
```
DavyLandman: I think this should be the case for `jar+file` and then after resolving it, we call ourselves (with a custom param for the module root).

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -130,19 +132,60 @@ public RascalLanguageServices(RascalTextDocumentService docService, BaseWorkspac
...
     }
 
+    public InterruptibleFuture<@Nullable IConstructor> getSummary(ISourceLocation occ, Function<ISourceLocation, PathConfig> computePathConfig) {
+        Function<Evaluator, IConstructor> computeSummary;
+        var tplLoc = libTplLoc(occ);
+        if (tplLoc != null) {
+            computeSummary = eval -> (IConstructor) eval.call("makeSummary", VF.string(pathToModuleName(occ)), tplLoc);
+        } else {
+            try {
+                var pcfg = computePathConfig.apply(occ);
```
DavyLandman: this calculation should be done inside the completable future (so moved inside of the `eval ->` part)

DavyLandman (changes requested):
Some small changes, especially for the case of the `jar+file`

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -130,17 +132,98 @@ public RascalLanguageServices(RascalTextDocumentService docService, BaseWorkspac
...
         try {
-            IString moduleName = VF.string(pcfg.getModuleName(occ));
-            return runEvaluator("Rascal makeSummary", semanticEvaluator, eval -> {
-                IConstructor result = (IConstructor) eval.call("makeSummary", moduleName, pcfg.asConstructor());
-                return result != null && result.asWithKeywordParameters().hasParameters() ? result : null;
-            }, null, exec, false, client);
-        } catch (IOException e) {
-            logger.error("Error looking up module name from source location {}", occ, e);
-            return InterruptibleFuture.completedFuture(null, exec);
+            var tplFolder = libraryTplRoot(modPath);
```
DavyLandman, line 149: yes, this is nice 👍🏼

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -130,17 +132,98 @@ public RascalLanguageServices(RascalTextDocumentService docService, BaseWorkspac
...
+        if ("jar+file".equals(modPath.getScheme())) {
+            // For a file within a JAR, return the sub-path within the JAR
+            return jarFilePath(modPath);
+        }
+        // Otherwise, return just the sub-path
+        return modPath.getPath();
+    }
+
+    private static ISourceLocation jarBasePath(ISourceLocation l) throws URISyntaxException {
+        if (!"jar+file".equals(l.getScheme())) {
```
DavyLandman: in principle we should see `l.getScheme().startsWith("jar+")` as it should behave the same for anything under it. for example `jar+home` or `jar+project` all we care about is that we're looking in a jar.

(this is true for all the cases that match `jar+file`)

DavyLandman (changes requested):
Looks fine, except we can be a bit more defensive and not just support `jar+file` but any  `jar+` scheme.

DavyLandman (approved):

PR #1051, opened 2026-04-09
Run UI test on Ubuntu 22.04 again

toinehartman:
Since 24.04 results in many (unpredictable) failures (see experiments [here](https://github.com/usethesource/rascal-language-servers/pull/1040)), we revert to 22.04 until we can practically use 24.04.

PR #1055, opened 2026-04-15
Fix use-before-initialize (NPE) in `TextDocumentState`

sungshik:
The UI tests on Ubuntu have become flaky (#1034 and #1037). This PR fixes a bug in the constructor of `TextDocumentState` to remove one cause of the flakiness.

### Problem

Before this PR, the constructor of `TextDocumentState`:
 1. Creates an initial `Update` object. The constructor of `Update`:
     1. Submits the parse call for the initial content to the worker pool.
     2. After parsing, stores the tree in an `AtomicReference` field of `TextDocumentState`.
 2. Initializes an `AtomicReference` object to store a tree after parsing.

Due to scheduling, it could happen that step 1.ii was executed before step 2. In that case, an NPE occurs, as an attempt is made to set the `AtomicReference` with the tree, but it hasn't been initialized yet. This NPE is swallowed inside a completion stage, though, so only the fallout (of not having stored the tree) is observable -- this is what happened in the UI tests.

### Solution

This PR:
  - Makes sure that the `AtomicReference` object is initialized before use. This addresses the root cause of the issue.
  - Makes sure that the post-parsing logic is executed in the worker pool (instead of the request pool). This doesn't address the root cause of the issue, but it is adjacent to it.
  - Improves exception reporting to make future debugging easier.

Related PRs (subsumed by this PR):
 - Fix that didn't work: #1051
 - Debugging: #1040

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java
@@ -30,6 +30,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiFunction;
@@ -62,6 +63,7 @@ public class TextDocumentState {
 
     private final BiFunction<ISourceLocation, String, CompletableFuture<ITree>> parser;
     private final ISourceLocation location;
+    private final ExecutorService exec;
 
     private final AtomicReference<Versioned<Update>> current;
     private final AtomicReference<@Nullable Versioned<ITree>> lastWithoutErrors;
@@ -70,12 +72,14 @@ public class TextDocumentState {
     public TextDocumentState(
             BiFunction<ISourceLocation, String, CompletableFuture<ITree>> parser,
             ISourceLocation location,
-            int initialVersion, String initialContent, long initialTimestamp) {
+            int initialVersion, String initialContent, long initialTimestamp,
+            ExecutorService exec) {
 
         this.parser = parser;
         this.location = location;
         this.lastWithoutErrors = new AtomicReference<>();
         this.last = new AtomicReference<>();
+        this.exec = exec;
 
         var u = new Update(initialVersion, initialContent, initialTimestamp);
         this.current = new AtomicReference<>(new Versioned<>(initialVersion, u));
@@ -200,7 +204,7 @@ public class TextDocumentState {
         private void parse() {
             try {
                 parser.apply(location, content)
-                    .whenComplete((ITree t, Throwable e) -> {
+                    .whenCompleteAsync((ITree t, Throwable e) -> {
                         if (e instanceof CompletionException && e.getCause() != null) {
                             e = e.getCause();
                         }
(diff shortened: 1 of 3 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java
```diff
@@ -200,28 +204,35 @@ public CompletableFuture<Versioned<List<Diagnostics.Template>>> getDiagnosticsAs
...
-                        }
 
-                        // Complete future to get diagnostics
-                        var diagnostics = new Versioned<>(version, diagnosticsList);
-                        diagnosticsAsync.complete(diagnostics);
-                    });
+                            // Complete future to get diagnostics
+                            var diagnostics = new Versioned<>(version, diagnosticsList);
+                            diagnosticsAsync.complete(diagnostics);
+                        } catch (Exception exc) {
```
toinehartman, line 229: Good catch! I think I'd prefer an exception handler for futures (like `.exceptionally`) at the end of the future chain instead of this try-catch where we don't expect an exception.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java
```diff
@@ -70,15 +72,17 @@ public class TextDocumentState {
...
             BiFunction<ISourceLocation, String, CompletableFuture<ITree>> parser,
             ISourceLocation location,
-            int initialVersion, String initialContent, long initialTimestamp) {
+            int initialVersion, String initialContent, long initialTimestamp,
+            ExecutorService exec) {
 
         this.parser = parser;
         this.location = location;
+        this.lastWithoutErrors = new AtomicReference<>();
+        this.last = new AtomicReference<>();
```
toinehartman, line 81: Good catch as well! This is the main fix.

toinehartman (approved):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java
```diff
@@ -200,28 +204,35 @@ public CompletableFuture<Versioned<List<Diagnostics.Template>>> getDiagnosticsAs
...
-                        diagnosticsAsync.complete(diagnostics);
-                    });
+                            // Complete future to get diagnostics
+                            var diagnostics = new Versioned<>(version, diagnosticsList);
+                            diagnosticsAsync.complete(diagnostics);
+                        } catch (Exception exc) {
+                            // The action of `whenCompleteAsync` shouldn't throw an exception (see JavaDoc): if it
+                            // unexpectedly does, then it is almost surely a bug, but the exception is swallowed, so it
+                            // is very hard to debug. The try/catch block and logger call aim to make it easier.
+                            logger.error("Unexpected exception after parsing: {}", exc);
```
toinehartman:
```suggestion
                            logger.error("Unexpected exception after parsing", exc);
```

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java
sungshik: I chose the try/catch approach here because the JavaDoc of `whenCompleteAsync` states: "the supplied action should not throw an exception. However, if it does, ..." Catching exceptions inside the action, instead of letting them leak out and handling them afterward, seemed closer to the spec.

PieterOlivier (approved):

DavyLandman (approved):

PR #1056, opened 2026-04-15
Fix flaky code lens test

toinehartman:
- Fix based on exploration of flaky tests here: #1040

### DIFF
--- a/rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
+++ b/rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
@@ -222,8 +222,15 @@ end
     it("code lens works", async function() {
         if (errorRecovery) { this.skip(); }
         const editor = await ide.openModule(TestWorkspace.picoFile);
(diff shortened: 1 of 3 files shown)


rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
```diff
@@ -222,8 +222,15 @@ end
...
-        await lens!.click();
+        await driver.wait(async () => {
+            try {
+                const lens = await editor.getCodeLens("Rename variables a to b.");
+                await lens!.click();
+                return true;
+            } catch (e) {
+                return false;
+            }
+        }, Delays.verySlow, "Rename lens should be available");
```
DavyLandman: Maybe we should move this towards the utilty function? since we have other places where we search for a code-lens I think?

DavyLandman (approved):
Looks good, just 1 question regarding introducing this as a lib feature for the tests?

rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
toinehartman: Done

DavyLandman (approved):

PR #1059, opened 2026-04-15
switched parameter order of rascal/edit event to fix #1049

jurgenvinju:
Fixes #1049 which has downstream (unreported) issues in uses of IDEServices::edit in vis::Graph and vis::Chart, salix en salix-contrib, and drAmbiguity.

### DIFF
--- a/rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
+++ b/rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
@@ -63,7 +63,7 @@ export async function activateLanguageClient(
     });
 
 
-    client.onNotification("rascal/editDocument", (uri: string, viewColumn: integer, range: vscode.Range) => {
+    client.onNotification("rascal/editDocument", (uri: string, range: vscode.Range, viewColumn: integer) => {
         openEditor(uri, range, viewColumn);
     });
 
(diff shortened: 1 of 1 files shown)


rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -63,7 +63,7 @@ export async function activateLanguageClient(
     });
 
 
-    client.onNotification("rascal/editDocument", (uri: string, viewColumn: integer, range: vscode.Range) => {
+    client.onNotification("rascal/editDocument", (uri: string, range: vscode.Range, viewColumn: integer) => {
```
rodinaarssen:
```suggestion
    client.onNotification("rascal/editDocument", (uri: string, range: vscode.Range | undefined, viewColumn: integer) => {
```

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
rodinaarssen:
```suggestion
async function openEditor(uriString: string, range: vscode.Range | undefined, viewColumn: integer) {
```

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -127,9 +127,11 @@ async function openEditor(uriString: string, range:vscode.Range, viewColumn: int
         // don't use the `selection` field here because we can not control scrolling behavior from that with editors which are already open
     });
 
-    // set the primary selection and move it into view (but don't scroll unless necessary)
-    editor.selection = new vscode.Selection(range.start, range.end);
-    editor.revealRange(range, vscode.TextEditorRevealType.InCenterIfOutsideViewport);
+    if (range !== null) {
```
rodinaarssen:
```suggestion
    if (range !== undefined) {
```

rodinaarssen (changes requested):
Good catch!

I've added some suggestions.

If this is an important feature many downstream projects depend on, it might be worth it to add a UI test for this.

DavyLandman (approved):

rodinaarssen (approved):

PR #1061, opened 2026-04-17
Restart extension without restarting Rascal server

sungshik:
This PR fixes an issue that affected developers of the Rascal VS Code extension. The issue manifested as follows:
  - Launch "Rascal Server"
  - Launch "Run Extension"
  - Execute command "Developer: Reload Window" (CTRL+R). After reloading the window, reconnecting with the Rascal server fails.

The primary cause of the issue is that the worker thread pool is shutdown when the extension disposes the connection. This PR fixes this by creating fresh pairs of thread pools for each next connection (both `requestPool` and `workerPool`). As before this PR, only one connection is usable at a time, so only one pair of thread pools is active at a time.

A secondary cause of the issue is that, after reloading the window and disposing the connection, there might still be uninterruptible tasks running (e.g., "Loading Rascal semantics", "Loading Rascal compiler"). These tasks will continue to (attempt to) send many progress updates to the extension, but the connection is no longer usable. It results in excessive exceptions inside the language client, to the extent that their impact on performance is noticeable (i.e., reconnecting the reloaded window to the Rascal server is slowed down substantially). This PR fixes this by introducing a proxy that, before forwarding calls to the actual language client, explicitly checks if the underlying thread pool (which runs the uninterruptible tasks) has already received the shutdown signal. Related issue: https://github.com/eclipse-lsp4j/lsp4j/issues/849.

This PR also makes the pairing of workspace service and document service more symmetric (mainly to avoid having a constructor with three args, to keep the pre-existing `BiFunction` style of `startLanguageServer`).

reload

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
@@ -45,7 +45,6 @@ import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.function.Function;
-
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
@@ -334,7 +334,7 @@ public abstract class BaseLanguageServer {
             var interfaces = new Class<?>[] { IBaseLanguageClient.class };
             var handler = (InvocationHandler) (Object proxy, Method method, Object[] args) -> {
                 if (this.executor.isShutdown()) {
-                    throw new IllegalStateException("Its thread pool was shut down, so the language client must no longer be used.");
+                    throw new IllegalStateException("The language client can no longer be used, as its thread pool is shutdown.");
                 }
                 return method.invoke(client, args);
             };
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
@@ -142,11 +142,16 @@ public abstract class BaseLanguageServer {
             var requestPool = NamedThreadPool.single(requestPoolName);
             var workerPool = NamedThreadPool.cached(workerPoolName);
 
-            var docService = docServiceProvider.apply(workerPool);
-            var wsService = workspaceServiceProvider.apply(workerPool);
-            docService.pair(wsService);
-            wsService.pair(docService);
-            startLSP(constructLSPClient(capturedIn, capturedOut, new ActualLanguageServer(() -> System.exit(0), workerPool, docService, wsService), requestPool));
+            try {
+                var docService = docServiceProvider.apply(workerPool);
+                var wsService = workspaceServiceProvider.apply(workerPool);
+                docService.pair(wsService);
+                wsService.pair(docService);
+                startLSP(constructLSPClient(capturedIn, capturedOut, new ActualLanguageServer(() -> System.exit(0), workerPool, docService, wsService), requestPool));
+            } finally {
+                requestPool.shutdown();
+                workerPool.shutdown();
+            }
         }
         else {
             try (ServerSocket serverSocket = new ServerSocket(portNumber, 0, InetAddress.getByName("127.0.0.1"))) {
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
@@ -31,6 +31,9 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.PrintStream;
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
 import java.net.InetAddress;
 import java.net.ServerSocket;
 import java.net.Socket;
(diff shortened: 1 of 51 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseWorkspaceService.java
```diff
@@ -67,17 +68,29 @@ public abstract class BaseWorkspaceService implements WorkspaceService, Language
     public static final String RASCAL_META_COMMAND = "rascal-meta-command";
     public static final String RASCAL_COMMAND = "rascal-command";
 
+    private final BooleanSupplier isConnected;
```
toinehartman: Is a use of this missing somewhere? If not, can we remove it?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/RascalLSPMonitor.java
```diff
@@ -115,6 +122,9 @@ private CompletableFuture<Void> createProgressBar(@UnderInitialization LSPProgre
         }
 
         private CompletableFuture<Void> tryRegisterProgress(@UnderInitialization LSPProgressBar this, String id) {
+            if (!isConnected.getAsBoolean()) {
```
toinehartman: Do we need to check for this every time we want to call functions on the language client?

toinehartman (commented):
Thanks for fixing this! I have some questions about keeping track of whether we're connected.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -131,24 +132,41 @@ private static void printClassPath() {
...
+                    var requestPool = NamedThreadPool.single(requestPoolName);
+                    var workerPool = NamedThreadPool.cached(workerPoolName);
+
+                    try (Socket clientSocket = serverSocket.accept()) {
+                        BooleanSupplier isConnected = () -> !clientSocket.isClosed();
+                        var docService = docServiceProvider.apply(isConnected, workerPool);
+                        var wsService = workspaceServiceProvider.apply(isConnected, workerPool);
+                        docService.pair(wsService);
+                        wsService.pair(docService);
+                        startLSP(constructLSPClient(clientSocket, new ActualLanguageServer(() -> {}, workerPool, docService, wsService), requestPool));
```
DavyLandman: Maybe the `isConnected` should be an `AtomicBoolean`? in that we we can also put it to false after `startLSP` returns, just to that all sides can stop doing things if one of them stops?

DavyLandman (commented):
Thanks for diving in why that fails 👍🏼 

I have a small remark, and the same question as @toinehartman , why is this `isConnected` property only check for the progress? Shouldn't it be on all the notifications we send (like for example diagnostics)?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/RascalLSPMonitor.java
sungshik: Fixed with a general proxy

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
sungshik: The current version directly checks the state of the underlying thread pool -- no more need for auxiliary booleans.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -131,24 +134,39 @@ private static void printClassPath() {
...
+                    var workerPool = NamedThreadPool.cached(workerPoolName);
+
+                    try (Socket clientSocket = serverSocket.accept()) {
+                        var docService = docServiceProvider.apply(workerPool);
+                        var wsService = workspaceServiceProvider.apply(workerPool);
+                        docService.pair(wsService);
+                        wsService.pair(docService);
+                        startLSP(constructLSPClient(clientSocket, new ActualLanguageServer(() -> {}, workerPool, docService, wsService), requestPool));
+                    }
+
```
DavyLandman: nit: remove this empty line before  the `finally`

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -131,24 +134,39 @@ private static void printClassPath() {
...
                 logger.info("Rascal LSP server listens on port number: {}", portNumber);
                 while (true) {
-                    var docService = docServiceProvider.apply(workerPool);
-                    var wsService = workspaceServiceProvider.apply(workerPool, docService);
-                    docService.pair(wsService);
-                    startLSP(constructLSPClient(serverSocket.accept(), new ActualLanguageServer(() -> {}, workerPool, docService, wsService), requestPool));
+                    var requestPool = NamedThreadPool.single(requestPoolName);
+                    var workerPool = NamedThreadPool.cached(workerPoolName);
+
+                    try (Socket clientSocket = serverSocket.accept()) {
```
DavyLandman, line 163: maybe we should print some kind of log message that we received a new connection to the server socket?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -131,24 +134,39 @@ private static void printClassPath() {
...
         if (DEPLOY_MODE) {
+            var requestPool = NamedThreadPool.single(requestPoolName);
+            var workerPool = NamedThreadPool.cached(workerPoolName);
+
             var docService = docServiceProvider.apply(workerPool);
-            var wsService = workspaceServiceProvider.apply(workerPool, docService);
+            var wsService = workspaceServiceProvider.apply(workerPool);
             docService.pair(wsService);
+            wsService.pair(docService);
             startLSP(constructLSPClient(capturedIn, capturedOut, new ActualLanguageServer(() -> System.exit(0), workerPool, docService, wsService), requestPool));
```
DavyLandman: I think we should also have a try finally there to shudown the worker pools after this has completed

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -304,13 +322,26 @@ public void setTrace(SetTraceParams params) {
...
-            lspWorkspaceService.connect(actualClient);
-            remoteIDEServicesConfiguration = RemoteIDEServicesThread.startRemoteIDEServicesServer(client, lspDocumentService, executor);
+            var proxy = proxyLanguageClient(client);
+            lspDocumentService.connect(proxy);
+            lspWorkspaceService.connect(proxy);
+            remoteIDEServicesConfiguration = RemoteIDEServicesThread.startRemoteIDEServicesServer(proxy, lspDocumentService, executor);
             logger.debug("Remote IDE Services Port {}", remoteIDEServicesConfiguration);
         }
 
+        private IBaseLanguageClient proxyLanguageClient(LanguageClient client) {
```
DavyLandman: Can we think of a better name than `proxyLanguageClient` that reflect what it does not how it does it?

something like `helpCleanShutdown` or something? or something about notifiactions?

I would also appreciate a link to the issue on lsp4j and a bit of documentation. Such that at a later point we remember why we have to do this, and possible come back to it later after lsp4j has improved this behavior.

DavyLandman (approved):

PR #1064, opened 2026-04-21
Lint & fix floating+misused promises

### LINKED ISSUE #1062: Check for unawaited promises with ESlint
**Describe the bug**

Our TS code base (or proposed changes) sometimes have unawaited promises.

Examples:
- https://github.com/usethesource/rascal-language-servers/pull/967#discussion_r3098968110
- https://github.com/usethesource/rascal-language-servers/blob/11e15f088c2b4691f10cef4077ca59a76d8f61c5/rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts#L250
- https://github.com/usethesource/rascal-language-servers/pull/1063

ESlint provides a rule to catch those: https://typescript-eslint.io/rules/no-floating-promises/
While at it, we should probably also check for misused promises: https://typescript-eslint.io/rules/no-misused-promises

toinehartman:
This PR adds ESLint checks for suspicous use of promises. In particular, it disallows implicit floating promises. This PR also fixes all errors produced by the linter. In most cases by adding `void`, but in some cases (for example in specific `async` functions, `await` seems like the more sensible choice.

Closes #1062.

### DIFF
--- a/rascal-vscode-extension/src/ux/RascalMFValidator.ts
+++ b/rascal-vscode-extension/src/ux/RascalMFValidator.ts
@@ -45,7 +45,7 @@ export class RascalMFValidator implements vscode.Disposable {
         vscode.workspace.onDidChangeWorkspaceFolders(async ws => {
             for (const added of ws.added) {
                 if (await isRascalProject(added.uri)) {
-                    this.verifyRascalMF(added.uri);
+                    await this.verifyRascalMF(added.uri);
                 }
             }
             for (const rem of ws.removed) {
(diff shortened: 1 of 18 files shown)


rascal-vscode-extension/src/ux/RascalMFValidator.ts
```diff
@@ -45,7 +45,7 @@ export class RascalMFValidator implements vscode.Disposable {
         vscode.workspace.onDidChangeWorkspaceFolders(async ws => {
             for (const added of ws.added) {
                 if (await isRascalProject(added.uri)) {
-                    this.verifyRascalMF(added.uri);
+                    await this.verifyRascalMF(added.uri);
```
DavyLandman: In this case we don't want to await, as we want to fire them off in the background. so more like the project validator.

DavyLandman (approved):

Issue #1065, opened 2026-04-22
Diagnostics of untitled files aren't always cleared (race)


sungshik:
**Describe the bug**

When "Revert and Close Editor" is used on a non-empty untitled file (e.g., in the UI tests), two `publishDiagnostics` calls race each other:
  - Call A is triggered by the `didChange` event (i.e., the "Revert" part of the command);
  - Call B is triggered by the `didClose` event (i.e., the "Close" part of the command).

Both calls are scheduled on the worker pool. If call A precedes call B, then:
  1. First, diagnostics are set to those of the empty file. In the case of Pico (e.g., in the UI tests), this is a parse error.
  2. Next, diagnostics are cleared.

This is ok. However, if call B precedes call A (which occasionally happens), then the diagnostics are first cleared and next set, so they reappear and are never cleared again. This isn't ok.

**To Reproduce (Manual)**

Steps:
 1. Open a Rascal terminal, import the language server of Pico (`import demo::lang::pico::LanguageServer;`), and register it (`main()`).
 2. Open an untitled file.
 3. Run command "Change Language Mode" and select "Parametric Rascal LSP". An error diagnostic is reported.
 4. Edit the file.
 5. Run command "Revert and Close Editor". Sometimes the error diagnostic is removed; sometimes it isn't.

Steps 2-5 might need to be repeated a few times because of scheduling.

**To Reproduce (UI Tests)**

To reproduce in UI tests, the following could be added to `afterEach` in `dsl.test.ts` to check if all error diagnostics have been cleared:

https://github.com/usethesource/rascal-language-servers/blob/6033ab06bc4027e4ac8c00e43c7ef7f93204cb67/rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts#L97-L103

```ts
// Add after line 102:
const bottomBar = bench.getBottomBar();
const problemsView = await bottomBar.openProblemsView();
const allVisibleMarkers = await problemsView.getAllVisibleMarkers(MarkerType.Error);
expect(allVisibleMarkers.length, "Not all error diagnostics have been cleared").to.equal(0);
```

With this addition, the UI tests most of the time fail:
```
  1) DSL+recovery
       "after each" hook for "has syntax highlighting in documents without extension":

      Not all error diagnostics have been cleared
      + expected - actual

      -1
      +0
```

[...]

PR #1066, opened 2026-04-22
Diagnostics of extension-less files are cleared

### LINKED ISSUE #1058: Diagnostics of extension-less files aren't cleared
**Describe the bug**

Diagnostics of extension-less files aren't cleared.

**To Reproduce (Manual)**

Steps:

 1. Open a Rascal terminal, import the language server of Pico (`import demo::lang::pico::LanguageServer;`), and register it (`main()`).
 2. Create and open an extension-less file.
 3. Run command "Change Language Mode" and select "Parametric Rascal LSP". An error diagnostic is reported.
 4. Close the file. The error diagnostic is still in the Problems view. **(A)**
 5. Delete the file. The error diagnostic is still in the Problems view. **(B)**

Issues:

  - **A:** I wonder if the error diagnostic should have been cleared after step 3.
      - It shouldn't have: The file still exists.
      - It should have: The file isn't in language mode "Parametric Rascal LSP" anymore (as witnessed by the fact that, upon reopening, the language mode has reverted to "Plain Text").
  - **B:** The error diagnostic should have been cleared after step 4. 

**To Reproduce (UI Tests)**

To reproduce in UI tests, the following could be added to `afterEach` in `dsl.test.ts` to check if all error diagnostics have been cleared:

https://github.com/usethesource/rascal-language-servers/blob/6033ab06bc4027e4ac8c00e43c7ef7f93204cb67/rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts#L97-L103

```ts
// Add after line 102:
const bottomBar = bench.getBottomBar();
const problemsView = await bottomBar.openProblemsView();
const allVisibleMarkers = await problemsView.getAllVisibleMarkers(MarkerType.Error);
expect(allVisibleMarkers.length, "Not all error diagnostics have been cleared").to.equal(0);
```

With this addition, the UI tests fail:
```
  1) DSL+recovery
       "after each" hook for "has syntax highlighting in documents without extension":

      Not all error diagnostics have been cleared
      + expected - actual

      -1
      +0
```

(In general, the addition isn't perfect yet, as it checks only the "visible" diagnostics, so if scrolling down past a large number of warnings is needed to find the first error, then that error is missed and the check mistakenly succeeds. For all current tests, though, I think this works fine.)

sungshik:
This PR fixes #1058. It doesn't add UI tests yet, because diagnostics cannot be reliably counted until #1065 is fixed.

With help from @toinehartman.

clear-diagnostics-extension-less-file

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseWorkspaceService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseWorkspaceService.java
@@ -33,8 +33,6 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutorService;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
@@ -47,12 +45,6 @@ import org.eclipse.lsp4j.DidChangeWatchedFilesParams;
 import org.eclipse.lsp4j.DidChangeWorkspaceFoldersParams;
 import org.eclipse.lsp4j.ExecuteCommandParams;
 import org.eclipse.lsp4j.FileDelete;
-import org.eclipse.lsp4j.FileOperationFilter;
-import org.eclipse.lsp4j.FileOperationOptions;
-import org.eclipse.lsp4j.FileOperationPattern;
-import org.eclipse.lsp4j.FileOperationPatternKind;
-import org.eclipse.lsp4j.FileOperationsServerCapabilities;
-import org.eclipse.lsp4j.FileOperationsWorkspaceCapabilities;
 import org.eclipse.lsp4j.RenameFilesParams;
 import org.eclipse.lsp4j.ServerCapabilities;
 import org.eclipse.lsp4j.WorkspaceClientCapabilities;
@@ -98,16 +90,6 @@ public abstract class BaseWorkspaceService implements WorkspaceService, Language
             folderOptions.setSupported(true);
             folderOptions.setChangeNotifications(true);
             workspaceCapabilities.setWorkspaceFolders(folderOptions);
-        }
-
-        // Make sure that all delete events are listened to by this workspace service (and propagated to the linked
-        // document service by method `didDeleteFiles`), so all diagnostics of deleted files can be properly cleared.
-        var fileOperationCapabilities = Nullables.ensureNonNullAndGet(workspaceCapabilities, WorkspaceServerCapabilities::getFileOperations, WorkspaceServerCapabilities::setFileOperations, FileOperationsServerCapabilities::new);
-        if (Nullables.has(clientCap.getWorkspace(), WorkspaceClientCapabilities::getFileOperations, FileOperationsWorkspaceCapabilities::getDidDelete)) {
-            var anyFile = new FileOperationPattern("**/*");
-            anyFile.setMatches(FileOperationPatternKind.File);
-            var didDelete = new FileOperationOptions(Stream.of(anyFile).map(FileOperationFilter::new).collect(Collectors.toList()));
-            fileOperationCapabilities.setDidDelete(didDelete);
         }
     }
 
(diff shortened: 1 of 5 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseWorkspaceService.java
```diff
@@ -91,6 +99,16 @@ public void initialize(ClientCapabilities clientCap, @Nullable List<WorkspaceFol
...
         }
+
+        // Make sure that all delete events are listened to by this workspace service (and propagated to the linked
+        // document service by method `didDeleteFiles`), so all diagnostics of deleted files can be properly cleared.
+        var fileOperationCapabilities = Nullables.ensureNonNullAndGet(workspaceCapabilities, WorkspaceServerCapabilities::getFileOperations, WorkspaceServerCapabilities::setFileOperations, FileOperationsServerCapabilities::new);
+        if (Nullables.has(clientCap.getWorkspace(), WorkspaceClientCapabilities::getFileOperations, FileOperationsWorkspaceCapabilities::getDidDelete)) {
+            var anyFile = new FileOperationPattern("**/*");
+            anyFile.setMatches(FileOperationPatternKind.File);
+            var didDelete = new FileOperationOptions(Stream.of(anyFile).map(FileOperationFilter::new).collect(Collectors.toList()));
+            fileOperationCapabilities.setDidDelete(didDelete);
```
toinehartman: I don't think is correct, since for parametric, we set this capability dynamically here: https://github.com/usethesource/rascal-language-servers/blob/45c839a5406c2265becc1fe910e3e0e91942db88/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java#L265-L268

Setting the same capability statically and dynamically is not allowed ([LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#client_registerCapability)).

toinehartman (changes requested):
I believe we need to do this registration dynamically instead of statically.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/FileOperationCapability.java
```diff
@@ -84,6 +84,11 @@ protected final FileOperationOptions mergeOptions(FileOperationOptions o1, FileO
         anyFolder.setMatches(FileOperationPatternKind.Folder);
         patterns.add(new FileOperationFilter(anyFolder));
 
+        // Only extension-less files would be enough, but it seems that cannot be expressed using LSP's glob patterns.
```
sungshik: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#workspace_willCreateFiles

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseWorkspaceService.java
sungshik: Thanks! Updated the PR as discussed in the call, except it seems LSP's glob patterns cannot express extension-less files.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/FileOperationCapability.java
toinehartman: Can we not do `**/[!.-.]`? ([source](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#fileOperationPattern))

sungshik: Nope, `[!.-.]` matches only a single character, so this pattern matches `foo/b`, but not `foo/ba`.

toinehartman (approved):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/FileOperationCapability.java
```diff
@@ -84,6 +84,11 @@ protected final FileOperationOptions mergeOptions(FileOperationOptions o1, FileO
...
+        var anyFile = new FileOperationPattern("**/*");
+        anyFile.setMatches(FileOperationPatternKind.File);
+        patterns.add(new FileOperationFilter(anyFile));
```
DavyLandman: Can we limit the file operation to only fire for deletes? and not all events?

PR #1069, opened 2026-04-28
[POM-leading for DSLs] Routing to dedicated remote language servers

### LINKED ISSUE #1076: Parametric routing layer with dedicated servers
**Goals**

- Basic DSL features should remain working.
- Each DSL should run in its own process with its own dependency versions on the class path.

**Not in scope**

- Dynamic registration of capabilities.
- Reading editor contents of other DSLs.
- IDE services.

**Steps**

1. Replace parametric server with a thin routing server that forwards LSP requests to remote servers (1 per language).
2. Launch remote server in dedicated process with class path based on language project configuration (i.e. dependency versions).
3. Forward LSP server requests/notifications to server(s) that they apply to.
4. Forward client requests from remote servers to routing layer back to VS Code client.

toinehartman:
This PR is the first of a series. gradually implementing POM-leading behavior in DSLs/the parametric server.
This PR introduces a routing language server, which delegates incoming LSP requests to the appropriate parametric language server, dedicated to a single language. A dedicated server is started upon initial language registration.

Closes #1076


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/conversion/SemanticTokenizer.java
```diff
@@ -76,7 +76,7 @@ public SemanticTokens semanticTokensFull(ITree tree, boolean specialCaseHighligh
         return new SemanticTokens(tokens.getTheList());
     }
 
-    public SemanticTokensWithRegistrationOptions options() {
+    public static SemanticTokensWithRegistrationOptions options() {
```
toinehartman, line 79: This change allows us to make `ParametricTextDocumentService::setStaticServerCapabilities` `static` as well and re-use it in the new `RoutingTextDocumentService`

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
```diff
@@ -43,13 +43,12 @@
...
 import org.eclipse.lsp4j.WorkspaceServerCapabilities;
 import org.rascalmpl.vscode.lsp.BaseWorkspaceService;
-import org.rascalmpl.vscode.lsp.IBaseTextDocumentService;
 import org.rascalmpl.vscode.lsp.util.Nullables;
 
 public class RascalWorkspaceService extends BaseWorkspaceService {
 
-    RascalWorkspaceService(ExecutorService exec, IBaseTextDocumentService documentService) {
-        super(exec, documentService);
+    RascalWorkspaceService(ExecutorService exec) {
```
toinehartman, line 52: This argument is replaced by a `pair` function

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,592 @@
...
+                    , "-Dlog4j2.configurationFactory=org.rascalmpl.vscode.lsp.log.LogJsonConfiguration"
+                    , "-Dlog4j2.level=DEBUG"
+                    , "-Drascal.fallbackResolver=org.rascalmpl.vscode.lsp.uri.FallbackResolver"
+                    , "-Drascal.lsp.deploy=true"
+                    , "-Drascal.compilerClasspath=" + classPath
+                    , "-Xmx2048M"
+                    , "-cp", classPath
+                    , "org.rascalmpl.vscode.lsp.parametric.ParametricLanguageServer"
+                    , "--exitWhenEmpty"
+                    // , new GsonBuilder().create().toJson(lang, LanguageParameter.class).replace("\"", "\\\"") // escape JSON string on command line
```
toinehartman: It would be nice you use the dedicated language feature of the parametric LSP here, but then we should also probably swallow the `registerLanguage` call that triggered this, as otherwise, we are registering twice and wasting resources.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,592 @@
...
+ * A language server implementation that routes LSP requests to dedicated remote language servers.
+ */
+public class ActualRoutingLanguageServer extends BaseLanguageServer.ActualLanguageServer implements IBaseLanguageClient, DocumentRouter<CompletableFuture<IBaseLanguageServerExtensions>> {
+
+    private static final Logger logger = LogManager.getLogger(ActualRoutingLanguageServer.class);
+
+    private final Map<String, String> languagesByExtension = new ConcurrentHashMap<>();
+    // NOTE
+    // 1. This map should only contains running server processes.
+    // 2. Upon removal from this map, the process should be killed to avoid resource leaks.
```
DavyLandman, line 110: we should also register a shutdown hook to always close our child processes.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,592 @@
...
+    }
+
+    @Override
+    public void connect(LanguageClient client) {
+        this.client = (IBaseLanguageClient) client;
+        super.connect(client);
+    }
+
+    public CompletableFuture<IBaseLanguageServerExtensions> route(String extension) {
+        return safeLanguage(extension).map(this::languageByName).orElseThrow(() -> {
```
DavyLandman: this exception is thrown in the context of `route` not in the context of the completable future, so that might be not so nice.

Same is true for the `safeLanguage` exception.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,592 @@
...
+        var c = art.getCoordinate();
+        if (!c.getGroupId().equals("org.rascalmpl")) {
+            return false;
+        }
+        var id = c.getArtifactId();
+        return "rascal-lsp".equals(id);
+    }
+
+    private static String classPath(LanguageParameter lang) {
+        var separator = System.getProperty("path.separator");
```
DavyLandman: let's use `File.pathSeparator` instead. (this maybe negates the need for this).

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,592 @@
...
+        PathConfig pcfg;
+        try {
+            pcfg = PathConfig.parse(lang.getPathConfig());
+        } catch (IOException e) {
+            logger.error("Error while parsing path config {}", lang.getPathConfig(), e);
+            logger.warn("Could not compute class path for {}; falling back to class path of routing server", lang.getName());
+            return System.getProperty("java.class.path");
```
DavyLandman: I'm not sure if we should have a fallback like this. It can lead to all kinds of down stream errors that a user didn't suspect, all because of like some parse error in the pom.xml that wasn't reported to the user as a breaking error.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,592 @@
...
+
+        var pom = Locations.toPhysicalIfPossible(URIUtil.getChildLocation(pcfg.getProjectRoot(), "pom.xml"));
+        var p = new MavenParser(Path.of(pom.getURI()));
+        try {
+            var rootProject = p.parseProject();
+
+            // Check if we are in Rascal-LSP
+            var classPath = new StringBuilder();
+            if (isRascalLspProject(rootProject)) {
+                classPath.append(separator);
```
DavyLandman: Maybe we should collect all entries in a list, and then at the end join them with the path sep as element seperator?

so that he final substring can also be skipped.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,592 @@
...
+                classPath.append(Path.of(Locations.toUri(pcfg.getBin())));
+            }
+
+            // Check the project dependencies
```
DavyLandman: the path config construction code already contains logic like this, it's used for the construction of the REPL for example. I rather we reuse that same logic, instead of duplicating it here.

As it's more complex than this, for example in the presence of local projects in the workspace for example.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,592 @@
...
+            }
+            // strip of the initial separator
+            return classPath.substring(separator.length());
+        } catch (ModelResolutionError e) {
+            logger.error("Error while parsing POM at {}", pom, e);
+        }
+
+        // If all else fails, just use the same class path that we have
+        logger.warn("Could not compute class path for {}; falling back to class path of routing server", lang.getName());
+        return System.getProperty("java.class.path");
```
DavyLandman: this should be an exception , not a fallback that get's easy to miss.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,592 @@
...
+        logger.debug("{} runs with class path {}", lang.getName(), classPath);
+        // In deployment, we start a process and connect to it via input/output streams
+        try {
+            var proc = new ProcessBuilder(ProcessHandle.current().info().command().orElse("java")
+                    , "-Dlog4j2.configurationFactory=org.rascalmpl.vscode.lsp.log.LogJsonConfiguration"
+                    , "-Dlog4j2.level=DEBUG"
+                    , "-Drascal.fallbackResolver=org.rascalmpl.vscode.lsp.uri.FallbackResolver"
+                    , "-Drascal.lsp.deploy=true"
+                    , "-Drascal.compilerClasspath=" + classPath
+                    , "-Xmx2048M"
```
DavyLandman, line 272: this should become a vs code setting I think.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,592 @@
...
+                    , "-cp", classPath
+                    , "org.rascalmpl.vscode.lsp.parametric.ParametricLanguageServer"
+                    , "--exitWhenEmpty"
+                    // , new GsonBuilder().create().toJson(lang, LanguageParameter.class).replace("\"", "\\\"") // escape JSON string on command line
+                )
+                .redirectError(Redirect.INHERIT) // Show logs in current process
```
DavyLandman: smart. I was wondering about this. but won't the thread names get mixed now? so rascal-lsp-1 from lsp server 1 and rascal-lsp-1 from server 2 are not able to keep apart.

we might have to proxy the json log output as well, and add a prefix or something of the language.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,592 @@
...
+
+    @Override
+    public void editDocument(URI uri, @Nullable Range range, int viewColumn) {
+        availableClient().editDocument(uri, range, viewColumn);
+    }
+
+    @Override
+    public void startDebuggingSession(int serverPort) {
+        availableClient().startDebuggingSession(serverPort);
+    }
```
DavyLandman: I don't get all the `availableClient` calls, should this class maybe not implement these functions? but instead leave that to the actual client implemetation?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ProxiedIValue.java
```diff
@@ -0,0 +1,189 @@
...
+                throw new IOException("Malformed JSON");
+            default:
+                throw new IOException("Unknown JSON token");
+        }
+    }
+
+    private static void readString(JsonReader reader, StringBuilder sb) throws IOException {
+        sb.append('"');
+        sb.append(reader.nextString());
+        sb.append('"');
```
DavyLandman: We should not be manually constructing a json string. this is too brittle.

For example, what if the string contained a nested quote?

We should GSON itself to write a json stream to a string writer or something, but not like this.

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -168,7 +168,7 @@ async function buildRascalServerOptions(jarPath: string, isParametricServer: boo
     ];
     let mainClass: string;
     if (isParametricServer) {
-        mainClass = 'org.rascalmpl.vscode.lsp.parametric.ParametricLanguageServer';
+        mainClass = 'org.rascalmpl.vscode.lsp.parametric.routing.RoutingLanguageServer';
         commandArgs.push(calculateDSLMemoryReservation(dedicated));
```
DavyLandman, line 172: I think we can dedicate less memory to the routing language server now?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
```diff
@@ -0,0 +1,424 @@
...
+public class RoutingTextDocumentService implements IBaseTextDocumentService, DocumentRouter<CompletableFuture<IBaseTextDocumentService>> {
+
+    private static final Logger logger = LogManager.getLogger(RoutingTextDocumentService.class);
+
+    private @MonotonicNonNull LanguageClient client;
+    private @MonotonicNonNull BaseWorkspaceService wsService;
+    private @MonotonicNonNull ActualRoutingLanguageServer parentServer;
+
+    @SuppressWarnings("unused")
+    private final ExecutorService exec;
```
DavyLandman: why keep it around if we don't use it?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingWorkspaceService.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public RoutingWorkspaceService(ExecutorService exec) {
+        super(exec);
+    }
+
+    public void setParentServer(ActualRoutingLanguageServer server) {
+        this.parentServer = server;
+    }
+
+    private ActualRoutingLanguageServer availableServer() {
```
DavyLandman: I don't understand this extra edge, why do we need to loop back to the server? I've had this same question when reading this PR, I don't like all these cycles, so can we do without?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingWorkspaceService.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Override
+    public CompletableFuture<Object> executeCommand(ExecuteCommandParams commandParams) {
+        if (commandParams.getCommand().startsWith(RASCAL_META_COMMAND) || commandParams.getCommand().startsWith(RASCAL_COMMAND)) {
+            var languageName = ((JsonPrimitive) commandParams.getArguments().get(0)).getAsString();
+            return availableServer().languageByName(languageName)
+                .thenCompose(s -> s.getWorkspaceService().executeCommand(commandParams));
+        }
+
+        return CompletableFutureUtils.completedFuture(commandParams.getCommand() + " was ignored.", getExecutor());
```
DavyLandman: can we say why it was ignored?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IBaseLanguageServerExtensions.java
```diff
@@ -63,4 +64,7 @@ default CompletableFuture<Two<String, URI[]>[]> supplyPathConfig(PathConfigParam
 
     @JsonNotification("rascal/logLevel")
     void setMinimumLogLevel(String level);
+
+    @JsonDelegate
+    IBaseTextDocumentService getIBaseTextDocumentService();
```
DavyLandman: Why this extension? can we add a comment? I don't see how the `IBaseTextDocumentService` could ever be serialized over json?

rascal-lsp/src/main/rascal/library/demo/lang/pico/LanguageServer.rsc
```diff
@@ -361,9 +361,16 @@ in the presence of error trees. See ((util::LanguageServer)) for more details.
 Any feedback (errors and exceptions) is faster and more clearly printed in the terminal.
 }
 void main(bool errorRecovery=false) {
+    loc root;
+    try {
+        root = resolveLocation(|project://rascal-lsp|);
+    } catch SchemeNotSupported(_): {
+        root = resolveLocation(|cwd:///../../../rascal-lsp|);
```
DavyLandman, line 370: let's add a comment to explain why we're doing this.

DavyLandman (changes requested):
I think this is heading in the right direction. But it's not ready yet.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ProxiedIValue.java
toinehartman: Replaced the string by `JsonElement`

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingWorkspaceService.java
toinehartman: Refactored routing and using the `DocumentRouter` interface here instead to get rid of some dependencies.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IBaseLanguageServerExtensions.java
toinehartman: Left over from implementing command execution, but now cleaned up since we refactored the routing.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
toinehartman: Now using the path config passed during registration. Unfortunately we still need to find a way from `mvn` URIs to something that Java understands. Now mapping them manually.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
toinehartman: Not a good idea since the assumptions of dedicated languages are not clear. E.g. what happens when we register more contributions later?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,515 @@
...
+    private @MonotonicNonNull InitializeParams initializeParams;
+    private final JsonWriter logForwarder;
+
+    private static final int REMOTE_BASE_PORT = 9990;
+    private AtomicInteger remotePortOffset = new AtomicInteger(0);
+
+    @SuppressWarnings("java:S106") // System.out
+    public ActualRoutingLanguageServer(Runnable onExit, ExecutorService exec, IBaseTextDocumentService lspDocumentService, BaseWorkspaceService lspWorkspaceService) {
+        super(onExit, exec, lspDocumentService, lspWorkspaceService);
+        logForwarder = new JsonWriter(new BufferedWriter(new OutputStreamWriter(System.out)));
```
DavyLandman: We should also close (or at least flush) this log forwarder, or the underlying `BufferedWriter`.

Also, let's add a comment why we're taking a direct writer around system.out, and the relation ship to the regular log4j2 writers that are active.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,515 @@
...
+
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> destroyChildProcesses()));
+    }
+
+    private static void destroyChildProcesses() {
+        ProcessHandle.current().children().forEach(p -> {
+            if (p.isAlive() && !p.destroy()) {
+                p.destroyForcibly();
+            }
+        });
```
DavyLandman: I think we should try and catch an exception here, to make sure the for-each doesn't terminate to soon.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,515 @@
...
+    @Override
+    public CompletableFuture<IBaseLanguageServerExtensions> route(String lang) {
+        return CompletableFutureUtils.completedFuture(languageServers.get(lang), getExecutor())
+            .thenApply(service -> {
+                if (service == null) {
+                    throw new UnsupportedOperationException(String.format("Rascal Parametric LSP has no support for this file, since no language is registered with name '%s'", lang));
+                }
+                return service;
+            })
+            .thenCompose(Function.identity());
```
DavyLandman: how about:

```java
var service = language.get(lang);
if (service == null) {
   return CompletableFuture.exceptionally(...);
}
return service;
```

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,515 @@
...
+    private static List<Path> classPath(LanguageParameter lang) throws IOException, ModelResolutionError {
+        var pcfg = PathConfig.parse(lang.getPathConfig());
+        var pom = Locations.toPhysicalIfPossible(URIUtil.getChildLocation(pcfg.getProjectRoot(), "pom.xml"));
+        var maven = new MavenParser(Path.of(pom.getURI()));
+
+        var project = maven.parseProject();
+        var deps = project.resolveDependencies(Scope.COMPILE, maven);
+
+        if (isRascalLsp(project)) {
+            // We add our bin directory and all of our dependencies
```
DavyLandman: let's exaplain the why, like why is rascalLsp special?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,515 @@
...
+                }).collect(Collectors.toList());
+            return Lists.union(List.of(target), depPaths);
+        }
+
+        return deps.stream()
+            .filter(d -> isRascal(d) || isRascalLsp(d))
+            .flatMap(a -> {
+                var r = a.getResolved();
+                return r == null ? Stream.of() : Stream.of(r);
+            })
```
DavyLandman: nit: it can also be a:

```java
.map(ArtifactCoordinate::resolved)
.filter(r => r != null)
```

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,515 @@
...
+            }
+        } catch (Exception e) { /* ignored */ }
+    }
+
+    @SuppressWarnings("java:S106") // System.out
+    private void forwardLogs(InputStream logStream, String langName) {
+        getExecutor().execute(() -> {
+            var line = "";
+            var threadPrefix = langName + " | ";
+            var reader = new BufferedReader(new InputStreamReader(logStream));
```
DavyLandman: this read should be closed right? so maybe using the `try-with-resources` block?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,515 @@
...
+            if (threadName != null) {
+                obj.addProperty("threadName", prefix + threadName);
+            }
+        } catch (Exception e) { /* ignored */ }
+    }
+
+    @SuppressWarnings("java:S106") // System.out
+    private void forwardLogs(InputStream logStream, String langName) {
+        getExecutor().execute(() -> {
+            var line = "";
```
DavyLandman: we can also say `String line;` and move it to just before the while loop starts.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,515 @@
...
+                            gson.toJson(json, logForwarder);
+                            logForwarder.flush();
+                            // One object per line; this is what log4j does as well.
+                            System.out.println();
+                        }
+                    } catch (JsonSyntaxException e) {
+                        // Sometimes the child process logs non-JSON (e.g. logs while setting up the JSON logger).
+                        // In this case, just forward the raw line.
+                        if (!line.isBlank()) {
+                            System.out.println(line);
```
DavyLandman: let's add a small comment that explains that `println` already locks on `System.out` by itself.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,515 @@
...
+            var classPath = String.join(File.pathSeparator, classPath(lang).stream().map(Path::toString).collect(Collectors.toList()));
+            logger.debug("{} runs with class path {}", lang.getName(), classPath);
+
+            var proc = new ProcessBuilder(ProcessHandle.current().info().command().orElse("java")
+                    , "-Dlog4j2.configurationFactory=org.rascalmpl.vscode.lsp.log.LogJsonConfiguration"
+                    , "-Dlog4j2.level=DEBUG"
+                    , "-Drascal.fallbackResolver=org.rascalmpl.vscode.lsp.uri.FallbackResolver"
+                    , "-Drascal.lsp.deploy=true"
+                    , "-Drascal.compilerClasspath=" + classPath
+                    , "-Xmx2048M"
```
DavyLandman, line 272: we should make this memory a setting/calculation. as 2GB can be quite big if you start a few of them. right now we calculate them in VS Code, but we now maybe have to calculate it here, and have an setting that can override it.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,515 @@
...
+        logger.info("Starting LSP process for {}", lang.getName());
+
+        // In deployment, we start a process and connect to it via input/output streams
+        try {
+            var classPath = String.join(File.pathSeparator, classPath(lang).stream().map(Path::toString).collect(Collectors.toList()));
+            logger.debug("{} runs with class path {}", lang.getName(), classPath);
+
+            var proc = new ProcessBuilder(ProcessHandle.current().info().command().orElse("java")
+                    , "-Dlog4j2.configurationFactory=org.rascalmpl.vscode.lsp.log.LogJsonConfiguration"
+                    , "-Dlog4j2.level=DEBUG"
```
DavyLandman: we should do the same log level as we're started with.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,515 @@
...
+            try {
+                while ((line = reader.readLine()) != null) {
+                    try {
+                        var json = JsonParser.parseString(line);
+                        prependThreadName(threadPrefix, json);
+                        synchronized (System.out) { // Lock, so we can make sure our JSON is followed by a newline.
+                            gson.toJson(json, logForwarder);
+                            logForwarder.flush();
+                            // One object per line; this is what log4j does as well.
+                            System.out.println();
```
DavyLandman: When I read this, I think we should instead write to our `System.err` as the `System.out` doesn't flush on a newline. 

In a way, the fact we retarged system.out is nice, but logging should be happening to `system.err` so instead we should use that directly here, instead of depending on the redirection to `system.out`.

Since our own log4j2 also writes to `system.err` we also further reduce the chance of deadlocks.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,515 @@
...
+                            languagesByExtension.remove(ext, lang.getName());
+                        }
+                    }
+                }
+            }
+            try {
+                serverParams.getRight().run();
+            } catch (Exception e) {
+                logger.error("Unexpected error while cleaning up connection to language server for {}", lang.getName(), e);
+            }
```
DavyLandman: should this not be part of the finallized block?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,515 @@
...
+        var serverLauncher = new Launcher.Builder<IBaseLanguageServerExtensions>()
+            .setRemoteInterface(IBaseLanguageServerExtensions.class)
+            .setLocalService(client)
+            .setInput(serverParams.getLeft())
+            .setOutput(serverParams.getMiddle())
+            .configureGson(ActualRoutingLanguageServer::configureProxyGson)
+            .setExecutorService(getExecutor())
+            .create();
+
+        var runner = serverLauncher.startListening();
```
DavyLandman, line 367: there is also a clean way to stop this server, it has a cancel function. maybe we can use that somewhere for a cleaner shutdown?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingLanguageServer.java
```diff
@@ -0,0 +1,53 @@
...
+ * A language-parametric server that assigns a dedicated server to each language.
+ * This server routes LSP requests to the appropriate language server.
+ */
+public class RoutingLanguageServer extends ParametricLanguageServer {
+
+    public static void main(String[] args) {
+        var serverArgs = parseArgs(args);
+        if (serverArgs.getDedicatedLanguage() != null) {
+            // If we get a dedicated language argument, we just start a single parametric server
+            startParametric(serverArgs);
```
DavyLandman, line 41: good 👍🏼

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
DavyLandman: this should be closed, at least the `BufferedWriter`

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,515 @@
...
+        logForwarder = new JsonWriter(new BufferedWriter(new OutputStreamWriter(System.out)));
+
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> destroyChildProcesses()));
+    }
+
+    private static void destroyChildProcesses() {
+        ProcessHandle.current().children().forEach(p -> {
+            if (p.isAlive() && !p.destroy()) {
+                p.destroyForcibly();
+            }
```
DavyLandman: add a catch to make sure the `forEach` doesn't terminate too soon.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
DavyLandman: alternative:

```java
var service = languageServers.get(lang);
if (service == null) {
   return CompletableFuture.failed(..);
}
return service;
```

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,515 @@
...
+            .thenCompose(this::route);
+    }
+
+    @Override
+    public void connect(LanguageClient client) {
+        super.connect(client); // first let the super class proxy the client
+        this.client.connect(availableClient());
+    }
+
+    private Optional<String> safeLanguage(String extension) {
```
DavyLandman: since the dedicated servers still need this, is there like  common thing we can use?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
DavyLandman: can we explain why?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
DavyLandman: nit:

```java
.map(a => a.getResolved())
.filter(r => r != null)
```

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
DavyLandman: this should be in a `try-with-resources` block I think?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
DavyLandman: looking at this, I realize that we should be using `System.err` instead. As the redirection of `System.out` is primarly a safeguard. and we want to have the raw `System.err` since log4j does use it. and also, it has flushing on println, which we want for log messages.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
DavyLandman: let's add a comment that explains that println already locks.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
DavyLandman: I think this should follow the setting we got ourselves.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
DavyLandman, line 367: a runner has a `cancel` method, to interrupt it, maybe we can use that for clean shutdown?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
DavyLandman: why is this not part of the main try block? (or the finally block?)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,515 @@
...
+            } catch (Exception e) {
+                logger.error("Unexpected error while cleaning up connection to language server for {}", lang.getName(), e);
+            }
+        });
+
+        return initializedServer; // When initialization is done, we can use the server
+    }
+
+    private int getNextPort() {
+        return REMOTE_BASE_PORT + remotePortOffset.getAndIncrement();
```
DavyLandman: isn't the `remotePortOffset`  a incorrect in case we restart a downstream server? shouldn't we keep track of the ports we connected to, and use the first port after `REMOTE_BASE_PORT` that we're not connected to right now?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
```diff
@@ -0,0 +1,416 @@
...
+        if (client == null) {
+            throw new IllegalStateException("Client not connected yet.");
+        }
+        return client;
+    }
+
+    @Override
+    public void didOpen(DidOpenTextDocumentParams params) {
+        try {
+            route(params.getTextDocument()).thenAccept(s -> s.didOpen(params)).get(10, TimeUnit.SECONDS);
```
DavyLandman: Is it intended we block on the processing of `didOpen`? if so, let's document it.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingWorkspaceService.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @Override
+    public CompletableFuture<Object> executeCommand(ExecuteCommandParams commandParams) {
+        if (commandParams.getCommand().startsWith(RASCAL_META_COMMAND) || commandParams.getCommand().startsWith(RASCAL_COMMAND)) {
+            var languageName = ((JsonPrimitive) commandParams.getArguments().get(0)).getAsString();
+            return route(languageName).thenCompose(s -> s.executeCommand(commandParams));
+        }
+
+        return CompletableFutureUtils.completedFuture(commandParams.getCommand() + " was ignored, since it is not a Rascal LSP command.", getExecutor());
```
DavyLandman, line 92: should it perphas be a response error with this message?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -129,12 +129,21 @@ private static Launcher<IBaseLanguageClient> constructLSPClient(InputStream in,
         return clientLauncher;
     }
 
-    private static void printClassPath() {
+    protected static void printClassPath() {
         logger.trace("Started with classpath: {}", () -> System.getProperty("java.class.path"));
     }
 
+    @FunctionalInterface
+    protected interface ServerConstructor<A, B, C, D, E> {
```
DavyLandman: Do the arguments need to be generic? since the class is named specifically for this use case?

DavyLandman (changes requested):
Yes, this looks good, my only consern is around the log forwarding code.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
toinehartman: I had this first, but CF made a fuss. Explicitly casting the method reference with a nullable return type solves this.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
toinehartman: This also gave troubles with CF, but together with the other changes you proposed, it works.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingWorkspaceService.java
toinehartman: This mirrors exactly what `BaseWorkspaceService` does:

https://github.com/usethesource/rascal-language-servers/blob/d8387236b585102e4ef9bae811b3362240f473f3/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseWorkspaceService.java#L191

I agree that handling this differently might be better, but we should align all workspace implementations, and I didn't want to change the existing behaviour here silently.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
toinehartman: I thought this was needed, but it looks like it's not, so reverted it.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,527 @@
...
+            .collect(Collectors.toList());
+    }
+
+    private static void prependThreadName(String prefix, JsonElement json) {
+        try {
+            var obj = json.getAsJsonObject();
+            var threadName = obj.getAsJsonPrimitive("threadName").getAsString();
+            if (threadName != null) {
+                obj.addProperty("threadName", prefix + threadName);
+            }
```
DavyLandman: what if there is no thread name? we should still set our own I think.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,527 @@
...
+            throw new IllegalStateException("Pool of dev ports is exhausted. Stop some unused servers or increase the size of the pool.");
+        }
+        try {
+            @SuppressWarnings("java:S2095") // no need to close the socket here - we close it on server shutdown
+            Socket socket = new Socket(InetAddress.getLoopbackAddress(), port);
+            socket.setTcpNoDelay(true);
+            return Triple.of(socket.getInputStream(), socket.getOutputStream(), () -> {
+                try {
+                    logger.debug("Closing socket for language {} on port {}", lang.getName(), port);
+                    socket.close();
```
DavyLandman, line 308: we should also do a `cancel` on the language client, as otherwise LSP4J prints all kinds of error messages.

DavyLandman (approved):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
toinehartman: We never cancel the communication actively from the router. Instead, we send `shutdown` or `exit` notifications, which leads to the child server terminating itself. The router waits for this termination and then cleans up, for example by closing the socket.

toinehartman:
I squashed this PR initially, while it should have been a regular merged. Manually reverted/re-did the merge in 900d81843.

PR #1070, opened 2026-04-29
Factor out text document management

toinehartman:
Factor out management of open text document contents, reducing the amount of shared code between Rascal and language-parametric LSP implementations. This work also server as preparation for #1010, which will need the same document management on yet another class.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentStateManager.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentStateManager.java
@@ -41,7 +41,6 @@ import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.function.TriConsumer;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.checkerframework.checker.initialization.qual.UnknownInitialization;
 import org.checkerframework.checker.nullness.qual.KeyFor;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.eclipse.lsp4j.Diagnostic;
@@ -72,11 +71,12 @@ public class TextDocumentStateManager implements ITextDocumentStateManager {
     private final Map<ISourceLocation, TextDocumentState> files = new ConcurrentHashMap<>();
     private final ColumnMaps columns;
 
+    @SuppressWarnings({"methodref.receiver.bound"}) // this::getContents
     public TextDocumentStateManager() {
         this.columns = new ColumnMaps(this::getContents);
     }
 
-    public String getContents(@UnknownInitialization TextDocumentStateManager this, ISourceLocation file) {
+    public String getContents(ISourceLocation file) {
         file = file.top();
         var ideState = files.get(file);
         if (ideState != null) {
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentStateManager.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentStateManager.java
@@ -72,7 +72,7 @@ public class TextDocumentStateManager {
     private final ColumnMaps columns;
 
     public TextDocumentStateManager() {
-        this.columns = new ColumnMaps(l -> getContents(l));
+        this.columns = new ColumnMaps(this::getContents);
     }
 
     public String getContents(@UnknownInitialization TextDocumentStateManager this, ISourceLocation file) {
@@ -114,7 +114,7 @@ public class TextDocumentStateManager {
         loc = loc.top();
         TextDocumentState file = files.get(loc);
         if (file == null) {
-            throw new FileNotFoundException(String.format("Unknown file: {}", loc));
+            throw new FileNotFoundException(String.format("Unknown file: %s", loc));
         }
         return file;
     }
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentStateManager.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentStateManager.java
@@ -29,25 +29,32 @@ package org.rascalmpl.vscode.lsp;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.Reader;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.function.BiFunction;
+import java.util.stream.Collectors;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.function.TriConsumer;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.checkerframework.checker.initialization.qual.UnknownInitialization;
 import org.checkerframework.checker.nullness.qual.KeyFor;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.TextDocumentItem;
+import org.eclipse.lsp4j.VersionedTextDocumentIdentifier;
 import org.eclipse.lsp4j.jsonrpc.messages.ResponseError;
 import org.eclipse.lsp4j.jsonrpc.messages.ResponseErrorCode;
 import org.rascalmpl.uri.URIResolverRegistry;
 import org.rascalmpl.util.locations.ColumnMaps;
 import org.rascalmpl.util.locations.LineColumnOffsetMap;
 import org.rascalmpl.values.parsetrees.ITree;
+import org.rascalmpl.vscode.lsp.rascal.conversion.Diagnostics;
+import org.rascalmpl.vscode.lsp.util.Versioned;
 import org.rascalmpl.vscode.lsp.util.locations.Locations;
 
 import io.usethesource.vallang.ISourceLocation;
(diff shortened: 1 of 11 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -328,23 +296,31 @@ public void initialized() {
         }
     }
 
+    private void reportDiagnostics(ISourceLocation file, Versioned<List<Diagnostics.Template>> diagnostics, List<Diagnostic> parseErrors) {
+        facts(file).reportParseErrors(file, diagnostics.version(), parseErrors);
+    }
```
DavyLandman: The name of the function and what it calls is a bit strange? It also seems to ignore the `diagnostics` parameter contents?

And why is diagnostic versioned and parseErrors not?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -711,18 +669,17 @@ private ParametricFileFacts facts(ISourceLocation doc) {
...
-        if (file == null) {
+    @Override
+    protected TextDocumentState getFile(@UnknownInitialization ParametricTextDocumentService this, ISourceLocation loc) {
+        try {
+            return super.getFile(loc);
+        } catch (FileNotFoundException e) {
             throw new ResponseErrorException(unknownFileError(loc, loc));
         }
-        return file;
+    }
```
DavyLandman: why is this not in the super class?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
```diff
@@ -317,25 +281,10 @@ public void didSave(DidSaveTextDocumentParams params) {
...
-            logger.trace("Finished parsing tree, reporting new parse errors: {} for: {}", parseErrors, file.getLocation());
-            if (facts != null) {
-                facts.reportParseErrors(file.getLocation(), parseErrors);
-            }
-        });
+    private void reportDiagnostics(ISourceLocation file, Versioned<List<Diagnostics.Template>> diagnostics, List<Diagnostic> parseErrors) {
+        if (facts != null) {
+            facts.reportParseErrors(file, parseErrors);
+        }
     }
```
DavyLandman: why is here the `diagnostic` version ignored?

and again, this function is not doing anything with the diagnostics. (I think at least the name should be changed to `reportParseErrors`  or something.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
```diff
@@ -556,20 +505,20 @@ private static <T> T last(List<T> l) {
...
-    protected TextDocumentState getFile(ISourceLocation loc) {
-        TextDocumentState file = documents.get(loc);
-        if (file == null) {
-            throw new ResponseErrorException(new ResponseError(-1, "Unknown file: " + loc, loc));
+    @Override
+    protected TextDocumentState getFile(@UnknownInitialization RascalTextDocumentService this, ISourceLocation loc) {
+        try {
+            return super.getFile(loc);
+        } catch (FileNotFoundException e) {
+            throw new ResponseErrorException(unknownFileError(loc, loc));
```
DavyLandman: looks to be the same as the one on the other class, so perhaps also move it to the super class?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentStateManager.java
```diff
@@ -0,0 +1,179 @@
...
+import org.rascalmpl.util.locations.LineColumnOffsetMap;
+import org.rascalmpl.values.parsetrees.ITree;
+import org.rascalmpl.vscode.lsp.rascal.conversion.Diagnostics;
+import org.rascalmpl.vscode.lsp.util.Versioned;
+import org.rascalmpl.vscode.lsp.util.locations.Locations;
+
+import io.usethesource.vallang.ISourceLocation;
+
+/**
+ * Management of {{@link TextDocumentState}} associated with locations.
```
DavyLandman: let's explain a bit more why this class exist, and not just repeat the class name.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentStateManager.java
```diff
@@ -0,0 +1,179 @@
...
+ *
+ *
+ */
+public class TextDocumentStateManager {
```
DavyLandman: the fact that the methods in this class implement part of an interface in `RascalTextDocumentService`  is quite subtle. 

If someone changes an interface it becomes a error on a sub class, and not on this class. Can we improve that (maybe with a smaller interface with only these methods and add them to the larger interface via inheritance?).

DavyLandman (commented):
Yes, this is good to factor this out.

I think doing this you've also discovered slight differences between the 2 classes. They are most likely the indication of bugs. Let's smooth that out.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
toinehartman: Good catch! I changed parse error reporting to take a `Versioned<List<Diagnostic>>` and lifted it most of it to the base class (777f1fd9e5216692436caa88e3f9e03d2cefccf9).

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
toinehartman: Refactored to two functions in the base class. Both throw an exception when the requested file 'does not exist' (i.e. is not open in an editor).
1. `protected getFile`, which throws an unchecked `ResponseErrorException`, for use in LSP endpoints.
2. `public getEditorState`, which throws a checked `FileNotFoundException`, for use elsewhere (e.g. in `RascalLanguageServices::makeParseTreeGetter`)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
toinehartman: See answer here: https://github.com/usethesource/rascal-language-servers/pull/1070#discussion_r3167638522

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentStateManager.java
toinehartman: Extracted to `ITextDocumentStateManager`, and now `IBaseTextDocumentService extends TextDocumentService, ITextDocumentStateManager`.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/FileFacts.java
```diff
@@ -134,7 +138,7 @@ public void close(ISourceLocation file) {
     }
 
     private interface FileFact {
-        void reportParseErrors(List<Diagnostic> msgs);
+        void reportParseErrors(Versioned<List<Diagnostic>> msgs);
```
DavyLandman, line 141: it's a bit weird that we don't also do this for typecheck errors. I guess we don't track it there?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -712,17 +661,7 @@ private ParametricFileFacts facts(ISourceLocation doc) {
...
-    }
-
-    private TextDocumentState getFile(ISourceLocation loc) {
-        loc = loc.top();
-        TextDocumentState file = files.get(loc);
-        if (file == null) {
-            throw new ResponseErrorException(unknownFileError(loc, loc));
-        }
-        return file;
+        return openFile(doc, contributions(Locations.toLoc(doc))::parsing, timestamp, exec);
```
DavyLandman: the difference is that now every open also looksup the contributions, which used to be delayed to only if it's not in the file map yet.

I think the 2nd parameter should be a closure, just to keep the whole lookup in contributions delayed?

DavyLandman (approved):
This looks fine, the tests fail, so maybe there is something there?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
toinehartman: Good catch, fixed

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/FileFacts.java
toinehartman: I agree. The issue for Rascal is that type checks are triggered by saves instead of changes, so consecutive type check results do not necessarily correspond to atomically increasing versions. Also, since the type check is not triggered by document changes, I am not sure how to fit it in this refactoring.

For parametric, instead of **type check**, we have **analysis** and **build**, which also behave in a slightly different way.

DavyLandman: ah yes, that is true. okay, that would make sense for now.

DavyLandman (approved):

PR #1074, opened 2026-05-05
Clear diagnostics of untitled files

### LINKED ISSUE #1065: Diagnostics of untitled files aren't always cleared (race)

sungshik:
This PR fixes #1065. It also adds a check after each UI test to make sure that all error diagnostics have been cleared (and it fixes a few tests that didn't do this yet).

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
@@ -347,9 +347,9 @@ public class ParametricTextDocumentService extends TextDocumentStateManager impl
             var language = safeLanguage(loc);
             if (language.isPresent()) {
                 // Special care: Cannot use method `facts`, because it throws when `get` returns `null`.
-                var factsOrNull = facts.get(language.get());
-                if (factsOrNull != null) {
-                    factsOrNull.remove(loc);
+                var facts = this.facts.get(language.get());
+                if (facts != null) {
+                    facts.remove(loc);
                 }
             }
         }
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
@@ -339,12 +339,10 @@ public class ParametricTextDocumentService extends TextDocumentStateManager impl
 
     @Override
     public void didDeleteFiles(DeleteFilesParams params) {
-        exec.submit(() -> {
-            // if a file is deleted, and we were tracking it, we remove our diagnostics
-            for (var f : params.getFiles()) {
-                availableClient().publishDiagnostics(new PublishDiagnosticsParams(f.getUri(), List.of()));
-            }
-        });
+        for (var f : params.getFiles()) {
+            var loc = Locations.toLoc(f.getUri());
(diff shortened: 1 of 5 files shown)


toinehartman (approved):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -339,12 +338,21 @@ public void didClose(DidCloseTextDocumentParams params) {
...
+        // This method might also be called when files are deleted that aren't
+        // tracked by this text document service. "Special care" is needed to
+        // silently ignore such files (without throwing exceptions).
+        for (var f : params.getFiles()) {
+            var loc = Locations.toLoc(f.getUri());
+            // Special care: Cannot use method `language`, because it throws when `safeLanguage` returns an empty optional.
+            var language = safeLanguage(loc);
+            if (language.isPresent()) {
+                // Special care: Cannot use method `facts`, because it throws when `get` returns `null`.
+                var factsOrNull = facts.get(language.get());
```
DavyLandman: nit: `factsOrNull`  seems a bit like a weird name? wouldn't `fact` be enough?

DavyLandman (approved):

PR #1085, opened 2026-05-20
Remove empty shutdown functions.

toinehartman:
Remove functions that have been empty/unused since #1067.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
@@ -298,10 +298,7 @@ public abstract class BaseLanguageServer {
 
         @Override
         public CompletableFuture<Object> shutdown() {
-            return CompletableFuture.supplyAsync(() -> {
-                lspDocumentService.shutdown();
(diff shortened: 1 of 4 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -298,10 +298,7 @@ public void initialized(InitializedParams params) {
 
         @Override
         public CompletableFuture<Object> shutdown() {
-            return CompletableFuture.supplyAsync(() -> {
-                lspDocumentService.shutdown();
-                return true;
-            }, executor);
+            return CompletableFuture.completedFuture(new Object());
```
DavyLandman, line 301: do we still need to implement this?

DavyLandman (approved):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
toinehartman: Yes, it comes from the `LanguageServer` interface without a default.

PR #1089, opened 2026-05-26
Extract test-only contributions from demo module.

### LINKED ISSUE #363: Simplify example by splitting it from the test case
Based on this comment in the PR:

> Ok, but this is demo code meant to educate newbies; so let's move that to test code and keep it simple here.

_Originally posted by @jurgenvinju in https://github.com/usethesource/rascal-language-servers/pull/357#discussion_r1515992791_

toinehartman:
Extract some elements from the Pico demo language server
- Syntax extensions in a separate module.
- Test-only commands in a separate module.

Closes #363

### DIFF
--- a/rascal-lsp/src/main/rascal/library/demo/lang/pico/LanguageServer.rsc
+++ b/rascal-lsp/src/main/rascal/library/demo/lang/pico/LanguageServer.rsc
@@ -55,7 +55,7 @@ Each ((LanguageService)) for pico is implemented as a function.
 Here we group all services such that the LSP server can link them
 with the ((LanguageServer-Language)) definition later.
 }
-set[LanguageService] picoLanguageServer(bool allowRecovery) = {
+private set[LanguageService] picoLanguageServer(bool allowRecovery) = {
     parsing(picoParser(allowRecovery), usesSpecialCaseHighlighting = false),
     documentSymbol(picoDocumentSymbolService),
     codeLens(picoCodeLenseService),
@@ -70,8 +70,7 @@ set[LanguageService] picoLanguageServer(bool allowRecovery) = {
     completion(picoCompletionService, additionalTriggerCharacters = ["="])
 };
 
-set[LanguageService] picoLanguageServer() = picoLanguageServer(false);
-set[LanguageService] picoLanguageServerWithRecovery() = picoLanguageServer(true);
+set[LanguageService] picoLanguageServer() = picoLanguageServer(true);
 
 @synopsis{This set of contributions runs slower but provides more detail.}
 @description{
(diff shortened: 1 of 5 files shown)


rascal-lsp/src/main/rascal/library/demo/lang/pico/LanguageServer.rsc
```diff
@@ -76,7 +66,6 @@ set[LanguageService] picoLanguageServer(bool allowRecovery) = {
     rename(picoRenamingService, prepareRenameService = picoRenamePreparingService),
     didRenameFiles(picoFileRenameService),
     selectionRange(picoSelectionRangeService),
-    callHierarchy(picoPrepareCallHierarchy, picoCallsService),
```
DavyLandman, line 79: I would prefer to have every service at least once in this file.

toinehartman: Me too. However, call hierarchies require an extension to the Pico syntax, which feels like a strange thing to do in an example language server, so I opted to extract that as well. Maybe that's too aggressive?

DavyLandman: yes, I think that's too aggressive of a cleanup.

but we could move the language extension itself to a dedicated module and extend that, such that this module is smaller?

toinehartman: Done

rascal-lsp/src/main/rascal/library/demo/lang/pico/LanguageServer.rsc
```diff
@@ -65,7 +55,7 @@ Each ((LanguageService)) for pico is implemented as a function.
 Here we group all services such that the LSP server can link them
 with the ((LanguageServer-Language)) definition later.
 }
-set[LanguageService] picoLanguageServer(bool allowRecovery) = {
+private set[LanguageService] picoLanguageServer(bool allowRecovery) = {
```
DavyLandman: how about we drop this `allowRecovery`  parameter, and instead call this function in the testing language server, and replace the `parsing` contribution in the language server?

Such that for the demo, it's always on, and not a parameter, and we for testing overwrite that?

DavyLandman (commented):
Looks good, I have one more idea to remove a detail from the demo code.

rascal-lsp/src/main/rascal/library/demo/lang/pico/LanguageServer.rsc
toinehartman: Updated. A downside is that this calls `ParseTree::parser` twice, discarding the first result.

DavyLandman: true, but there are caches for the 2nd call.

PR #1090, opened 2026-05-27
POM-leading for DSLs

### LINKED ISSUE #1010: Adapt VS Code Parametric Language server to become a forwarder and start a new actual parametric server per rascal version

toinehartman:
Long-running feature PR, comprising of several sub-PRs for reviewability.

- #1069
- https://github.com/usethesource/rascal-language-servers/pull/1109
- https://github.com/usethesource/rascal-language-servers/pull/1095
- #1112
- https://github.com/usethesource/rascal-language-servers/pull/1132
- https://github.com/usethesource/rascal-language-servers/pull/1117
- https://github.com/usethesource/rascal-language-servers/pull/1156

Closes #1010.


rascal-vscode-extension/.vscode/launch.json
```diff
@@ -37,6 +37,22 @@
...
+    {
+			"name": "Run Extension (deployed)",
+			"type": "extensionHost",
+			"request": "launch",
+			"args": [
+				"--disable-extensions",
+				"--extensionDevelopmentPath=${workspaceFolder}"
+			],
+			"outFiles": [
+				"${workspaceFolder}/out/**/*.js"
```
sungshik:
```suggestion
				"${workspaceFolder}/dist/**/*.js"
```
(Not sure about this one; I forget where exactly the files are written. Configuring the right location here matters for debugger support IIRC.)

rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
```diff
@@ -86,6 +86,16 @@ parameterizedDescribe(function (errorRecovery: boolean) {
         protectedFiles = await ProtectedFiles.protect(TestWorkspace.picoFile);
     });
 
+    after(async() => {
```
sungshik, line 89: It could be useful for future maintainers to add a comment to explain why it is needed at the end of the test suite to open a seemingly arbitrary module, import the Pico language server, but then immediately unregister the language.

rascal-lsp/.vscode/launch.json
```diff
@@ -28,6 +28,50 @@
...
+      "vmArgs": [
+        "-Dlog4j2.level=TRACE",
+        "-Drascal.compilerClasspath=${workspaceFolder}/target/lib/rascal.jar",
+        "-Drascal.remoteResolverRegistryPort=8889",
+        "-Drascal.customRemoteResolverRegistryClass=org.rascalmpl.vscode.lsp.uri.jsonrpc.VSCodeFileSystemInRascal"
+      ]
+    },
+    {
+      "type": "java",
+      "name": "Delegate Parametric Server [1]",
```
sungshik, line 48: Sometimes throughout this PR, the "children" of the routing server are called "delegate" (such as here). Sometimes, they are called "remote". Consider to use only one term.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/log/LogJsonConfiguration.java
```diff
@@ -63,19 +63,24 @@ public Configuration getConfiguration(LoggerContext loggerContext, Configuration
...
-    private Configuration buildConfiguration() {
-        Level targetLevel = Level.getLevel(System.getProperty("log4j2.level", "INFO"));
+    public static Level getLogLevel() {
+        var targetLevel = Level.getLevel(System.getProperty("log4j2.level", "INFO"));
         if (targetLevel == null) {
-            targetLevel = Level.INFO;
+            return Level.INFO;
         }
+        return targetLevel;
+    }
```
sungshik, line 72: Nit: It feels a bit excessive, but should this method be moved to a separate utility class? It is used also in `LogRedirectConfiguration`, which looks otherwise unrelated to `LogJsonConfiguration`.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/AbstractDynamicCapability.java
```diff
@@ -150,7 +150,8 @@ public boolean equals(@Nullable Object obj) {
         }
         var other = (AbstractDynamicCapability<?>) obj;
         return Objects.equals(id, other.id)
-            && Objects.equals(methodName, other.methodName);
+            && Objects.equals(methodName, other.methodName)
+            && Objects.equals(preferStaticRegistration, other.preferStaticRegistration);
```
sungshik, line 154: `preferStaticRegistration` isn't included yet in the computation of the hash code (not sure if it should; just an observation)

rascal-lsp/src/main/rascal/library/demo/lang/pico/LanguageServer.rsc
```diff
@@ -311,6 +311,18 @@ list[CompletionItem] picoCompletionService(Focus focus, int cursorOffset, Comple
     return sort(items, bool(CompletionItem i1, CompletionItem i2) {return i1.label < i2.label; });
 }
 
+PathConfig getPicoPathConfig() {
```
sungshik, line 314: This function seems to be useful especially for testing purposes instead of demo-ing/how-to-write-a-language-server purposes. If so, then consider to move it to `testing::lang::pico::LanguageServer` instead. If not, then it could be useful to add a bit of `@synopsis` to clarify what/why exactly this function does, so users get a better idea of how to apply it to their own DSLs.

rascal-lsp/src/main/rascal/library/testing/lang/json2/LanguageServer.rsc
```diff
@@ -0,0 +1,128 @@
...
+import util::PathConfig;
+import util::Reflective;
+
+private Tree (str _input, loc _origin) jsonParser(bool allowRecovery) {
+    return ParseTree::parser(#start[JSONText], allowRecovery=allowRecovery, filters=allowRecovery ? {createParseErrorFilter(false)} : {});
+}
+
+set[LanguageService] jsonLanguageServer() = {
+    parsing(jsonParser(true), usesSpecialCaseHighlighting = false),
+    completion(jsonCompletion, additionalTriggerCharacters = [",", "{", "[", ":"])
```
sungshik, line 45: It isn't clear to me what is the purpose of the completion service in the context of this PR. A small comment to clarify could be useful.

At the same time, the style in which the completion service is defined, is quite different from the completion service in `demo::lang::pico::LanguageService`. It could be useful to "promote" (parts of) `JSON2` to `demo::lang::json2::LanguageService` (which currently doesn't exist yet), so it is easier/more predictable to find by users who want to implement their own completion service and are looking for examples.

rascal-lsp/src/main/rascal/library/testing/lang/pico/LanguageServer.rsc
sungshik, line 68: Nit: Should this function also be renamed?

rascal-lsp/src/main/rascal/library/testing/lang/pico/LanguageServer.rsc
```diff
@@ -112,6 +118,7 @@ lrel[loc, Command] testingCodeLensService(start[Program] input)
         <declOffset(input, 0), removeTodo(input.src, title="Unregister TODO")>,
         <declOffset(input, 0), showWarning("Test warning", input.src, title="Show warning")>,
         <declOffset(input, 0), showContents("Some text", title="Show some text")>,
+        <declOffset(input, 1), copyFileContents(input.top.src.parent.parent + "json2" + "example.json2", input.top.src.parent.parent + "json2" + "example-copy.json2", title="Copy contents of example.json2")>,
```
sungshik, line 121: Why is this JSON2-related command in the Pico language server?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentStateManager.java
```diff
@@ -205,7 +208,19 @@ protected void closeFile(ISourceLocation loc) {
     }
 
     protected Set<@KeyFor("this.files") ISourceLocation> getOpenFiles() {
-        return files.keySet();
+        return Collections.unmodifiableSet(files.keySet());
```
sungshik, line 215: Not sure this is relevant, but since the unmodifiability seems important here... IIRC, `unmodifiableSet` doesn't make a shallow copy of the underlying collection. So, the returned set is unmodifiable by the caller, but if `this.files` somehow changes via a different route, then I expect the changes to be visible in the returned set (i.e., from the caller's perspective, it appears the returned set has spontaneously changed, even though it looks unmodifiable to it).

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -144,12 +150,21 @@ private static Launcher<IBaseLanguageClient> constructLSPClient(InputStream in,
...
     }
 
-    private static void printClassPath() {
+    protected static void printClassPath() {
         logger.trace("Started with classpath: {}", () -> System.getProperty("java.class.path"));
     }
 
+    @FunctionalInterface
+    protected interface ServerBuilder {
+        ActualLanguageServer apply(String name, Runnable a, ExecutorService b, IBaseTextDocumentService c, BaseWorkspaceService d);
```
sungshik: Nit: Use more descriptive names for the arguments? Perhaps even add a bit of JavaDoc to explain how these four main components of a language server fit together.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -291,13 +310,43 @@ public CompletableFuture<Two<String, URI[]>[]> supplyPathConfig(PathConfigParame
...
+                if (specVersion == null) {
+                    throw new IOException(String.format("No Specification-Version in %s", manifest));
+                }
+                return specVersion;
+            }
+        }
+
         @Override
-        public CompletableFuture<Void> sendRegisterLanguage(LanguageParameter lang) {
+        public CompletableFuture<Void> sendRegisterLanguage(RegistrationParameter param) {
```
sungshik, line 352: Nit: The registration parameter is used only to get the language parameter. Perhaps it is possible to revert the signature of `sendRegisterLanguage` back to the way it was (with `LanguageParameter`) to keep the symmetry with `sendUnregisterLanguage`?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -291,13 +310,43 @@ public CompletableFuture<Two<String, URI[]>[]> supplyPathConfig(PathConfigParame
             }, executor);
         }
 
+        protected static String getPomVersion() throws IOException {
```
sungshik, line 324: There are two causes why an `IOException` might be thrown by this method:
  - "No package" -- According to the comment, this shouldn't happen, so it is really an exceptional situation.
  - "No specification version found" -- This seems to be a situation that could be considered normal/acceptable control flow (judging from the call site of this method).

I think it would make sense to be able to distinguish between these situations at the call site. For the former case, an exception seems appropriate and an ERROR message could be printed to the log. For the latter case, it seems a `ServerInfo` object can normally be created, but perhaps printing an INFO message to the log could be useful. If so, then instead of throwing an exception, a special return value could be used (e.g., just `null` or an `Optional`).

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -291,13 +310,43 @@ public CompletableFuture<Two<String, URI[]>[]> supplyPathConfig(PathConfigParame
...
+        }
+
+        protected static String getJarVersion(ISourceLocation jarFile) throws IOException {
+            jarFile = JarURIResolver.jarify(jarFile);
+            var manifestLoc = URIUtil.getChildLocation(jarFile, "META-INF/MANIFEST.MF");
+            try (var is = URIResolverRegistry.getInstance().getInputStream(manifestLoc)) {
+                var manifest = new Manifest(is);
+                var specVersion = manifest.getMainAttributes().getValue("Specification-Version");
+                if (specVersion == null) {
+                    throw new IOException(String.format("No Specification-Version in %s", manifest));
```
sungshik, line 345: Same comment as above probably

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/DocumentRouter.java
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * Map an {@link ISourceLocation} to a {@link T}.
+     * @param loc The input location.
+     * @return The mapped value.
+     */
+    T route(ISourceLocation loc);
+
+    /**
+     * Map a {@link String} name to a {@link T}.
+     * @param doc The name key.
```
sungshik:
```suggestion
     * @param name The name key.
```

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/DocumentRouter.java
```diff
@@ -0,0 +1,76 @@
...
+
+    default T route(TextDocumentIdentifier id) {
+        return route(Locations.toLoc(id.getUri()));
+    }
+
+    Stream<T> allRoutes();
+
+    default <U> Stream<U> allRoutes(Function<T, U> f) {
+        return allRoutes().map(f);
+    }
```
sungshik: Nit: I wonder if this function is needed. I think it is more clear (for a reader who doesn't know this class) to have the explicit `allRoutes().map(f)` at the call site instead of `allRoutes(f)` (because to know the purpose of `f`, the reader needs to look at the implementation of `allRoutes(Function)`).

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/DocumentRouter.java
```diff
@@ -0,0 +1,76 @@
...
+ * @param <T> The type of the mapped value.
+ */
+public interface DocumentRouter<T> {
+
+    /**
+     * Map an {@link ISourceLocation} to a {@link T}.
+     * @param loc The input location.
+     * @return The mapped value.
+     */
+    T route(ISourceLocation loc);
```
sungshik, line 48: I would expect there to be exactly one `default`-less `route` method (probably the one with the `String` argument), in terms of which all other `route` methods are defined. Otherwise, implementations of this interface could be (morally) inconsistent between `route(ISourceLocation)` and `route(String)`.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/concurrent/CompletableFutureUtils.java
```diff
@@ -34,17 +34,42 @@
...
         return CompletableFuture.supplyAsync(() -> value, exec);
     }
 
+    public static <T> CompletableFuture<T> retry(Supplier<T> supplier, int times, Executor exec) {
+        return retry(CompletableFuture.supplyAsync(supplier, exec), times);
+    }
+
+    public static <T> CompletableFuture<T> retry(CompletableFuture<T> future, int times) {
+        var f = future;
+        for (int i = 1; i <= times; i++) {
```
sungshik, line 60:
```suggestion
        for (int i = 1; i <= times && !f.isDone(); i++) {
```
(I think this avoids redundant iterations in which the same result is re-wrapped in completed futures?)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/model/ParametricFileFacts.java
```diff
@@ -67,7 +66,7 @@ public class ParametricFileFacts implements DiagnosticsReporter {
     private final Map<ISourceLocation, FileFact> files = new ConcurrentHashMap<>();
 
     @SuppressWarnings("java:S3077") // Reads/writes happen sequentially
-    private volatile @MonotonicNonNull LanguageClient client;
+    private volatile LanguageClient client;
```
sungshik, line 69: It can now be `final`?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/CapabilityRegistration.java
sungshik, line 1: Changes in this file were a bit hard for me to understand (no suggestions; just a comment)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/LanguageRegistry.java
```diff
@@ -55,6 +55,22 @@ public interface LanguageRegistry {
     @JsonRequest("receiveUnregisterLanguage")
     CompletableFuture<Void> unregisterLanguage(LanguageParameter lang);
 
+    /**
+     * Extensible arguments related to language registration.
+     */
+    class RegistrationParameter {
+        private LanguageParameter lang;
+        private String remoteMemoryArg;
```
sungshik, line 63: It could be useful to document what exactly this property is used for (and along the way remind the reader that, when a language is registered, a new server process is spawned etc.).

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricLanguageServer.java
```diff
@@ -28,24 +28,75 @@
...
+
+        public @Nullable LanguageParameter getDedicatedLanguage() {
+            return dedicatedLanguage;
+        }
+
+        public void setDedicatedLanguage(LanguageParameter dedicatedLanguage) {
+            this.dedicatedLanguage = dedicatedLanguage;
+        }
+
+        public boolean isExitWhenEmpty() {
```
sungshik: Nit: The `is...` naming sounds a bit unnatural for instances of this class. Maybe use `get...` instead?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricLanguageServer.java
```diff
@@ -28,24 +28,75 @@
...
+            switch (args[i]) {
+                case "--port":
+                    serverArgs.setPort(Integer.parseInt(args[++i]));
+                    break;
+                case "--exitWhenEmpty":
+                    serverArgs.setExitWhenEmpty(true);
+                    break;
+                default:
+                    if (serverArgs.getDedicatedLanguage() == null) {
+                        serverArgs.setDedicatedLanguage(new GsonBuilder().create().fromJson(args[i], LanguageParameter.class));
```
sungshik, line 95: I suppose this call can ungracefully fail with a runtime exception when `args[i]` isn't of the right format. Could this be a problem at the call sites of `parseArgs`?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricLanguageServer.java
```diff
@@ -28,24 +28,75 @@
...
     }
+
+    public static void main(String[] args) {
+        startParametric(parseArgs(args));
+    }
+
+    public static class ServerArgs {
+        private int port = 9999;
+        private @Nullable LanguageParameter dedicatedLanguage = null;
+        private boolean exitWhenEmpty = false;
```
sungshik, line 54: Emptiness of what?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -326,7 +341,9 @@ public void didClose(DidCloseTextDocumentParams params) {
         logger.debug("Did Close file: {}", params.getTextDocument());
         var loc = Locations.toLoc(params.getTextDocument());
         closeFile(loc);
-        facts(loc).close(loc);
+        if (safeLanguage(loc).isPresent()) {
```
sungshik, line 345: If no language can be found for `loc`, then is this a sign something is wrong (in which case it could be useful to print a message to the log for debugging).

sungshik: Other question: Is it possible for `safeLanguage(loc).isPresent()` to be false, but `URIResolverRegistry.getInstance().exists(loc)` (condition of the next if-then below) to be true? I expected the next if-then to be part of the body of the current if-then.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -882,11 +909,12 @@ public CompletableFuture<Either<List<CompletionItem>, CompletionList>> completio
     public synchronized void registerLanguage(LanguageParameter lang) {
         logger.info("registerLanguage({})", lang.getName());
 
+        var clientCopy = (IBaseLanguageClient) availableClient();
```
sungshik, line 913: Nit: Why not just `client`?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -1001,10 +1027,14 @@ private static String buildContributionKey(LanguageParameter lang) {
         return lang.getMainFunction() + "::" + lang.getMainFunction();
     }
 
+    public static boolean isLanguageCompletelyRemoved(LanguageParameter lang) {
```
sungshik: Nit: Removed from what? I feel the name of this method isn't well-aligned with its implementation. My impression is this method checks if the language is *ready to be* removed (as there is no meaningful main module), but not actually removed from the collection of interest yet. Instead, only when this method returns true, the caller can perform the actual removal.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -1031,7 +1061,12 @@ public synchronized void unregisterLanguage(LanguageParameter lang) {
         }
 
         // Should be called from the main, single-threaded request pool
-        updateCapabilities();
+        updateCapabilities().thenAccept(v -> {
+            if (exitWhenEmpty && contributions.isEmpty()) {
+                logger.debug("Shutting down; no more registered languages");
```
sungshik:
```suggestion
                logger.debug("Shutting down; no more registered contributions for language " + dedicatedLanguageName);
```
(Or something along those lines. I got confused for a moment by reading this error message because of the difference between "languages" and "contributions".)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -1031,7 +1061,12 @@ public synchronized void unregisterLanguage(LanguageParameter lang) {
...
+                System.exit(0);
```
sungshik, line 1068: Killing the whole process from relatively deep inside it looks a bit concerning to me. (Maybe it isn't needed currently, but in general, I imagine it could be useful to be able to perform pre-exit actions higher up, such as closing resources, printing logs, writing to caches on disk for the next session, etc.). I think it would be better to manage the lifecycle of the process as close as possible to its origin.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
```diff
@@ -0,0 +1,404 @@
...
+    public void didClose(DidCloseTextDocumentParams params) {
+        closeFile(Locations.toLoc(params.getTextDocument()));
+
+        // Inform all remote servers about this file, so they can maintain its state.
+        notifyAllRemotes(TextDocumentService::didClose, params);
+    }
+
+    @Override
+    public void didSave(DidSaveTextDocumentParams params) {
+        // Contrary to `didOpen`, `didChange` and `didClose`, inform only the remote server for this language, since this does not change file state
```
sungshik, line 220: It isn't completely clear to me why a server for language X needs to be aware of opens/changes/closes of files in language Y.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
```diff
@@ -0,0 +1,404 @@
...
+
+    private <P, R> CompletableFuture<R> route(TextDocumentIdentifier doc, BiFunction<TextDocumentService, P, CompletableFuture<R>> endpoint, P params) {
+        return routeCompose(route(doc), endpoint, params);
+    }
+
+    private <P, R> CompletableFuture<R> route(ISourceLocation loc, BiFunction<TextDocumentService, P, CompletableFuture<R>> endpoint, P params) {
+        return routeCompose(route(loc), endpoint, params);
+    }
+
+    private <P, R> CompletableFuture<R> routeCompose(TextDocumentService sf, BiFunction<TextDocumentService, P, CompletableFuture<R>> endpoint, P params) {
```
sungshik: Nit: I think this is the first time the term "endpoint" is used. Is it intentional (i.e., does it have a specific technical meaning in this context)? If so, then consider to document it explicitly.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
```diff
@@ -0,0 +1,404 @@
...
+
+    @Override
+    protected DiagnosticsReporter getDiagnosticsReporter(ISourceLocation file) {
+        // NOP; we delegate diagnostic reporting to our dedicated remote servers
+        return (l, msgs) -> {};
+    }
+
+    @Override
+    public void didCreateFiles(CreateFilesParams params) {
+        // NOP; RoutingWorkspaceService nevers calls us, but forwards to remotes instead
```
sungshik:
```suggestion
        // RoutingWorkspaceService never calls us, but forwards to remotes instead
```

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
```diff
@@ -0,0 +1,404 @@
...
+        throw unsupportedRoutingOperation("didCreateFiles");
+    }
+
+    @Override
+    public void didRenameFiles(RenameFilesParams params, List<WorkspaceFolder> workspaceFolders) {
+        // NOP; RoutingWorkspaceService nevers calls us, but forwards to remotes instead
```
sungshik:
```suggestion
        // RoutingWorkspaceService never calls us, but forwards to remotes instead
```

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
```diff
@@ -0,0 +1,404 @@
...
+        throw unsupportedRoutingOperation("didRenameFiles");
+    }
+
+    @Override
+    public void didDeleteFiles(DeleteFilesParams params) {
+        // NOP; RoutingWorkspaceService nevers calls us, but forwards to remotes instead
```
sungshik:
```suggestion
        // RoutingWorkspaceService never calls us, but forwards to remotes instead
```

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ProxiedIValue.java
```diff
@@ -0,0 +1,84 @@
...
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
+import java.io.IOException;
+
+import io.usethesource.vallang.IExternalValue;
+import io.usethesource.vallang.IValue;
+import io.usethesource.vallang.type.Type;
+
+/**
+ * Wraps a JSON element representing an IValue as an IValue.
```
sungshik: "IValue as an IValue" -- This sounds a bit circular 😉.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingLanguageServer.java
```diff
@@ -0,0 +1,56 @@
...
+public class RoutingLanguageServer extends ParametricLanguageServer {
+
+    @SuppressWarnings("java:S9149") // hides `ParametricLanguageServer::main`
+    public static void main(String[] args) {
+        var serverArgs = parseArgs(args);
+        if (serverArgs.getDedicatedLanguage() != null) {
+            // If we get a dedicated language argument, we just start a single parametric server
+            startParametric(serverArgs);
+        } else {
+            startLanguageServer(
```
sungshik: Nit: Seeing these calls side by side, consider to give the methods more descriptive names that reflect their intent/differences

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -0,0 +1,343 @@
...
+
+    private static final Logger logger = LogManager.getLogger(MultipleClientProxy.class);
+    private static final Supplier<CompletableFuture<Map<Object, List<Registration>>>> EMPTY_REGISTRATIONS = () -> CompletableFuture.completedFuture(new HashMap<>());
+
+    private final IBaseLanguageClient client;
+    private final ExecutorService exec;
+
+    /**
+     * The current registrations.
+     * Map of method names to current registrations.
```
sungshik, line 85: I was expecting registrations to be associated with contributions instead of methods. Perhaps a high-level description of the design of the routing process (both in server role and in client role) could be useful.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -0,0 +1,343 @@
...
+     * The inner map is keyed by registration options, with a list of registrations with those exact options. The first registration in this list is always registered with the actual client, while the others are kept for internal administration.
```
sungshik, line 86: This sounds a bit ad-hoc. Consider introducing a separate class to represent a client + bookkeeping data structures.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -0,0 +1,343 @@
...
+     */
+    private final Map<String, CompletableFuture<Map<Object, List<Registration>>>> registrations = new ConcurrentHashMap<>();
```
sungshik, line 88: Isn't there a more precise type than `Object`?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -0,0 +1,343 @@
...
+import io.usethesource.vallang.IInteger;
+import io.usethesource.vallang.IString;
+
+/**
+ * Client proxy implementation that aggregates results from multiple servers before forwarding to its own client.
+ */
+public class MultipleClientProxy implements IBaseLanguageClient {
+
+    private static final Logger logger = LogManager.getLogger(MultipleClientProxy.class);
+    private static final Supplier<CompletableFuture<Map<Object, List<Registration>>>> EMPTY_REGISTRATIONS = () -> CompletableFuture.completedFuture(new HashMap<>());
```
sungshik, line 78: This field seems to be used in only one place. Consider to inline it there. (Reading the file from top to bottom, it isn't clear/obvious what the purpose of this field is, nor why it has such an unexpected type. I think this is better understandable at the place of usage.)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
sungshik: Update: I think my confusion is about registration of capabilities (what this class is about) vs registration of contributions (what `registerLanguage` etc. is about, which I read earlier today). This could be better reflected in the JavaDoc, and also in the name of class `Registration` (e.g., rename to `CapabilityRegistration`) to avoid confusion.

sungshik: Oh, `Registration` is an Lsp4j class. Never mind the rename then.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -0,0 +1,343 @@
...
+    /**
+     * This method is responsible for managing the registrations from remotes.
+     *
+     * Since we cannot register a single capability with the same options multiple times, and remotes do not know about each others' capabilities,
+     * this method (together with `unregisterCapability`) makes sure that the capabilities registered with the client are the sum of the
+     * capabilities registered by the remote servers.
+     */
+    private CompletableFuture<Void> registerCapability(Registration r) {
+        logger.trace("Incoming registration request for {}", r.getMethod());
+        var res = registrations.compute(r.getMethod(), (method, f) ->
```
sungshik, line 240: Nit: Why is `f` called like that? `m` (for "map") would be more descriptive?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -0,0 +1,343 @@
...
+                    // However, we do write down this registration for our own administration, in case we need it later.
+                    logger.trace("This exact capability was registered with the client before - we ignore it for now: {}", r);
+                    equalOptRegs.add(r);
+                    return CompletableFuture.completedStage(currentRegs);
+                }
+
+                logger.trace("Registering {} with the client: {}", method, r);
+                return client.registerCapability(new RegistrationParams(List.of(r)))
+                    .thenAccept(v -> equalOptRegs.add(r))
+                    .handle((v, t) -> {
```
sungshik, line 255: Nit: Isn't it possible to check first for an exception and add `r` to `equalOptRegs` only when none has been thrown (instead of always adding and then possibly removing)?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -0,0 +1,343 @@
...
+import org.rascalmpl.uri.remote.jsonrpc.ISourceLocationChanged;
+import org.rascalmpl.vscode.lsp.IBaseLanguageClient;
+import org.rascalmpl.vscode.lsp.parametric.LanguageRegistry.LanguageParameter;
+import org.rascalmpl.vscode.lsp.util.concurrent.CompletableFutureUtils;
+
+import io.usethesource.vallang.IInteger;
+import io.usethesource.vallang.IString;
+
+/**
+ * Client proxy implementation that aggregates results from multiple servers before forwarding to its own client.
```
sungshik, line 73: My impression is this class works as intended only in the absence of concurrency (i.e., no races between registrations etc.). If so, then this could be documented here.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -0,0 +1,343 @@
...
+            .thenAccept(v -> {}); // convert to Void
+    }
+
+    private boolean matches(Registration r, Unregistration u) {
+        return r.getId().equals(u.getId())
+            && r.getMethod().equals(u.getMethod());
+    }
+
+    private CompletableFuture<Void> unregisterCapability(Unregistration u) {
+        var res = registrations.compute(u.getMethod(), (method, f) ->
```
sungshik, line 286:
```suggestion
        var res = registrations.computeIfPresent(u.getMethod(), (method, f) ->
```
Not sure about this suggestion, but the current code seems to insert an entry in `registrations` even if `u` is an unregistration for a method that isn't in the map yet, which looks a bit unexpected.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -0,0 +1,343 @@
...
+            computeIfAbsent(f).thenCompose(currentRegs -> {
```
sungshik, line 287: If `f` is null, then return immediately instead?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
sungshik: Update: Oh, I missed it is a future! Perhaps call it `currentRegsFuture` to also link it to `currentRegs` on the next line? I realize it is a bit lengthy, but it is non-trivial code, and I think the wordiness is worth it in this case.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
sungshik: (Also same comment about the name of `f` as above)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -0,0 +1,343 @@
...
+                for (var entry : currentRegs.entrySet()) {
```
sungshik, line 288: Nit:
```suggestion
                for (var regs : currentRegs.values()) {
```
(And update code below accordingly)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -0,0 +1,343 @@
...
+                    var unreg = entry.getValue().stream().filter(r -> matches(r, u)).findAny();
+                    if (!unreg.isPresent()) {
+                        continue;
+                    }
+
+                    var regs = entry.getValue();
+                    var idx = regs.indexOf(unreg.get());
```
sungshik, line 295: Nit: I think (the intent of) the code is easier to follow if a helper function `findIndex` were to be defined (inspired by [`findIndex`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) in JS), along the lines of:

```java
public static <T> int findIndex(List<T>, Predicate<T> predicate) {
    var found = list.stream().filter(predicate).findAny();
    return found.isPresent() ? list.indexOf(found.get()) : -1;
}
```

Then, the body of the loop can be refactored along the lines of:

```suggestion
                    var idx = findIndex(regs, r -> matches(r, u));
                    switch (idx) {
                        case -1: continue;
                        case 0: // `r` is registered with the client
                            ...
                        default: // `r` isn't registered with the client
                            ...
                    }
```

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,641 @@
...
+public class ActualRoutingLanguageServer extends BaseLanguageServer.ActualLanguageServer implements DocumentRouter<IBaseLanguageServerExtensions> {
+
+    static class NoLanguageException extends RuntimeException {
+        public NoLanguageException(String message) {
+            super(message);
+        }
+    }
+
+    private static final String THREAD_NAME_KEY = "threadName";
+    private static final ComparableVersion MINIMAL_COMPATIBLE_VERSION = new ComparableVersion("2.22.6-SNAPSHOT");
```
sungshik, line 126: Consider to add a comment to explain why this is needed

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
sungshik: Update: Hm, I am not sure if the "absence of concurrency" assumption is true. The executor service `exec` seems to be used in two places: `registerCapability` and `unregisterCapability`. Inside, these methods call `CompletableFuture.supplyAsync`. It haven't been able to think it all the way through, but at least this suggests there is some asynchronous processing going on as part of each `registerCapability` or `unregisterCapability` call. Even if the thread pool consists of only one thread, I think this means that such calls aren't necessarily executed atomically. As a result, for instance, it might be possible for two concurrent `unregisterCapability` calls to interfere with each other? The use of several completable futures, which might not always be related to each other, and of which it isn't obvious to me by which thread they are completed, makes it hard for me to see if this really is a problem, but also, it isn't clear to me it isn't.

My suggestion is to document somewhere, while the code is still fresh in our minds, what kind of concurrency can be expected, which relevant invariants should always be true (regardless of the interleavings), and possibly on specific lines of the code some "reassuring/comforting" small comments to explain why the invariants won't be violated.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,641 @@
...
+    // NOTE To be able to route to arbitrary third-party language servers, remote servers should implement `LanguageServer` (instead of `IBaseLanguageServerExtensions`)
+    private final Map<String, IBaseLanguageServerExtensions> languageServers = new ConcurrentHashMap<>();
+    private final Map<String, String> languagesByExtension = new ConcurrentHashMap<>();
+
+    private @MonotonicNonNull MultipleClientProxy remoteClient;
+    private @MonotonicNonNull InitializeParams initializeParams;
+    private final JsonWriter logForwarder;
+
+    private static final int REMOTE_BASE_PORT = 9990;
+    private static final int PORT_POOL_SIZE = 9;
```
sungshik: Should these be hard-coded?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,641 @@
...
+        }
+
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> destroyChildProcesses()));
+    }
+
+    private static void destroyChildProcesses() {
+        ProcessHandle.current().children().forEach(p -> {
+            try {
+                if (p.isAlive() && !p.destroy()) {
+                    p.destroyForcibly();
```
sungshik, line 164: Isn't there a less "forcible" way to do this? E.g., send a shutdown message to the process, but make the process responsible to exit (so it can do some cleanup etc. if needed)?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,641 @@
...
+        var service = languageServers.get(lang);
+        if (service == null) {
+            throw new NoLanguageException(String.format("Rascal Parametric LSP has no support for this file, since no language is registered with name '%s'", lang));
+        }
+        return service;
+    }
+
+    @Override
+    public Stream<IBaseLanguageServerExtensions> allRoutes() {
+        return languageServers.values().parallelStream();
```
sungshik, line 183: Why a parallel stream?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,641 @@
...
+        }
+
+        throw new IOException("No rascal-lsp dependency. Could not start language server");
+    }
+
+    private static void prependThreadName(String langName, JsonElement json) {
+        try {
+            var obj = json.getAsJsonObject();
+            obj.addProperty(THREAD_NAME_KEY, langName + (obj.has(THREAD_NAME_KEY) ? " | " + obj.getAsJsonPrimitive(THREAD_NAME_KEY).getAsString() : ""));
+        } catch (Exception e) { /* ignored */ }
```
sungshik: Why ignore?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,641 @@
...
+        try {
+            var json = JsonParser.parseString(line);
+            prependThreadName(langName, json);
+            // Lock, so we can make sure our JSON is followed by a newline.
+            synchronized (DEPLOYMENT_OUTPUT_STREAM) {
+                gson.toJson(json, logForwarder);
+                logForwarder.flush();
+                // One object per line; this is what log4j does as well.
+                DEPLOYMENT_OUTPUT_STREAM.println();
+            }
```
sungshik: Consider to move this out of the try block, so it is clear that the exception doesn't occur during the synchronized block.

Also, `DEPLOYMENT_OUTPUT_STREAM` isn't managed by the current class but exists on the outside. So, there could be unsynchronized uses of `DEPLOYTMENT_OUTPUT_STREAM` that interfere with the seemingly synchronized use here. I don't think there is much that can be done about this. However, to not give the impression that the uses of `DEPLOYMENT_OUTPUT_STREAM` here are thread-safe, consider to synchronize on `logForwarder` instead.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,641 @@
...
+            portPool.add(port); // return port to the pool, so the developer can start the delegate server and try again
+            return null;
+        }
+    }
+
+    /**
+     * Special GSON configuration that (un)wraps IValues as-is.
+     *
+     * Encoding and decoding an {@link IValue} loses dynamic type information, hence a decoded value can not be encoded properly again.
+     * `encode(decode(encode(v))) != encode(v)`
```
sungshik:
```suggestion
     * `decode(encode(v)) != v`
```
?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,641 @@
...
+     * Since the router should just proxy values passed from remote servers, without changing them, it uses a special encoder/decoder.
```
sungshik:
```suggestion
     * Since the router should just proxy values passed from remote servers, without changing them, it uses a special encoder/decoder, implemented in `ProxiedIValue`
```
(I think?)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,641 @@
...
+        var serverParams = BaseLanguageServer.DEPLOY_MODE
+            ? startServerProcess(lang, memoryArg)
+            : connectToServer(lang)
+            ;
+
+        if (serverParams == null) {
+            return null;
+        }
+
+        var requestPool = NamedThreadPool.single(String.format("parametric-lsp-router-%s", lang.getName().toLowerCase()));
```
sungshik, line 418: Nit: "router" shouldn't be part of the name (as it isn't a router)?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,641 @@
...
+
+    private void addExitHook(Future<Void> runner, LanguageParameter lang, IBaseLanguageServerExtensions server, Runnable exitHook, ExecutorService pool) {
+        getExecutor().execute(() -> {
+            try {
+                runner.get();
+                logger.info("Language server for {} terminated", lang.getName());
+            } catch (CancellationException | ExecutionException e) {
+                logger.error("Language server for {} terminated with an exception", lang.getName(), e);
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
```
sungshik, line 457: If an `InterruptedException` is thrown and caught here, then it doesn't necessarily mean that something went wrong at the remote server I think, which might still be alive and running without issues. After executing the finally block below, though, the router server doesn't know anymore that the remote server exists, so it becomes a kind of reverse zombie. (Not sure this is a problem in practice.)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -0,0 +1,641 @@
...
+    }
+
+    @Override
+    public synchronized CompletableFuture<Void> sendUnregisterLanguage(LanguageParameter lang) {
+        logger.debug("rascal/sendUnregisterLanguage({})", lang.getName());
+
+        if (ParametricTextDocumentService.isLanguageCompletelyRemoved(lang)) {
+            // Do not remove the connection to the server.
+            // For the deployed scenario, this is handled by the process onExit hook.
+            // For the development scenario, we maintain the connection, since the remote server does not exit.
```
sungshik, line 594: In general, it isn't clear to me anymore where lifecycles of remote servers are managed. Consider to document a high-level overview somewhere.

sungshik (commented):
Massive effort, nice improvements! Most of my comments are minor. The two larger items concern (see comments for details):
 1. thread-safety of `MultipleClientProxy`;
 2. `System.exit` in `ParametricTextDocumentService` (and lifecycle management of remote servers in general).

Point 1 might be solvable with additional documentation (I couldn't understand the code well enough to convince myself that it is sufficiently thread-safe). Point 2 isn't a blocking issue for this PR I think (primarily a maintenance concern) and can also be addressed in a different PR (if needed at all).

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentStateManager.java
toinehartman: I think this is fine. `unmodifiableSet` serves to prevent changes to the set by the caller modifying the underlying map. I added some comments to clarify this.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/DocumentRouter.java
toinehartman: Inlined

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/model/ParametricFileFacts.java
toinehartman: No, since there is still `setClient`.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricLanguageServer.java
toinehartman: Documented in `ParametricTextDocumentService`

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
toinehartman: > If no language can be found for `loc`, then is this a sign something is wrong

Before this PR, yes. But now, each remote server will only know about a single language, but is still informed of changes to all files (also of other languages), so we can read unsaved cross-language editor contents. Therefore, notifications come in for files that we might not know the extension of, and do not have a parser for, but that we still want to update the contents of.

> Is it possible for `safeLanguage(loc).isPresent()` to be false, but `URIResolverRegistry.getInstance().exists(loc)` to be true?

Given the answer to the previous question; yes.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
toinehartman: Because that is statically possibly null (`@MonotonicNonNull`). In practice, we know that it should always be non-null here. We use this pattern in various places.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
toinehartman: Renamed to `doesUnregisterAllRegistrations`

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
toinehartman: As answered on another comment as well, we do this to be able to read cross-language editor contents. From his PR on, cross-language also implies cross-process.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
toinehartman: Renamed to `request` to better align with known JSON-RPC terminology.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ProxiedIValue.java
toinehartman: I get the confusion. Should be read as "[JSON element representing an IValue] as an IValue". I'll rephrase.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
toinehartman: Unfortunately, there isn't. There is not common supertype for registration options.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
toinehartman: This is just for development. The port number was already hard-coded before. I'll rename to make clear that these are just for development/testing purposes.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
toinehartman: An attempt to gain some performance. I think it does not hurt, since language servers should be completely independent.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
toinehartman: Should always be an object, so this exception should never really be thrown. Inlined this so we can catch (and properly handle) this exception elsewhere.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
toinehartman: `DEPLOYTMENT_OUTPUT_STREAM` is typically `System.err` (but has been factored out to improve maintainability), which I believe is always accessed in a `synchronized` way. I remember some back-and-forth with @DavyLandman about this. It we synchronize on `logForwarder` instead, someone could write in between our writes to `System.err` and mess up the JSON objects.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/concurrent/CompletableFutureUtils.java
toinehartman: Only if the future completes before the loop finishes iterating, I think.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
toinehartman: It's the routing server, routing requests (on this thread). From the perspective of the user, reading the logs, isn't "router" an applicable identifiable name?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/log/LogJsonConfiguration.java
toinehartman: I think that's a bit excessive, although I recognize that reusing the function from one class in the other is also a bit confusing.

rascal-lsp/src/main/rascal/library/testing/lang/pico/LanguageServer.rsc
toinehartman: To test cross language editor contents

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
toinehartman: > The registration parameter is used only to get the language parameter.

In this class, yes, but not in the override in `ActualRoutingLanguageServer`.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricLanguageServer.java
toinehartman: This would already happen before as well:
https://github.com/usethesource/rascal-language-servers/blob/e69bbdb32179b98b17ae2e574da3c3e9313e330f/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricLanguageServer.java#L38

`parseArgs` is only called from `main` functions, which are called from
1) launch configs
2) deployed extensions
3) routing server spawning remote servers

I think all indicate programmer error and therefore are not really a problem.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
toinehartman: Agreed. I think this is an issue with the existing exit hook as well? If so, let's address in a separate PR.

https://github.com/usethesource/rascal-language-servers/blob/e69bbdb32179b98b17ae2e574da3c3e9313e330f/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java#L165

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
toinehartman: Refactored to use proxy registrations instead (see #1200)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
DavyLandman: I agree, let's drop this, as a regular stream is more than fast enough, so until it's a perf problem let's not add more parallelism to the mix.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
DavyLandman: I think that `p.destroy` is already the softer version, and if that fails we force kill it?

this is primarily in case of an hard exit. I think in other cases vs code would close the connection, and we would forward that close to the downstream servers.

rascal-lsp/src/main/rascal/library/testing/lang/json2/LanguageServer.rsc
DavyLandman: I think for testing language server it's fine if it's not a clear example, but I agree we should add a comment on that this is a syntatic example.

rascal-lsp/src/main/rascal/library/testing/lang/pico/LanguageServer.rsc
DavyLandman: let's add a comment that explains the case

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
DavyLandman: this problem has also been solved in #1183 let's try to align that.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
DavyLandman: agreed, how we exit should be something we take a look at, but outside this PR.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
DavyLandman: PrintWriters already lock themselves, adding locks around them can create all kinds of messy deadlocks.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
DavyLandman: the shutdown hook would take care of this I think.

PR #1092, opened 2026-05-29
Update dependencies where possible

DavyLandman:
Note that lsp4j 1.0.0 is released, but it should first be bumped on rascal side.


rascal-vscode-extension/package.json
```diff
@@ -313,7 +313,7 @@
   "devDependencies": {
     "@tsconfig/node20": "20.x",
     "@tsconfig/strictest": "2.x",
-    "@types/chai": "4.x",
+    "@types/chai": "5.x",
```
toinehartman, line 316: Chai `6.x` but the types for `5.x`? Is that right?

toinehartman (commented):
Thanks for doing a bunch of updates! Does this include all the updates in the currently open dependabot PRs?

toinehartman (changes requested):

toinehartman (approved):

rascal-vscode-extension/package.json
DavyLandman: Yes, type annotattions for chai 6 aren't published yet. and it's quite close to each other.

DavyLandman:
> Thanks for doing a bunch of updates! Does this include all the updates in the currently open dependabot PRs?

yeah, they should also automatically close after this one is merged.

PR #1099, opened 2026-06-03
added action to upgrade annotations to keyword parameters

jurgenvinju:
Work done for this PR:
* [x] link a version of Rascal with the upgrade function in the library
* [x] test the use of treeDiff on such a large scale
* [x] test the interaction with the progress bar and monitor warnings

The big thing was the first real application of treeDiff. I re-played the patch on all libraries and that seems ok.

There is only one usability issue that remains: all files which are changes are opened by the VScode client in an unsaved state. This goes for files which were already open, _and_ for files which were closed before starting the quickfix.

@DavyLandman this can be problematic, especially if the files are not visible in the tabs list.

I did some research and the LSP's `workspaces/applyEdit` feature is not causing this side-effect. It could be that one of our implementations or the LSP4J has this unintended side-effect. If you have time, I'd like to look at this together.

### DIFF
--- a/rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/Actions.rsc
+++ b/rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/Actions.rsc
@@ -29,6 +29,7 @@ POSSIBILITY OF SUCH DAMAGE.
 module lang::rascal::lsp::Actions
 
 import lang::rascal::\syntax::Rascal;
+import lang::rascal::upgrade::UpgradeAnnotationsToKeywordParameters;
 import util::LanguageServer;
 import analysis::diff::edits::TextEdits;
 import ParseTree;
@@ -46,6 +47,7 @@ The commands must be evaluated by ((evaluateRascalCommand))
 data Command
     = visualImportGraphCommand(PathConfig pcfg)
     | sortImportsAndExtends(Header h)
+    | upgradeAnnotationsToKeywordFields(PathConfig pcfg)
     ;
 
 @synopsis{Detects (on-demand) source actions to register with specific places near the current cursor}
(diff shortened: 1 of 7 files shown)


rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/Actions.rsc
```diff
@@ -66,6 +68,13 @@ list[CodeAction] rascalCodeActions(Focus focus, PathConfig pcfg=pathConfig()) {
                ;
     }
 
+    if ([*_, (Declaration) `<Tags _> <Visibility _> anno <Type _> <Type _> @ <Name _>;`, *_] := focus
```
jurgenvinju: should add the deprecated uses of annotation related functions as well (`removeAllAnnotations(...)`)

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/Actions.rsc
```diff
@@ -154,3 +163,8 @@ value evaluateRascalCommand(sortImportsAndExtends(Header h)) {
     applyDocumentsEdits([changed(h.src.top, [replace(h.imports.src, newHeader)])]);
     return ("result":true);
 }
+
+value evaluateRascalCommand(upgradeAnnotationsToKeywordFields(PathConfig pcfg)) {
+    applyDocumentsEdits(editsPathConfig(pcfg));
```
jurgenvinju, line 199: It's best to do this lazily because it takes a few seconds to parse all files.

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/Actions.rsc
DavyLandman: wouldn't it be nicer (and faster) to make these patterns in a switch? like these and the others in this `rascalCodeActions` function?

as this code is executed a lot, I'm wondering about the performance impact.

DavyLandman (commented):
I think we need at least 1 test for this behavior.

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/Actions.rsc
jurgenvinju: yes; let's do that when cleaning up. I want to get it working and then we can optimize.

DavyLandman: sorry, I just noticed my review wasn't request yet. I'll come back later.

jurgenvinju (commented):
needs automated tests to avoid unnoticed regressions. Otherwise I'm done.

PR #1101, opened 2026-06-04
Add support for VFS capabilities and fix REPL hyperlinks not opening anymore

DavyLandman:
This PR adds support for capabilities (companiont to usethesource/rascal#2792), and thus also marks our server as not supporting logical URIs, which improves the performance of rascal.

During the testing I also noticed the hyperlink detector didn't work anymore after the #967 changes.

### DIFF
--- a/rascal-vscode-extension/src/fs/JsonRpcMessages.ts
+++ b/rascal-vscode-extension/src/fs/JsonRpcMessages.ts
@@ -119,7 +119,7 @@ export interface TimestampResponse {
 }
 
 export interface SourceLocationResponse {
-    /** can be a uri or an ISourceLocation string with offsets */
+    /** can be a plain uri or an ISourceLocation string with offsets, as in: `|uri|(coordinates)` */
     loc: ISourceLocation | undefined
 }
 
--- a/rascal-vscode-extension/src/fs/JsonRpcMessages.ts
+++ b/rascal-vscode-extension/src/fs/JsonRpcMessages.ts
@@ -130,3 +130,22 @@ export interface DirectoryEntry {
     name: string;
     types: vscode.FileType[]
 }
+
+export interface Capabilities {
+    input: Capability | undefined;
+    output: Capability | undefined;
+    watch: Capability | undefined;
+    logical: Capability | undefined;
+    getCharset: Capability | undefined;
+}
+
+export interface Capability {
+    level: CapabilityLevel;
+    onlyForSchemes: string[] | undefined;
+}
+
+export enum CapabilityLevel {
+    unsupported = 0,
+    partial = 1,
+    full = 2
+}
--- a/rascal-vscode-extension/src/fs/JsonRpcMessages.ts
+++ b/rascal-vscode-extension/src/fs/JsonRpcMessages.ts
@@ -119,8 +119,10 @@ export interface TimestampResponse {
 }
 
 export interface SourceLocationResponse {
+    /** can be a uri or an ISourceLocation string with offsets */
     loc: ISourceLocation | undefined
 }
+
 
 export interface DirectoryListingResponse {
     entries: DirectoryEntry[]
--- a/rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
+++ b/rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
@@ -62,20 +62,17 @@ export class RascalFileSystemInVSCode implements vscode.FileSystemProvider {
             RascalFileSystemInVSCode.notifyWatchers(vscode.Uri.parse(event.root), this.translateFileChangeType(event.type.valueOf()), this.activeWatches, this._emitter, "RascalFileSystemInVSCode", this.logger);
         });
         this.serverCapabilities = client.sendRequest<Capabilities>("rascal/vfs/serverCapabilities");
-        this.inputClient = buildClientProxy(client, "rascal/vfs/input/", logger, this.checkWithCapabilities(c => c.input));
-        this.outputClient = buildClientProxy(client, "rascal/vfs/output/", logger, this.checkWithCapabilities(c => c.output));
-        this.watchClient = buildClientProxy(client, "rascal/vfs/watcher/", logger, this.checkWithCapabilities(c => c.watch));
-        this.logicalClient = buildClientProxy(client, "rascal/vfs/logical/", logger, this.checkWithCapabilities(c => c.logical));
+        this.inputClient = buildClientProxy(client, "rascal/vfs/input/", logger, this.capabilityChecker(c => c.input));
+        this.outputClient = buildClientProxy(client, "rascal/vfs/output/", logger, this.capabilityChecker(c => c.output));
+        this.watchClient = buildClientProxy(client, "rascal/vfs/watcher/", logger, this.capabilityChecker(c => c.watch));
+        this.logicalClient = buildClientProxy(client, "rascal/vfs/logical/", logger, this.capabilityChecker(c => c.logical));
         // note that charset is not added as VS Code does not support that over the VFS
     }
 
-    checkWithCapabilities(cap: (all: Capabilities) => Capability | undefined) {
-        return async (req: JsonRpcRequest) : Promise<boolean> => {
-            const currentCap = cap(await this.serverCapabilities);
-            if (!currentCap) {
-                return false;
-            }
-            switch (currentCap.level) {
+    async capabilityChecker(capField: (caps: Capabilities) => Capability | undefined) {
+        const cap = capField(await this.serverCapabilities) ?? <Capability>{ level: CapabilityLevel.unsupported };
+        return (req: JsonRpcRequest) : boolean => {
+            switch (cap.level) {
                 case CapabilityLevel.full: return true;
                 case CapabilityLevel.unsupported: return false;
                 case CapabilityLevel.partial: {
@@ -87,7 +84,7 @@ export class RascalFileSystemInVSCode implements vscode.FileSystemProvider {
                         uri = req.from;
                     }
                     // if there was no uri, let's assume it supported, and the server should throw an error if not the case
-                    return uri === undefined || supportedScheme(currentCap.onlyForSchemes!, uri);
+                    return uri === undefined || supportedScheme(cap.onlyForSchemes!, uri);
                 }
             }
         };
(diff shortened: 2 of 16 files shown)


rascal-vscode-extension/src/fs/JsonRpcMessages.ts
```diff
@@ -119,9 +119,11 @@ export interface TimestampResponse {
 }
 
 export interface SourceLocationResponse {
+    /** can be a uri or an ISourceLocation string with offsets */
```
rodinaarssen: A `URI` is also a TS `string`, I'm not sure which distinction is made here

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -59,9 +61,30 @@ export class RascalFileSystemInVSCode implements vscode.FileSystemProvider {
...
         });
-        this.inputClient = buildClientProxy(client, "rascal/vfs/input/", logger);
-        this.outputClient = buildClientProxy(client, "rascal/vfs/output/", logger);
-        this.watchClient = buildClientProxy(client, "rascal/vfs/watcher/", logger);
+        this.serverCapabilities = client.sendRequest<Capabilities>("rascal/vfs/serverCapabilities");
+        this.inputClient = buildClientProxy(client, "rascal/vfs/input/", logger, this.checkWithCapabilities(c => c.input));
+        this.outputClient = buildClientProxy(client, "rascal/vfs/output/", logger, this.checkWithCapabilities(c => c.output));
+        this.watchClient = buildClientProxy(client, "rascal/vfs/watcher/", logger, this.checkWithCapabilities(c => c.output));
+        this.logicalClient = buildClientProxy(client, "rascal/vfs/logical/", logger, this.checkWithCapabilities(c => c.logical));
+    }
```
rodinaarssen, line 70: Maybe add a comment that `getCharset` is not listed here as VS Code does not have this concept

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -59,9 +61,30 @@ export class RascalFileSystemInVSCode implements vscode.FileSystemProvider {
             }
             RascalFileSystemInVSCode.notifyWatchers(vscode.Uri.parse(event.root), this.translateFileChangeType(event.type.valueOf()), this.activeWatches, this._emitter, "RascalFileSystemInVSCode", this.logger);
         });
-        this.inputClient = buildClientProxy(client, "rascal/vfs/input/", logger);
-        this.outputClient = buildClientProxy(client, "rascal/vfs/output/", logger);
-        this.watchClient = buildClientProxy(client, "rascal/vfs/watcher/", logger);
+        this.serverCapabilities = client.sendRequest<Capabilities>("rascal/vfs/serverCapabilities");
+        this.inputClient = buildClientProxy(client, "rascal/vfs/input/", logger, this.checkWithCapabilities(c => c.input));
+        this.outputClient = buildClientProxy(client, "rascal/vfs/output/", logger, this.checkWithCapabilities(c => c.output));
+        this.watchClient = buildClientProxy(client, "rascal/vfs/watcher/", logger, this.checkWithCapabilities(c => c.output));
```
rodinaarssen: This should be `c => c.watch`

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -59,9 +61,30 @@ export class RascalFileSystemInVSCode implements vscode.FileSystemProvider {
...
+        return async (req: JsonRpcRequest) : Promise<boolean> => {
+            const currentCap = cap(await this.serverCapabilities);
+            if (!currentCap) {
+                return false;
+            }
+            switch (currentCap.level) {
+                case CapabilityLevel.full: return true;
+                case CapabilityLevel.unsupported: return false;
+                case CapabilityLevel.partial: {
+                    if ('loc'  in req && typeof req.loc === 'string') {
```
rodinaarssen: nit: two spaces after `loc`

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -307,3 +378,64 @@ function uriToRequest(uri: vscode.Uri): ISourceLocationRequest {
...
+    offsetLength?: [offset: number, length: number];
+    beginLineColumn?: [line: number, column: number];
+    endLineColumn?: [line: number, column: number];
+}
+
+function asNumberPair(ar: number[]): [number, number] {
+    if (ar.length !==  2) {
+        throw new Error(`Cannot convert ${ar} to tuple pair`);
+    }
+    return [ar[0]!, ar[1]!];
```
rodinaarssen, line 402: Interesting that the TS compiler cannot deduce this

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
rodinaarssen: Do we have an example of partially supported functionality in combination with a non-loc request?

Also: functions such as `rename` do not have `loc`

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -130,6 +153,23 @@ export class RascalFileSystemInVSCode implements vscode.FileSystemProvider {
         return uriString;
     }
 
+    static addCoordinates(uri: string, coordinates: IRascalCoordinates | undefined): string {
+        if (!coordinates || (!coordinates.offsetLength && (!coordinates.beginLineColumn || !coordinates.endLineColumn))) {
```
rodinaarssen: I think this misses the case where a loc only has offset-length (without line-column)

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -130,6 +153,23 @@ export class RascalFileSystemInVSCode implements vscode.FileSystemProvider {
...
+            return uri;
+        }
+        let result = `|${uri}|(`;
+        if (coordinates.offsetLength) {
```
rodinaarssen: Again, any loc with offset-length has/should have this, right? That also makes the remainder of this function simpler

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
```diff
@@ -307,3 +378,64 @@ function uriToRequest(uri: vscode.Uri): ISourceLocationRequest {
...
+    }
+    return onlyForSchemes.includes(scheme);
+}
+
+
+/**
+ * Rascal coordinates in unicode codepoints (so utf32/24)
+ */
+export interface IRascalCoordinates {
+    offsetLength?: [offset: number, length: number];
```
rodinaarssen: I think this is not optional

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -46,23 +46,28 @@ interface VSCodeResolverServer extends ISourceLocationInput, ISourceLocationOutp
 function connectHandlers(connection: rpc.MessageConnection, handler: VSCodeResolverServer, disposables: Disposable[], logger: vscode.LogOutputChannel) {
     disposables.push(connection.onRequest((request, arg, _token) => {
         const [rascal, vfs, direction, method] = request.split("/");
-        if (rascal !== "rascal" || vfs !== "vfs" || !direction || !method) {
+        if (rascal !== "rascal" || vfs !== "vfs" || !direction || (!method && direction !== "getCharset" && direction !== "serverCapabilities")) {
```
rodinaarssen, line 49: good catch!

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -156,6 +175,16 @@ class ResolverClient implements VSCodeResolverServer, Disposable  {
         connectHandlers(connection, this, this.disposables, logger);
     }
 
+    async serverCapabilities(): Promise<Capabilities> {
+        return {
+            getCharset: fullySupported(),
```
rodinaarssen, line 180: Is there an API to do this in VS Code?

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
```diff
@@ -156,6 +175,16 @@ class ResolverClient implements VSCodeResolverServer, Disposable  {
...
+            input: fullySupported(),
+            output: fullySupported(),
```
rodinaarssen, line 182: Are we sure all VS Code schemes are supported? And if not, can we obtain this information somehow and make it partially supported?

rodinaarssen (changes requested):
Nice work! Few questions and remarks here and there.

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
DavyLandman: well spotted 👍

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
DavyLandman: maybe at some point, and then we just try. exceptions should take care on the other side.

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
DavyLandman: IF there is only a offsetLenght, the left side of the `&&` will be false right? so the whole expression will fail.

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
DavyLandman: yes, I forgot, only the line & column is optional, not the other way around.

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
DavyLandman: yup, agreed, I looked for a while in the docs and couldn't find it.

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
rodinaarssen: You're right, I misread this the other night. This is again a check for line-column-without-offset-length then, right?

DavyLandman: Yes, but you're right, I've managed to simplify the expression based on your further comments

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
DavyLandman: I think we implemented this right? we open the editor and ask about it?

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
DavyLandman: sadly we cannot. so we have to wait for an exception.

rodinaarssen: Ok!

rascal-vscode-extension/src/fs/VSCodeFileSystemInRascal.ts
rodinaarssen: Right, I forgot about that.

rascal-vscode-extension/src/fs/RascalFileSystemInVSCode.ts
rodinaarssen: Yes, this looks great now

rodinaarssen (approved):

PR #1102, opened 2026-06-05
feat/rascal formatter

jurgenvinju:
This work by @toinehartman and YT introduces the Rascal formatter to VScode.

* [x] links the Rascal formatter from the stdlib to the `formatDocument` and `formatSelection` capabilities in LSP for Rascal
* [x] maps all FormattingOptions, taken from: #677 
* [x] support for formatting only a selection via the `Focus` mechanism, also taken from #677
* [x] works with subTreeEditFormatter from `util::Formatters` correctly
* [x] adds two UI tests, one for whole file formatting and one for selection formatting


jurgenvinju (commented):
I'm happy with this. 

Note that some important changes have been inherited from #677 which will be finished later.

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/Formatter.rsc
```diff
@@ -0,0 +1,44 @@
...
+}
+module lang::rascal::lsp::Formatter
+
+import ParseTree;
+import analysis::diff::edits::TextEdits;
+import lang::rascal::\syntax::Rascal;
+extend lang::rascal::format::Rascal;
+import util::Formatters;
+
+// TODO:need to provide `opts` with the function generated by `treeEdits` rather
```
DavyLandman: shall we make an issue out of this TODO? And also include if there is some kind of dependency on another PR/issue before we can solve this.

or is it more of a "nice idea for performance" but only if we find it actually impacts performance? (and maybe in that case replace the `TODO` label)

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
```diff
@@ -291,4 +291,39 @@ describe('IDE', function () {
 
         await ide.checkNoDiagnosticsAnymore();
     });
+
+    it ("full file formatting works", async() => {
```
DavyLandman, line 295: nice tests 👍🏼

DavyLandman (approved):
Looks nice, I only have a question on a TODO, and the status of that.

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/Formatter.rsc
jurgenvinju: This is unfortunately a rather big change in the Rascal syntax and type system. It requires a full RAP.

jurgenvinju: So let's write that RAP and then change the Todo to refer to it but let's not wait with merging this PR. Do you agree?

DavyLandman: Maybe the idea for a RAP should not be left as a TODO here?

Let's write that RAP, and after it's done change this code.

(Maybe refer to this code as example in the RAP).

Such that this TODO doesn't keep floating here.

jurgenvinju: Deal

PR #1105, opened 2026-06-08
Reduce the regex parsing in the terminal link detector

DavyLandman:


rascal-vscode-extension/src/RascalTerminalLinkProvider.ts
```diff
@@ -45,22 +45,18 @@ export class RascalTerminalLinkProvider implements TerminalLinkProvider<Extended
...
 
-    linkDetector() {
-        // sadly java script regex store state, so we have to create a new one everytime
-        return new RegExp(
-            "\\|[^\\t-\\n\\r\\s\\|:]+://[^\\t-\\n\\r\\s\\|]*\\|" // |location|
-            + "(\\([^\\)]*\\))?" // (optional offset)
-            , "g");
-    }
+    private readonly linkDetector =
+        /\|[a-zA-Z][\w\-.+]*:\/\/[^\s|]*\|(\([^)]*\))?/g;
```
sungshik: The regex seems more restrictive than in the Rascal grammar. Relevant excerpts (ignoring interpolation):

```rascal
syntax LocationLiteral
	= \default: ProtocolPart protocolPart PathPart pathPart ;

syntax ProtocolPart
	= nonInterpolated: ProtocolChars protocolChars 
	| interpolated: PreProtocolChars pre Expression expression ProtocolTail tail ;

syntax PathPart
	= nonInterpolated: PathChars pathChars 
	| interpolated: PrePathChars pre Expression expression PathTail tail ;

lexical ProtocolChars
	= [|] URLChars "://" !>> [\t-\n \r \ \u00A0 \u1680 \u2000-\u200A \u202F \u205F \u3000];

lexical PathChars
	= URLChars [|] ;

lexical URLChars
	= ![\t-\n \r \  \< |]* ;
```

sungshik (approved):

rascal-vscode-extension/src/RascalTerminalLinkProvider.ts
DavyLandman: True, we're not parsing rascal here (which might try and be a bit more flexible such that a typechecker can correct these errors) but this is parsing the output of vallang. Vallang's scheme rules are following the RFC:

https://github.com/usethesource/vallang/blob/015005f0277d9e91c8b122c1bedf2ad0838672c2/src/main/java/io/usethesource/vallang/impl/primitive/SourceLocationURIValues.java#L94-L117

But the scheme-specific-part has no constrains from vallang, so I could indeed try and keep the rascal rules there.

PR #1106, opened 2026-06-08
Wait for diagnostics to clear in tests

toinehartman:
This PR attempts to fix flaky tests when diagnostics are not (yet) cleared when the `afterEach` hook runs.


DavyLandman:
This might be related to #1108

toinehartman:
Fixed by [f5d411c](https://github.com/usethesource/rascal-language-servers/commit/f5d411c9c3e3e2f86856f7d9ff044d94d839aea5) instead

Issue #1078, opened 2026-05-07
Cross-language unsaved editor state


toinehartman:
**Goals**

- Language `A` can read editor contents of unsaved file of language `B`.

**Steps**

1. Write multi-language tests (possibly on `main` first).
2. Forward all file open/change/close notifications to all languages.

toinehartman:
Closed via #1112.

Issue #1077, opened 2026-05-07
IDE Services tests & implementation


toinehartman:
**Goals**

- Tests covering all operations from IDE services (`browse`, `edit`, etc.).
- IDE operations remain working with routing server.

**Steps**

1. Write tests on `main`, where all should pass.
2. Merge tests into routing server implementation.
3. Implement IDE services in routing server.

toinehartman:
Closed via #1109 and #1083

PR #1109, opened 2026-06-09
[POM-leading for DLSs] Fix tests after merging in `main`

toinehartman:
Fix tests after priodic merge of `main` into POM-leading feature branch.

This incidentally does some preliminary work for https://github.com/usethesource/rascal-language-servers/issues/1077 & https://github.com/usethesource/rascal-language-servers/issues/1078.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
@@ -27,7 +27,6 @@
 package org.rascalmpl.vscode.lsp.parametric.routing;
 
 import java.util.Collection;
-import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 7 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
```diff
@@ -163,18 +169,26 @@ private LanguageClient availableClient() {
 
     @Override
     public void didOpen(DidOpenTextDocumentParams params) {
+        var timestamp = System.currentTimeMillis();
+        openFile(params.getTextDocument(), l -> (loc, contents) -> CompletableFutureUtils.completedFuture(IRascalValueFactory.getInstance().character(0), exec), timestamp, exec);
```
DavyLandman, line 172: why is this only now a thing?

DavyLandman (approved):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
toinehartman: Something changed in the way the fallback/open file resolver works/is called, which means we now need to implement `isManagingFile` in the `RoutingTextDocumentService`. We do so by inheriting from `TextDocumentStateManager` and calling the file management hooks (open/change/close).

DavyLandman: okay

PR #1112, opened 2026-06-10
[POM-leading for DSLs] Cross language filesystem

### LINKED ISSUE #1078: Cross-language unsaved editor state

toinehartman:
Closes #1078.


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/NoContributions.java
```diff
@@ -87,7 +87,7 @@ public String getName() {
 
     @Override
     public CompletableFuture<ITree> parsing(ISourceLocation loc, String input) {
-        return CompletableFuture.failedFuture(new NoContributionException("parsing"));
+        throw new NoContributionException("parsing");
```
toinehartman, line 90: I believe this was a bug. This exception is caught here:
https://github.com/usethesource/rascal-language-servers/blob/c490a3626fa939ab61c1af3d7c94277472ef134d/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentState.java#L236-L239

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentStateManager.java
```diff
@@ -93,6 +93,10 @@ public String getContents(ISourceLocation file) {
         if (ideState != null) {
             return ideState.getCurrentContent().get();
         }
+        if (!URIResolverRegistry.getInstance().exists(file)) {
+            logger.error("Trying to get the contents of a non-existent file: {}", file);
+            return "";
+        }
```
DavyLandman: why do we need this extra check? isn't `isFile` enough?

rascal-vscode-extension/src/test/vscode-suite/dsl-mix.test.ts
```diff
@@ -0,0 +1,133 @@
...
+    let bench: Workbench;
+    let ide : IDEOperations;
+
+    const languages = ["Pico", "JSON2"];
+
+    this.timeout(Delays.extremelySlow * 2);
+
+    printRascalOutputOnFailure('Language Parametric Rascal');
+
+    function isLanguageLoading(bench: Workbench, language: string): () => Promise<boolean> {
```
DavyLandman: can this not be shared with the `Pico` tests?

rascal-vscode-extension/src/test/vscode-suite/dsl-mix.test.ts
```diff
@@ -0,0 +1,133 @@
...
+        };
+    }
+
+    async function loadLanguages() {
+        const repl = new RascalREPL(bench, driver);
+        await repl.start();
+
+        for (const lang of languages) {
+            await repl.execute(`import testing::lang::${lang.toLowerCase()}::LanguageServer;`, false, Delays.extremelySlow);
+            const replExecuteMain = await repl.execute(`testing::lang::${lang.toLowerCase()}::LanguageServer::register();`, false); // we don't wait yet, because we might miss language loading window
```
DavyLandman: the comment does not agree with the code.

or perhaps it's the comment should be on a newline?

rascal-vscode-extension/src/test/vscode-suite/dsl-mix.test.ts
```diff
@@ -0,0 +1,133 @@
...
+    afterEach(async function () {
+        await ide.revertOpenChanges();
+
+        if (this.test?.title) {
+            await ide.screenshot(`DSL-mix-${this.test?.title}`);
+        }
+        await ide.cleanup();
+    });
+
+    it("reads unsaved editor contents across languages", async function() {
```
DavyLandman, line 104: great test 👍🏼

DavyLandman (approved):
Yes, this is mostly nice and clean how this should be 👍🏼

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentStateManager.java
toinehartman: The check is not necessary per se, but I found the log message to be confusing. When looking up a non-existent file, it will log en error that the URI refers to a directory.

rascal-vscode-extension/src/test/vscode-suite/dsl-mix.test.ts
toinehartman: Factored this out and reused it for Pico and Rascal. Benefit of this check is that is checks parallel/background progress in the notification area as well. Useful when the loading progress is not the front-most one.

rascal-vscode-extension/src/test/vscode-suite/dsl-mix.test.ts
toinehartman: Fixed and aligned with how Pico does this.

DavyLandman (approved):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/TextDocumentStateManager.java
toinehartman: I reverted to the single check and made the error message less specific.

PR #1113, opened 2026-06-10
More precise use-defs in rename code

sungshik:
This PR updates the rename code to make it compatible with companion PR usethesource/rascal#2801, which updates the computation of use-def relations.

### DIFF
--- a/rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/refactor/rename/Common.rsc
+++ b/rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/refactor/rename/Common.rsc
@@ -68,6 +68,17 @@ bool isContainedInScope(loc l, loc scope, TModel tm) {
 
 loc getModuleFile(TModel tm) = getModuleScopes(tm)[tm.modelName].top;
 
+@synopsis{
+    Gets the use-def relation of TModel `tm`. Unlike `tm.useDef`, this function
+    guarantees that the range of the relation is a subset of the domain of
+    `tm.definitions`. Thus, unlike pairs in `tm.useDef`, if `<u, d>` is a pair
+    in the relation, then `d` can safely be used to index `tm.definitions`.
+}
+rel[loc, loc] getUseDef(TModel tm) {
+    map[loc, loc] id2define = invertUnique(tm.define2id);
+    return {<u, id2define[d] ? d> | <loc u, loc d> <- tm.useDef};
+}
+
 private set[str] reservedNames = getRascalReservedIdentifiers();
 
 str forceUnescapeNames(str name) = replaceAll(name, "\\", "");
--- a/rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/refactor/rename/Parameters.rsc
+++ b/rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/refactor/rename/Parameters.rsc
@@ -54,9 +54,10 @@ TModel augmentFormalUses(Tree tr, TModel tm, TModel(loc) getModel) {
         | loc f <- getModuleFile(tm) + (tm.paths<to>)
         , fileTm := getModel(f.top)
     };
+    rel[loc, loc] useDef = getUseDef(tm);
     visit (tr) {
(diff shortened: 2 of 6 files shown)


rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/refactor/rename/Common.rsc
```diff
@@ -68,6 +68,17 @@ bool isContainedInScope(loc l, loc scope, TModel tm) {
 
 loc getModuleFile(TModel tm) = getModuleScopes(tm)[tm.modelName].top;
 
+@synopsis{
+    Gets the use-def relation of TModel `tm`. Unlike `tm.useDef`, this function
+    guarantees that the range of the relation is a subset of the domain of
+    `tm.definitions`. Thus, unlike pairs in `tm.useDef`, if `<u, d>` is a pair
+    in the relation, then `d` can safely be used to index `tm.definitions`.
+}
+rel[loc, loc] getUseDef(TModel tm) {
```
toinehartman, line 77: Did you check if this function appears in the profile (for example by running one of the rename test modules)? Some memoization might be useful here.

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/refactor/rename/Parameters.rsc
```diff
@@ -56,7 +56,7 @@ TModel augmentFormalUses(Tree tr, TModel tm, TModel(loc) getModel) {
     };
     visit (tr) {
         case (Expression) `<Expression e>(<{Expression ","}* _> <KeywordArguments[Expression] kwArgs>)`: {
-            funcKwDefs = keywordFormalDefs[tm.useDef[e.src]];
+            funcKwDefs = keywordFormalDefs[getUseDef(tm)[e.src]];
```
toinehartman: It seems expensive to do this inside the `visit`, but since we're modifying the use-defs inside this `visit` as well, it might be necessary to do it here...

toinehartman (approved):
Good stuff! I cannot completely understand how these changes will influence uses of the use-def relation that are *not* used to index `tm.definitions`.

I would expect the main benefit of this TModel extension to be that we can override (one of) the following functions from the rename framework. As you mentioned off-line, this is something we can address in the future.

I think with the addition of `tm.define2id`, we should be able to implement the following without any tree visits.

https://github.com/usethesource/typepal/blob/bde05af0748c6200f72403ad4f6ec420ccb051f8/src/analysis/typepal/refactor/Rename.rsc#L430-L437

Or even the (only) function that calls the above:

https://github.com/usethesource/typepal/blob/bde05af0748c6200f72403ad4f6ec420ccb051f8/src/analysis/typepal/refactor/Rename.rsc#L348-L372

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/refactor/rename/Parameters.rsc
sungshik: Good point! The use-defs are modified during the visit, but as far as I understand the code: (a) entries are only created, not deleted, not updated; (b) none of the created entries have `e.src` as use. So, I think it is safe to call `getUseDef` only once, before the visit, because the subset of it that is read during the visit doesn't change.

I made the change, ran the tests, and everything seems fine. What do you think?

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/refactor/rename/Common.rsc
sungshik: I ran the profiler on `Constructors` and `getUseDef` doesn't appear in the list. For now, it looks alright without memo.

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/refactor/rename/Parameters.rsc
toinehartman: This piece of code is basically a local fix for https://github.com/usethesource/rascal/issues/2186 so your explanation makes sense.

PR #1114, opened 2026-06-12
Using new LSP `textDocumentContent` feature for text document content that is not on disk

DavyLandman:
This new `textDocumentContent` LSP 3.18 feature allows us to remove the `RascalInVSCodeFilesystem` feature without losing the capability to view contents of non-`file:///` locs. This PR switches to that API and cleans up the previous code that would simulate a full file system.


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
```diff
@@ -76,11 +97,98 @@ public void initialize(ClientCapabilities clientCap, @Nullable List<WorkspaceFol
...
 
     @Override
     protected void projectRemoved(ISourceLocation loc) {
         ((RascalTextDocumentService) availableDocumentService()).projectRemoved(loc);
     }
 
+    private void registerAvailableSchemes() {
+        var client = availableClient();
+        for (var scheme : calculatePossibleSchemes()) {
+            var schemeCopy = scheme;
```
rodinaarssen: What's the copy for?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
```diff
@@ -76,11 +97,98 @@ public void initialize(ClientCapabilities clientCap, @Nullable List<WorkspaceFol
...
+                if (message == null || message.isEmpty()) {
+                    message = e.getClass().getName();
+                }
+                throw new ResponseErrorException(new ResponseError(-100, "Could not read " + params.getUri() + "due to: " + message, e));
+            }
+        }, getExecutor());
+    }
+
+    private static ISourceLocation parseVSCodeURI(String uri) {
+        var plainUri = URIUtil.assumeCorrect(uri);
```
rodinaarssen: Are we sufficiently confident that VS Code will only produce valid URIs?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
```diff
@@ -76,11 +97,98 @@ public void initialize(ClientCapabilities clientCap, @Nullable List<WorkspaceFol
...
+        if ((plainUri.getAuthority() == null || plainUri.getAuthority().isEmpty()) && !plainUri.getScheme().equals("file")) {
+            // VS Code omits the leading two slashes from URIs if the autority is empty *and* the scheme is not equal to "file"
+            // Rascal does not support this style of URIs, so we add the slashes before sending the URI over
+            uri = uri.replaceFirst(":/", ":///");
+        }
+        return URIUtil.assumeCorrectLocation(uri);
```
rodinaarssen: `URIUtil.assumeCorrectLocation` will already produce a valid `loc` when presented with a single-slash URI string; there's no need to "fix" the URI first

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
```diff
@@ -76,11 +97,98 @@ public void initialize(ClientCapabilities clientCap, @Nullable List<WorkspaceFol
...
+            allReadableSchemes().filter(Predicate.not(GENERIC_SCHEMES::contains)),
+            // nested schemes like jar+file etc
+            allReadableSchemes().filter(Predicate.not(CONTAINER_SCHEMES::contains))
+                .flatMap(s -> CONTAINER_SCHEMES.stream().map(c -> c + "+" + s))
+        ).collect(Collectors.toList());
+    }
+
+
+    @Override
+    public CompletableFuture<TextDocumentContentResult> textDocumentContent(TextDocumentContentParams params) {
```
rodinaarssen, line 171: `RascalFileSystemInVSCode` used to have a transformation step from VS Code-provided URIs to Rascal-ready locations (cf. `RascalFileSystemInVSCode::toRascalLocation`), is that no longer necessary?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
```diff
@@ -76,11 +97,98 @@ public void initialize(ClientCapabilities clientCap, @Nullable List<WorkspaceFol
...
+            // VS Code omits the leading two slashes from URIs if the autority is empty *and* the scheme is not equal to "file"
+            // Rascal does not support this style of URIs, so we add the slashes before sending the URI over
+            uri = uri.replaceFirst(":/", ":///");
+        }
+        return URIUtil.assumeCorrectLocation(uri);
+
+    }
+
+
 }
```
rodinaarssen, line 190: I guess that, next to writable, we also lose watches. Is that a problem?

Likewise for `stat` (and derivatives)?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -378,11 +379,44 @@ public void setMinimumLogLevel(String level) {
...
-            return CompletableFutureUtils.completedFuture(Sets.union(inputs, logicals).toArray(String[]::new), executor);
+        @Override
+        public CompletableFuture<SourceLocationResponse> resolve(ISourceLocationRequest req) {
+            logger.trace("resolve: {}", req.getLocation());
+            return CompletableFuture.supplyAsync(() -> {
+                var loc = toRascalLocation(req.getLocation());
+                ISourceLocation resolved = null;
+                if (!loc.getScheme().equals("std")) {
+                    // TODO: this works around the fact that `std` is a bit of a broken scheme in
+                    // VS Code, as REPL 1 migh have a different std than REPL2, and again different from the
```
rodinaarssen:
```suggestion
                    // VS Code, as REPL 1 might have a different std than REPL 2, and again different from the
```

rascal-vscode-extension/src/lsp/RascalLanguageServer.ts
```diff
@@ -79,5 +76,89 @@ export class RascalLanguageServer implements vscode.Disposable {
         void this.rascalClient.then(c => c.dispose());
         this.languageRegistry.dispose();
     }
+
+    async resolve(uri: vscode.Uri, coordinates?: IRascalCoordinates) : Promise<[vscode.Uri, IRascalCoordinates | undefined]> {
+        this.logger.debug("[RascalLanguageServer] resolve: ", [uri, coordinates]);
+        const result = await (await this.rascalClient).sendRequest<SourceLocationResponse>("rascal/vfs/logical/resolve", {
```
rodinaarssen: Remote IO errors are no longer handled properly

rascal-vscode-extension/src/test/vscode-suite/remotefs.test.ts
```diff
@@ -101,55 +101,13 @@ describe('RemoteFS', function () {
...
         await repl.start();
         await repl.execute("import IO;");
-        await repl.execute("l = |tmp:///rascal-remotefs-test/rascal-test-file|;");
-        await repl.execute('writeFile(l, "")');
-        await repl.execute("readFile(|rascal-vscode-test:///remotefs-api-test/test-rascalfs-write|)");
-        await repl.execute('readFile(l) == "hi"');
+        await repl.execute("l = |compressed+tmp:///rascal-remotefs-test/rascal-test-file.gz|;");
+        await repl.execute('writeFile(l, "hi")');
+        await repl.execute('readFile(|rascal-vscode-test:///remotefs-api-test/test-rascalfs-read|) == "hi"');
         expect(repl.lastOutput).is.equal("bool: true", "Writing Rascal Code fs works");
```
rodinaarssen:
```suggestion
        expect(repl.lastOutput).is.equal("bool: true", "Reading from Rascal fs works");
```

rascal-vscode-extension/src/RascalTerminalLinkProvider.ts
```diff
@@ -68,7 +68,7 @@ export class RascalTerminalLinkProvider implements TerminalLinkProvider<Extended
             return vscode.commands.executeCommand("vscode.open", sloc.uri) ;
         }
 
-        const [uri, coordinates] = await (await this.client).resolve(vscode.Uri.parse(sloc.uri), sloc.coordinates);
+        const [uri, coordinates] = await  this.client.resolve(vscode.Uri.parse(sloc.uri), sloc.coordinates);
```
rodinaarssen:
```suggestion
        const [uri, coordinates] = await this.client.resolve(vscode.Uri.parse(sloc.uri), sloc.coordinates);
```

rodinaarssen (approved):
That's a nice reduction.

Some small remarks

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
DavyLandman: Because I'm always a bit confused in which language I'm working in again, and unsure how the capturing works for closure. so yes, in java this is wrong :)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
DavyLandman: yes, they follow the same standard. luckily :')

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
DavyLandman: hmm, indeed, I ported it from typescript, but that was hitting the JSON serializer directly, not the `URIUtil` path.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
DavyLandman: we know we don't support opaque locations, as we won't be registering that URI? but the `Locations.toClientLocation` might be a good idea to call indeed.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalWorkspaceService.java
DavyLandman: yes, that doesn't really exist (although there is a way to report updates for contents, we'll not be supporting that for the readonly URIs)

rascal-vscode-extension/src/lsp/RascalLanguageServer.ts
DavyLandman: thanks, fixed that now.

PR #1116, opened 2026-06-15
Start REPL from command palette

### LINKED ISSUE #1087: Command "Start Rascal Terminal and Import this module" doesn't do that
**Describe the bug**

Command "Start Rascal Terminal and Import this module" doesn't do that. Nothing seems to happen (also no errors/warnings in the log). Same for "Command "Start Rascal Terminal, Import module and Run main function".

Image

**To Reproduce**

Run the command when the active editor contains a Rascal file.

sungshik:
Fixes #1087

### DIFF
--- a/rascal-vscode-extension/package.json
+++ b/rascal-vscode-extension/package.json
@@ -45,15 +45,15 @@
     "commands": [
       {
         "command": "rascalmpl.createTerminal",
-        "title": "Create Rascal Terminal"
+        "title": "Create New Rascal Terminal"
       },
       {
         "command": "rascalmpl.runMain",
-        "title": "Create Rascal Terminal, Import Module, and Run Main Function"
+        "title": "Create New Rascal Terminal, Import Module, and Run Main Function"
       },
       {
         "command": "rascalmpl.importModule",
-        "title": "Create Rascal Terminal and Import Module"
+        "title": "Create New Rascal Terminal and Import Module"
       },
       {
         "command": "rascalmpl.startDebuggerForRepl",
--- a/rascal-vscode-extension/package.json
+++ b/rascal-vscode-extension/package.json
@@ -49,11 +49,11 @@
       },
       {
         "command": "rascalmpl.runMain",
-        "title": "Start Rascal Terminal, Import module and Run main function"
+        "title": "Create Rascal Terminal, Import Module, and Run Main Function"
       },
       {
         "command": "rascalmpl.importModule",
-        "title": "Start Rascal Terminal and Import this module"
+        "title": "Create Rascal Terminal and Import Module"
       },
       {
         "command": "rascalmpl.startDebuggerForRepl",
--- a/rascal-vscode-extension/src/RascalExtension.ts
+++ b/rascal-vscode-extension/src/RascalExtension.ts
@@ -92,6 +92,7 @@ export class RascalExtension implements vscode.Disposable {
     private registerMainRun() {
         this.context.subscriptions.push(
             vscode.commands.registerTextEditorCommand("rascalmpl.runMain", (text, _edit, moduleName) => {
+                moduleName = moduleName ?? this.guessModuleName(text.document);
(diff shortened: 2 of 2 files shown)


rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -121,6 +123,35 @@ export class RascalExtension implements vscode.Disposable {
...
+            //   -                `Qux`
+            //
+            // For each of these possible module names (from long to short), the
+            // approach below is to derive it from the location (also taking
+            // into account escape characters), and then try it.
+
+            let segments = document.uri.path.slice(0, -4).split('/');
+            while ((segments = segments.slice(1)).length > 0) {
+                const guess = segments.map(s => `[\\\\]?${s}`).join('::');
+                const guessed = text.match(`module[\\s]+(${guess})`);
```
DavyLandman, line 145: what if the module name has an escaped name in there? like `\syntax` or `\test`?

why don't we use a regex to match the module name in there?

```ts
/^\smodule\s*(.*)$/
```

rascal-vscode-extension/package.json
```diff
@@ -49,11 +49,11 @@
       },
       {
         "command": "rascalmpl.runMain",
-        "title": "Start Rascal Terminal, Import module and Run main function"
+        "title": "Create Rascal Terminal, Import Module, and Run Main Function"
```
DavyLandman: why the `Start` => `Create`? Looking at VS Code they call it: `New terminal`

Image

Not sure if I like this better though.

DavyLandman (commented):
I'm not sure I like the regex loop in the module name.

rascal-vscode-extension/src/RascalExtension.ts
sungshik: > what if the module name has an escaped name in there? like `\syntax` or `\test`?

This is accounted for in part `[\\\\]?${s}` of `guess` (i.e., it puts an optional backslash in front of each segment of the qualified name.

> why don't we use a regex to match the module name in there?
> ```
> /^\smodule\s*(.*)$/
> ```

We could, but I tried to make it a bit more resilient to false positives. For instance, to also make it work with (wrongly matches `/^\smodule\s*(.*)$/`):
```rascal
/* DEPRECATED STUFF COPIED FROM THE OLD MODULE
module foo::Bar
import IO;

void main() {
  println("Hello, world")
}
*/

module foo::Baz
import IO;

void main() {
  println("asdf");
}
```
Or (wrongly matches `/module\s*(.*)/`):
```rascal
/*
 * This module replaces module foo::Bar. Don't use that old module. It is deprecated.
 */
module foo::Baz
import IO;

void main() {
  println("asdf");
}
```

rascal-vscode-extension/package.json
sungshik: The intention of this change was to make it uniform with the existing command (that doesn't import/run):
https://github.com/usethesource/rascal-language-servers/blob/241d2ded40bf10dbf4b0fcc15c88e6792ae6024a/rascal-vscode-extension/package.json#L46-L49

I am fine either way, but in the command palette, the VS Code command for new terminals also uses "Create":

image

So actually, maybe it is best to also add "New" to our commands:
  - "Create New Rascal Terminal"
  - "Create New Rascal Terminal and Import Module"
  - "Create New Rascal Terminal, Import Module, and Run Main Function"

DavyLandman: Yes, let's align that :+1:

rascal-vscode-extension/src/RascalExtension.ts
DavyLandman: AH okay, fair. I could construct a simalar thing for the other approach:

```
// it shouldn't be module src::X we we have a source folder
module X
```

for example.

parsing rascal with regex is always a thing..

but agreed, a bit harder to trigger than the other one.

toinehartman (approved):

PR #1117, opened 2026-06-15
[POM-leading for DSLs] Dynamic capabilities

### LINKED ISSUE #1079: Dynamic capability registration
**Goals**

- Aggregate & forward dynamic capabilities from remote servers in router and register with client.

**Steps**

- Maintain local administration of current state of capabilities.
- Push updates to client.

toinehartman:
Implement dynamic capabilities for POM-leading servers.

Closes #1079.


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -373,11 +379,15 @@ private void triggerAnalyzer(TextDocumentState state, Duration delay) {
...
-        var fileFacts = facts(location);
-        fileFacts.invalidateBuilder(location);
-        fileFacts.calculateBuilder(location, getFile(location).getCurrentTreeAsync(true));
+        if (safeLanguage(location).isPresent()) {
+            logger.trace("Triggering builder for {}", doc.getUri());
+            var fileFacts = facts(location);
+            fileFacts.invalidateBuilder(location);
+            fileFacts.calculateBuilder(location, getFile(location).getCurrentTreeAsync(true));
+        } else {
+            logger.debug("Not triggering builder, since no language is registered for {}", location);
```
toinehartman, line 388: This aligns the builder with the analyzer

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/AbstractDynamicCapability.java
```diff
@@ -150,7 +150,8 @@ public boolean equals(@Nullable Object obj) {
         }
         var other = (AbstractDynamicCapability<?>) obj;
         return Objects.equals(id, other.id)
-            && Objects.equals(methodName, other.methodName);
```
DavyLandman, line 153: looks like a bug in main?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/concurrent/CompletableFutureUtils.java
```diff
@@ -41,6 +41,8 @@
 public class CompletableFutureUtils {
     private CompletableFutureUtils() {/* hidden */ }
 
+    public static final CompletableFuture<Void> NOOP = CompletableFuture.completedFuture(null);
```
DavyLandman, line 44: while this is nice, it has no exec bound to it? isn't that a problem?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -360,11 +366,11 @@ public void write(JsonWriter writer, ProxiedIValue proxiedValue) throws IOExcept
 
         var serverLauncher = new Launcher.Builder<IBaseLanguageServerExtensions>()
             .setRemoteInterface(IBaseLanguageServerExtensions.class)
-            .setLocalService(client)
+            .setLocalService(remoteClient)
             .setInput(serverParams.getLeft())
             .setOutput(serverParams.getMiddle())
             .configureGson(ActualRoutingLanguageServer::configureProxyGson)
-            .setExecutorService(getExecutor())
+            .setExecutorService(NamedThreadPool.single(String.format("parametric-lsp-router-%s", lang.getName().toLowerCase())))
```
DavyLandman: why are we not using our own pool? 

that we're not tracking nor shutting down?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -529,7 +529,7 @@ public void exit() {
     @Override
     public void cancelProgress(WorkDoneProgressCancelParams params) {
         // Forward to everyone
-        allRoutes().forEach(r -> r.thenAccept(s -> s.cancelProgress(params)));
+        allRoutes(r -> r.thenAccept(s -> s.cancelProgress(params)));
```
DavyLandman, line 535: 👍🏼 that is better indeed

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -41,175 +51,247 @@
...
+            .reduce(params
+                .getRegistrations()
+                .parallelStream()
+                .map(this::registerCapability), exec)
+            .thenAccept(v -> {}); // convert to Void
+    }
+
+    private CompletableFuture<Void> registerCapability(Registration r) {
+        var c = currentRegistrations.computeIfAbsent(r.getMethod(), k -> new CopyOnWriteArraySet<>());
+        // Lock on the registrations for this method for a moment
```
DavyLandman: can we explain why we lock? who else might be doing this registration?

and why we lock for so long? 

My concern is that the `exceptionally` handler that manipulates `c` is executed outside the lock (and not relocking).

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -41,175 +51,247 @@
...
+                logger.trace("No registrations for {} exist; ignoring this request: {}", u.getMethod(), u);
+                // No registrations => nothing to unregister
+                return NOOP;
+            }
+
+            c.removeAll(cs);
+            return client.unregisterCapability(new UnregistrationParams(List.of(u)))
+                .exceptionally(t -> {
+                    logger.error("Exception while unregistering {} ({})", u.getMethod(), u, t);
+                    c.addAll(cs);
```
DavyLandman: same here, no lock here, is that fine? and if so, let's explain why.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
```diff
@@ -183,7 +192,13 @@ public void didOpen(DidOpenTextDocumentParams params) {
 
         // Inform all remote servers about this file, so they can maintain its state.
         // Note: floating futures
```
DavyLandman: I think this comment is outdated (and other in this class) as we do in fact not leave the future floating.


what if the `didOpen` was not processed on all the down-stream servers, is that always an error? if not, can we update the `logger.catching` to add some kind of message to explain this.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
```diff
@@ -193,7 +208,13 @@ public void didChange(DidChangeTextDocumentParams params) {
...
-        allRoutes().forEach(r -> r.thenAccept(s -> s.didChange(params)));
+        try {
+            notifyAllRoutes(s -> s.didChange(params)).get(1, TimeUnit.SECONDS);
```
DavyLandman: can we introduce a function that does this get + catch once instead of repeating this for all the 3 functions in this class?

rascal-lsp/src/main/rascal/library/testing/lang/pico/LanguageServer.rsc
```diff
@@ -137,7 +137,7 @@ set[LanguageService] testingLanguageServerWithRecovery() = testingLanguageServer
 set[LanguageService] testingLanguageServerSlowSummary() = testingLanguageServerSlowSummary(false);
 set[LanguageService] testingLanguageServerSlowSummaryWithRecovery() = testingLanguageServerSlowSummary(true);
 
-void register(bool errorRecovery=false) {
```
DavyLandman, line 140: why is the default switched?

rascal-vscode-extension/src/test/vscode-suite/dsl-mix.test.ts
```diff
@@ -37,6 +37,8 @@ describe('DSL [multi-language]', function () {
     let ide : IDEOperations;
 
     const languages = ["Pico", "JSON2"];
+    const jsonExampleDir = src(TestWorkspace.testProject, 'json2');
+    const jsonTestFile = path.join(jsonExampleDir, 'example.json2');
```
DavyLandman, line 42: maybe they should use the new protected files feature?

DavyLandman (approved):
Looks nice! Just some small points.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
toinehartman: I believe we need a single-threaded pool here to maintain the order of incoming/outgoing requests. We probably shut it down on when the server exits, indeed.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/AbstractDynamicCapability.java
toinehartman: Yes. Although there should never be multiple instances of this class with the same method in one parametric server, so I do not think it is a real issue currently. Otherwise we could fix on main first.

rascal-lsp/src/main/rascal/library/testing/lang/pico/LanguageServer.rsc
toinehartman: Because I wanted error recovery enabled in the mixed DSL tests, and the JSON server has it enabled by default. I figured it doesn't hurt - the DSL test explicitly set the value anyway, and for user testing error recovery is useful as well.

The alternative is to set it explicitly in the mixed DSL tests, and add this flag to the JSON testing server (or accept a warning for an unknown keyword argument).

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -41,175 +50,287 @@
...
+                                logger.trace("Re-registering {}, since other servers still support it: {}", method, reg);
+                                return client.registerCapability(new RegistrationParams(List.of(reg)));
+                            }
+                            return NOOP;
+                        })
+                        .thenApply(v -> currentRegs);
+                }
+
+                // We received an unregistration for a registration that we do not know
+                // TODO Throw an error?
```
toinehartman: We probably do not want to throw an error towards the remote server unless we really cannot recover?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/CapabilityRegistration.java
```diff
@@ -131,7 +132,7 @@ public CompletableFuture<Void> update(Collection<ICapabilityParams> languages) {
         */
         lastParams.set(List.copyOf(languages));
         return CompletableFutureUtils.reduce(dynamicCapabilities.stream().map(this::updateRegistration), exec)
-            .thenAccept(_v -> logger.debug("Done updating dynamic capabilities"));
+            .thenAccept(v -> logger.debug("Done updating dynamic capabilities"));
     }
```
DavyLandman, line 135: nit: shouldn't this still be `_v` as we're not using it?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -358,13 +364,14 @@ public void write(JsonWriter writer, ProxiedIValue proxiedValue) throws IOExcept
             return null;
         }
 
+        var requestPool = NamedThreadPool.single(String.format("parametric-lsp-router-%s", lang.getName().toLowerCase()));
```
DavyLandman, line 367: nice 👍🏼

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
DavyLandman: agreed, but let's add a log method that we know this thing happened.

DavyLandman (approved):
Looking good.

PR #809, opened 2025-10-01
Remember files opened as `Parametric Rascal LSP` before language registration

toinehartman:
This PR keeps track of files that are open with language `Parametric Rascal LSP`, for which there is no language registered yet. Once the user registers a language, we update the associated open editors.

This PR closes #795.

PR #1118, opened 2026-06-15
Document selectors for capabilities

toinehartman:
1. Set document selectors for capabilities, such that capabilities can only be triggered on documents that support them. Especially useful when registering multiple languages.
2. Dynamically register semantic highlighting capability (to test the former change, and make a start with registering all capabilities dynamically).
3. Instead of toggling the language for open editors, request a refresh of the open editors via LSP. This prevents races where requests can come in while the file is 'closed'.

This PR is a preparation to easily enable #1117.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
@@ -321,6 +321,7 @@ public class ParametricTextDocumentService extends TextDocumentStateManager impl
         var ext = URIUtil.getExtension(Locations.toLoc(uri));
         // We want the original scheme, not the one possibly modified when converting URI -> loc
         if (ext.equals("") && extensionLessSchemes.add(uri.getScheme())) {
+            // Should be called from the main, single-threaded request pool
             updateCapabilities();
         }
     }
@@ -956,6 +957,9 @@ public class ParametricTextDocumentService extends TextDocumentStateManager impl
 
         // Do this last, after the parser for open editors has been updated,
         // since this might trigger new requests on the editor contents.
+        // `updateCapabilities`/`CapabilityRegistration::update` should never be called asynchronously, since that might re-order incoming updates.
+        // Since `registerLanguage` is called from a single-threaded pool, calling it here is safe.
+        // Note: `updateCapabilities` returns a void future, which we do not have to wait on.
         updateCapabilities();
     }
 
@@ -973,9 +977,6 @@ public class ParametricTextDocumentService extends TextDocumentStateManager impl
     }
 
     private synchronized CompletableFuture<Void> updateCapabilities() {
-        // `CapabilityRegistration::update` should never be called asynchronously, since that might re-order incoming updates.
-        // Since `registerLanguage` is called from a single-threaded pool, calling it here is safe.
-        // Note: `CapabilityRegistration::update` returns a void future, which we do not have to wait on.
         return availableCapabilities()
             .update(buildLanguageParams())
             .thenCompose(v -> refreshEditors());
@@ -1051,6 +1052,7 @@ public class ParametricTextDocumentService extends TextDocumentStateManager impl
             contributions.remove(lang.getName());
         }
 
+        // Should be called from the main, single-threaded request pool
         updateCapabilities();
     }
 
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
@@ -138,6 +138,7 @@ import org.rascalmpl.vscode.lsp.parametric.capabilities.CapabilityRegistration;
 import org.rascalmpl.vscode.lsp.parametric.capabilities.CompletionCapability;
 import org.rascalmpl.vscode.lsp.parametric.capabilities.FileOperationCapability;
 import org.rascalmpl.vscode.lsp.parametric.capabilities.ICapabilityParams;
+import org.rascalmpl.vscode.lsp.parametric.capabilities.SemanticTokensCapability;
 import org.rascalmpl.vscode.lsp.parametric.model.ParametricFileFacts;
 import org.rascalmpl.vscode.lsp.parametric.model.ParametricSummary;
 import org.rascalmpl.vscode.lsp.parametric.model.ParametricSummary.SummaryLookup;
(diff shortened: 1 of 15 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/CapabilityRegistration.java
```diff
@@ -287,15 +290,42 @@ private <T> void handleError(@Nullable Throwable t, AbstractDynamicCapability<T>
...
+                        setDocumentSelector((TextDocumentRegistrationOptions) opts, providingLanguages);
+                    }
+                    return opts;
+                })
                 .thenApply(opts -> new Registration(cap.id(), cap.methodName(), opts));
         });
     }
 
+    private void setDocumentSelector(TextDocumentRegistrationOptions opts, Collection<ICapabilityParams> params) {
+        // If the document selector is set already by the capability implementation, we do not touch it.
```
DavyLandman: Can we explain why we do not touch it? as the condition in the if already says we do not touch it.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -180,6 +183,8 @@ public class ParametricTextDocumentService extends TextDocumentStateManager impl
 
     private final String dedicatedLanguageName;
     private final SemanticTokenizer tokenizer = new SemanticTokenizer();
+    private final Set<String> extensionLessSchemes = new HashSet<>();
```
DavyLandman: We will be writing to this from a single thread, but reading from different threads. That can cause messy problems.

To be safe we should use a concurrent hashset, (the trick with the keyset from the concurrent hashmap).

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -303,6 +308,21 @@ public void didOpen(DidOpenTextDocumentParams params) {
         TextDocumentState file = open(params.getTextDocument(), timestamp);
         handleParsingErrors(file, file.getCurrentDiagnosticsAsync());
         triggerAnalyzer(params.getTextDocument(), NORMAL_DEBOUNCE);
+
+        // Discover capabilities
+        discoverExtensionLessScheme(URIUtil.assumeCorrect(params.getTextDocument().getUri()));
```
DavyLandman, line 312: Let's reuse the `file.getLocation()` here.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -938,6 +953,32 @@ public synchronized void registerLanguage(LanguageParameter lang) {
                 refreshFileState(f);
             }
         }
+
+        // Do this last, after the parser for open editors has been updated,
+        // since this might trigger new requests on the editor contents.
+        updateCapabilities();
```
DavyLandman, line 954: The old comment warned about `registerLanguage` and a race around that, why is the time when this is called still moved? is that race removed?

rascal-vscode-extension/src/lsp/ParameterizedLanguageServer.ts
```diff
@@ -102,8 +102,7 @@ export class ParameterizedLanguageServer implements vscode.Disposable {
             for (const doc of vscode.workspace.textDocuments) {
                 const ext = path.extname(doc.uri.path);
                 if (ext !== "" && lang.extensions.includes(ext.substring(1))) {
-                    // (Re)set the language ID to re-trigger contribution requests
-                    await vscode.languages.setTextDocumentLanguage(doc, "plaintext");
+                    // Associate open editors with the given extensions with this language server
                     await vscode.languages.setTextDocumentLanguage(doc, this.languageId);
```
DavyLandman, line 106: That `plaintext` work around was added for a reason, what was it, and why is it not valid anymore?

I'm sure we have no tests for this edge-case? Something like a file already open in VS Code when it starts, already assigned to parametric, but the language is not registered yet?

DavyLandman (changes requested):
Great! I have some small concerns.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
toinehartman: We cannot, unfortunately, since if the opened file has an opaque URI (e.g. `untitled:Untitled-1`), `getLocation` will give a wrapped URI (`opaque-lsp-://untitled/Untitled-1`), and we lose the scheme. Of course, we could check whether `file.getLocation().getScheme()` is the opaque scheme (in which case we retrieve the original scheme from the authority), but that feels to dependent on the implementation of the wrapping of opaque URIs.

rascal-vscode-extension/src/lsp/ParameterizedLanguageServer.ts
toinehartman: It was added here: https://github.com/usethesource/rascal-language-servers/pull/809. I do not remember considering/discussing the `refresh*` requests, so we might just have missed those. I'll add a test for the case that you describe.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
toinehartman: I'll replace it with a thread-safe collection. I don't think we need to use the hashmap explicitly; there are [CopyOnWriteArraySet](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/CopyOnWriteArraySet.html) and [ConcurrentSkipListSet](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/ConcurrentSkipListSet.html).

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
toinehartman: The race is not removed. It did not have to do with the order of operations within `registerLanguage`, but with multiple calls to `updateCapabilities` at the same time/from different threads. The comments have now been updated to reflect this.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/CapabilityRegistration.java
```diff
@@ -287,15 +290,43 @@ private <T> void handleError(@Nullable Throwable t, AbstractDynamicCapability<T>
...
+        // This allows specific capabilities to add specific selectors if needed.
```
DavyLandman: what if that capability forgets the scheme filters? is that okay?

DavyLandman (approved):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/capabilities/CapabilityRegistration.java
toinehartman: Rewrote this to always add the extension- & scheme-based selectors to any selectors that the specific capability might have added. I am not yet aware of any capability that would need specific selectors that other capabilities do not need.

DavyLandman (approved):

PR #1122, opened 2026-06-16
Create analyzer feature for rascal

### LINKED ISSUE #1121: Add analyzer for rascal LSP server
For DSLs we've added (see  #356) support for quick typecheckers/lints called an analyzer, and expensive checks on save called builders. For rascal there are certain checks we could also do much faster&frequent, without waiting for the typechecker to trigger.

Every successful parse (so no errors) we trigger an analyzer, that triggers a rascal module that (ideally only once) visits the parse tree and reports warnings/errors on them. Ideally with quick fixes.

The first consumer would be the annotation rewrite code (introduced in #1099). Another one would be a fix on the filename vs module header.

Wait for the typechecker to complete.

DavyLandman:
Fixes #1121


rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/Analyzer.rsc
```diff
@@ -0,0 +1,83 @@
...
+                    )
+                ]
+            );
+        }
+        // since the single quickfix fixes the whole project, it's not useful to report it multiple times
+        // especially since a user might click "fix all" and then the upgrade would be run several times.
+        // instead we ignore all follow up cases of the same error
+    }
+
+    top-down-break visit (tree) {
```
toinehartman: I am not sure if `break` is wise here. What if ` project://` location appears as a key in the argument to `delAnnotations`?

toinehartman (commented):
Nice improvement!

Note that there are a couple of files with unused imports (reported by SQ) and the tests fail.

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/Analyzer.rsc
DavyLandman: Yeah, fair. I wanted to be quick..

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileResolver.java
```diff
@@ -26,21 +26,24 @@
...
 import org.rascalmpl.uri.ISourceLocationInput;
+import org.rascalmpl.uri.ISourceLocationOutput;
 import org.rascalmpl.uri.URIResolverRegistry;
 import org.rascalmpl.uri.URIUtil;
 import org.rascalmpl.vscode.lsp.TextDocumentState;
 import org.rascalmpl.vscode.lsp.util.Versioned;
 import io.usethesource.vallang.ISourceLocation;
 
-public class LSPOpenFileResolver implements ISourceLocationInput {
+public class LSPOpenFileResolver implements ISourceLocationInput, ISourceLocationOutput {
```
rodinaarssen: NP: there's `ISourceLocationInputOutput` for this

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileResolver.java
```diff
@@ -135,4 +138,32 @@ public FileAttributes stat(ISourceLocation uri) throws IOException {
         );
     }
 
+    private static IOException buildError(ISourceLocation loc) {
+        return new IOException("lsp+ is not a writable scheme. If you want to write to an file open in the editor us IDEServices::applyFileSystemEdits. Loc: " + loc);
```
rodinaarssen:
```suggestion
        return new IOException("lsp+ is not a writable scheme. If you want to write to a file open in the editor use IDEServices::applyFileSystemEdits. Loc: " + loc);
```

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Versioned.java
```diff
@@ -82,4 +88,41 @@ public static <T> boolean replaceIfNewer(AtomicReference<@PolyNull Versioned<T>>
...
+     * @param debouncedReplacementValue in case a debounce happened, this value will be reported on the completable future
+     * @param exec executor to use for the futures
+     *
+     * @return a future that supplies the calculated result if the current
+     * request was granted, or an debounceReplacementValue if the calcution was skipped due
+     * to a new version before delay was passed.
+     */
+    public <U> CompletableFuture<Versioned<U>> debounce(
+        AtomicReference<Versioned<T>> activeVersion, Duration delay,
+        Function<T, CompletableFuture<U>> calculation, U debouncedReplacementValue, Executor exec) {
```
sungshik: Nit: the name `debouncedReplacementValue` wasn't entirely self-explanatory to me ("replacement" could also refer to replacing the value stored in this `Versioned` object). How about renaming `calculation` and `debounceReplacementValue` to:
  - `onSuccess` and `onFailure` (not great because the debounce didn't really fail, though)
  - `onFire` and `onNotFire` (not "catchy", so to speak)
  - `whenFired` and `whenDiscarded` ("discard" is a term used in the MDN [glossary](https://developer.mozilla.org/en-US/docs/Glossary/Debounce))

Related nit: For generality (not needed right now, but how knows), `debouncedReplacementValue` could also be a `Function<T, CompletableFuture<U>>`

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/Analyzer.rsc
```diff
@@ -0,0 +1,83 @@
...
+    void reportAnnotationDeprecation(Tree t) {
+        if (!annotationAlreadyReported && !isWritable(t.src.top)) {
+            // not reporting issues for code we cannot rewrite
+            annotationAlreadyReported = true;
+        }
+
+        if (!annotationAlreadyReported) {
+            annotationAlreadyReported = true;
+            result += warning(
+                "annotations are no longer supported and will soon be removed, please use our build-in quick-fix to refactor them into keyword parameters",
```
rodinaarssen:
```suggestion
                "Annotations are no longer supported and will soon be removed, please use our built-in Quick Fix to refactor them into keyword parameters",
```

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/FileFacts.java
```diff
@@ -140,8 +149,10 @@ public void close(ISourceLocation file) {
 
     private interface FileFact {
         void reportParseErrors(Versioned<List<Diagnostic>> msgs);
+        void reportAnalyzeMessages(Versioned<List<Diagnostic>> msgs);
```
sungshik, line 152: Nit: For consistency with `reportParseErrors` and `reportTypeCheckerErrors`, I would be in favor of calling this `reportAnalyzeErrors` (even if the list of diagnostics might also contain warnings/infos; I suppose that is the same for the type checker).

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/Analyzer.rsc
```diff
@@ -0,0 +1,83 @@
...
+                t.src, fixes=[
+                    action(
+                        command=upgradeAnnotations(getPathConfig(t.src.top)),
+                        title="Upgrade all annotations to keyword fields in this project (annotation syntax is no longer supported)."
+                    )
+                ]
+            );
+        }
+        // since the single quickfix fixes the whole project, it's not useful to report it multiple times
```
rodinaarssen:
```suggestion
        // since the single Quick Fix fixes the whole project, it's not useful to report it multiple times
```

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/Analyzer.rsc
```diff
@@ -0,0 +1,83 @@
...
+        // especially since a user might click "fix all" and then the upgrade would be run several times.
+        // instead we ignore all follow up cases of the same error
```
rodinaarssen:
```suggestion
        // instead we ignore all follow-up cases of the same error
```

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/FileFacts.java
```diff
@@ -208,9 +229,28 @@ public void invalidate() {
...
+                .thenAccept(this::reportAnalyzeMessages);
+        }
+
+        private CompletableFuture<List<Diagnostic>> analyzer(Versioned<ITree> t) {
+            return rascal.analyze(t.get())
+                .thenApply(m -> Diagnostics.translateDiagnostics(m, cm))
+                .get();
+        }
+
+        private boolean noMessages() {
```
sungshik, line 245: 👍

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileResolver.java
sungshik: I missed why/where this is needed in this PR (and not very familiar with the resolver code), so just to be on the safe side, is it intentionally part of this PR or accidentally committed?

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
```diff
@@ -62,16 +62,20 @@ describe('IDE', function () {
...
+        // make sure we always reset, so errors don't propagate (since sometimes an afterEach is not run)
+        for (const [f, b] of originalFiles) {
+            await fs.writeFile(f, b);
+        }
     });
 
     afterEach(async function () {
+        for (const [f, b] of originalFiles) {
+            await fs.writeFile(f, b);
+        }
```
sungshik: I vaguely remember that there might be something about restoring the original files **after** `ide.cleanup` instead of before. Or maybe I misremember and it was about something else. Anyway, could be worth it to quickly double-check if it might matter.

rodinaarssen (approved):
LGTM.

SQ reports unused imports.

Could we add a test for the `lib` scheme detection, perhaps nested in an annotation expression?

sungshik (approved):
:+1:

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileResolver.java
DavyLandman: It's fixing some bugs, but I'm moving this to a separate PR.

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
DavyLandman: True, I've moved them into a finally, as I just want to make sure they're not missed.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileResolver.java
DavyLandman: Has been moved to #1125

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/model/FileFacts.java
DavyLandman: I've instead renamed the typechecker one, since I agree, they should be the same style.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Versioned.java
DavyLandman: Good point! I really disliked the names. I've chose, `onFire` and `onDiscard` how about that?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileResolver.java
DavyLandman: I've moved this to #1125

DavyLandman:
Note, it depends on the bugs fixed in #1125 before the test would pass.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/model/ParametricFileFacts.java
```diff
@@ -257,61 +258,16 @@ public void remove() {
...
-                }
-            }, delayed);
-
-            return summary.thenCompose(Function.identity());
-        }
-
         @Override
-        public void calculateAnalyzer(CompletableFuture<Versioned<ITree>> tree, int version, Duration delay) {
-            latestAnalyzerAnalysis = debounce(version, latestVersionCalculateAnalyzer, delay, () -> {
+        public void calculateAnalyzer(CompletableFuture<Versioned<ITree>> tree, Versioned<String> contentVersion, Duration delay) {
```
jurgenvinju, line 262: nice!

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/Versioned.java
```diff
@@ -82,4 +88,41 @@ public static <T> boolean replaceIfNewer(AtomicReference<@PolyNull Versioned<T>>
...
+     * meantime (in which case the current request is abandoned)
+     * @param onFire the actual calculation
+     * @param onDiscard in case a debounce happened, this value will be reported on the completable future
+     * @param exec executor to use for the futures
+     *
+     * @return a future that supplies the calculated result if the current
+     * request was granted, or an debounceReplacementValue if the calcution was skipped due
+     * to a new version before delay was passed.
+     */
+    public <U> CompletableFuture<Versioned<U>> debounce(
```
jurgenvinju, line 108: gold

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/Analyzer.rsc
```diff
@@ -0,0 +1,82 @@
...
+
+@synopsis{A fast analyzer, is run on most parse trees, so it should be fast}
+list[Message] analyze(start[Module] tree, PathConfig(loc file) getPathConfig) {
+    result = [];
+
+    annotationAlreadyReported = false;
+    void reportAnnotationDeprecation(Tree t) {
+        if (!annotationAlreadyReported) {
+            annotationAlreadyReported = true;
+            if (isWritable(t.src.top)) {
```
jurgenvinju, line 47: this should be checked by seeing if the current file is in `pcfg.srcs`. whether a file is writable or not has more underlying assumptions.

jurgenvinju: the rewriter rewrites all files in pcfg.srcs that contain annotations, so that has the same check. In fact, I think the test is not necessary because of this.

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/Analyzer.rsc
```diff
@@ -0,0 +1,82 @@
...
+                // only reporting issues for code we can rewrite
+                result += warning(
+                    "Annotations are no longer supported and will soon be removed, please use our build-in Quick Fix to refactor them into keyword parameters",
+                    t.src, fixes=[
+                        action(
+                            command=upgradeAnnotations(getPathConfig(t.src.top)),
```
jurgenvinju, line 53: here you see that the current file is not an input to `upgradeAnnotations`

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/Analyzer.rsc
jurgenvinju: if we don't want to report the warning in library code then let's check: `relativize(pcfg.srcs, t.src.top).scheme ==  "relative"` or a call to `find`.

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/Actions.rsc
```diff
@@ -47,10 +47,9 @@ The commands must be evaluated by ((evaluateRascalCommand))
 data Command
     = visualImportGraphCommand(PathConfig pcfg)
     | sortImportsAndExtends(Header h)
-    | upgradeAnnotations(PathConfig pcfg)
+    | upgradeAnnotations(PathConfig pcfg) // reported by the analyzer itself
```
jurgenvinju, line 50: remove comment

jurgenvinju: can follow def-use interactively to find out where it is coming from

jurgenvinju (commented):
I like it!
* [ ] remove unused imports
* [ ] remove check for writable, replace for existance under pcfg.srcs

PR #1123, opened 2026-06-17
Always unregister before registration

toinehartman:
This is enabled by #630.


rascal-lsp/src/main/rascal/library/testing/lang/pico/LanguageServer.rsc
```diff
@@ -129,6 +129,7 @@ set[LanguageService] testingLanguageServerSlowSummary() = testingLanguageServerS
 set[LanguageService] testingLanguageServerSlowSummaryWithRecovery() = testingLanguageServerSlowSummary(true);
 
 void register(bool errorRecovery=false) {
+    unregisterLanguage("Pico", {"pico", "pico-new"});
```
DavyLandman, line 134: let's add a small comment that explains that we're doing this to make sure that all versions (with or without error recovery) are cleared before we register a new one and possibly have a collision.

DavyLandman (approved):

sungshik (approved):

PR #1125, opened 2026-06-18
LSP Open File resolvers needs to be writeable (but throw errors on write)

DavyLandman:
While implementing a few rascal features (namely #1122 ) we figured out the `lsp+` scheme needed a bit more features to correctly work with.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileResolver.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileResolver.java
@@ -133,7 +133,7 @@ public class LSPOpenFileResolver implements ISourceLocationInputOutput {
         var current = state.getCurrentContent();
         return new FileAttributes(
             true, true,
-            current.getTimestamp(), onDiskStat.created(), //We fix the creation timestamp to be equal to the last modified time
+            onDiskStat.created(), current.getTimestamp(),
             true, onDiskStat.isWritable(),
             size(current)
         );
@@ -150,7 +150,7 @@ public class LSPOpenFileResolver implements ISourceLocationInputOutput {
 
     @Override
     public boolean isWritable(ISourceLocation uri) throws IOException {
-        // this communicates that yes, it's a writeable file system
+        // this communicates that yes, it's a writable file system
         // so if you indeed use `applyFileSystemEdits` it should succeed.
         return getEditorState(uri).getAttributesOnDisk().isWritable();
     }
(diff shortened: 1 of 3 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileResolver.java
```diff
@@ -116,23 +120,53 @@ private int size(Versioned<String> s) {
...
+    }
+
+    @Override
+    public OutputStream getOutputStream(ISourceLocation uri, boolean append) throws IOException {
+        throw buildError(uri);
+    }
+
+    @Override
+    public boolean isWritable(ISourceLocation uri) throws IOException {
+        // this communicates that yes, it's a writeable file system
```
rodinaarssen:
```suggestion
        // this communicates that yes, it's a writable file system
```

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileResolver.java
```diff
@@ -116,23 +120,53 @@ private int size(Versioned<String> s) {
...
-        var modified = current.getTimestamp();
-        var isWritable = URIResolverRegistry.getInstance().isWritable(stripLspPrefix(uri));
+        var state = getEditorState(uri);
+        var onDiskStat = state.getAttributesOnDisk();
+        var current = state.getCurrentContent();
         return new FileAttributes(
             true, true,
-            modified, modified, //We fix the creation timestamp to be equal to the last modified time
-            true, isWritable,
+            current.getTimestamp(), onDiskStat.created(), //We fix the creation timestamp to be equal to the last modified time
```
rodinaarssen: This comment is no longer applicable

rodinaarssen: These two argument should be flipped

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileResolver.java
```diff
@@ -116,23 +120,53 @@ private int size(Versioned<String> s) {
...
+
+    @Override
+    public void remove(ISourceLocation uri) throws IOException {
+        throw buildError(uri);
+    }
+
+    @Override
+    public void setLastModified(ISourceLocation uri, long timestamp) throws IOException {
+        throw buildError(uri);
+    }
```
rodinaarssen, line 171: Perhaps add in overloads for `getWritableOutputStream`, `rename`, and `copy`?

rodinaarssen (changes requested):
Some small remarks.

The call site of the `TextDocumentState` constructor needs to be updated for the build to succeed.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileResolver.java
DavyLandman: those all have defaults in the interface that will assure they never get called (as the boolean checks return false)

rodinaarssen (approved):

PR #1128, opened 2026-06-19
Run the validator in more cases such that we do not get stuck notifications

DavyLandman:
This should prevent some of the cancelled CI runs.

### DIFF
--- a/rascal-vscode-extension/src/ux/RascalMFValidator.ts
+++ b/rascal-vscode-extension/src/ux/RascalMFValidator.ts
@@ -64,6 +64,7 @@ export class RascalMFValidator implements vscode.Disposable {
         // we use the filesystem watcher since a user might not have the RASCAL.MF open, but we still want to warn them
         // note that warnings/errors are only cleared at a save
         const watcher = vscode.workspace.createFileSystemWatcher("**/" + MF_FILE, true, false, false);
+        watcher.onDidCreate(this.verifyRascalMF, this, this.toDispose);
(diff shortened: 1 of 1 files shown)


rascal-vscode-extension/src/ux/RascalMFValidator.ts
```diff
@@ -92,10 +93,14 @@ export class RascalMFValidator implements vscode.Disposable {
...
                 const mfBody = await vscode.workspace.openTextDocument(file);
                 const diagnostics: vscode.Diagnostic[] = [];
-                checkMissingLastLine(mfBody, diagnostics);
-                checkIncorrectProjectName(mfBody, diagnostics);
-                checkPickySeparator(mfBody, diagnostics);
-                this.diagnostics.set(file, diagnostics);
+                try {
+                    checkMissingLastLine(mfBody, diagnostics);
+                    checkIncorrectProjectName(mfBody, diagnostics);
+                    checkPickySeparator(mfBody, diagnostics);
```
sungshik, line 99: Do these checks depend on each other (as in, if one of them throws, all subsequent will necessarily also throw)? If not, they could be put in separate try/finally blocks, to avoid one failed check unnecessarily preventing another failed check.

sungshik (approved):

rascal-vscode-extension/src/ux/RascalMFValidator.ts
DavyLandman: non of them throw. I was just an extra safety in case there is some kind of edge case we forgot.

PR #1130, opened 2026-06-22
Protect files, but avoid rewriting them when they have not changed

DavyLandman:
This should reduce the changes of the race with the rascal.mf diagnostics


toinehartman (approved):
Nice improvement. Should we also use this utility for the Pico file backup in the DSL tests?

DavyLandman:
Agreed, I've done that 👍🏼

PR #1132, opened 2026-06-25
[POM-leading for DSLs] Main update 20260625

toinehartman:
Periodic merge of `main` into the long-running POM-leading feature branch.


toinehartman:
Since this requires work on the registration of dynamic capabilities to pass all tests, will do the merge directly on `feature/1010-pom-leading-for-dsls/final` and the rest of the work in #1117.

PR #1133, opened 2026-06-25
Read failed test logs from file

toinehartman:
Since we located the VS Code logs in #1075, we can also use them for the log snippets in the output of failed tests.
This PR attempts to read the logs from disk first, and only falls back to zooming/copying the output window if no logs can be found on disk. This should decrease the number of UI interactions and increase reliability of reporting on failed tests.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -50,7 +50,7 @@ jobs:
     strategy:
       matrix:
         os: [ubicloud-standard-4, windows-latest, macos-latest]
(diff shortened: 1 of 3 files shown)


toinehartman:
Note that, before merging this PR, 622cea091a0d67527f840301887f024655a28f32 should be reverted. This only serves to demonstrate how this works.

DavyLandman (approved):
Nice, no more UI interactions that can fail 👍🏼

PR #1134, opened 2026-06-29
Only run the analyzer on files open in the workspace

DavyLandman:
To avoid issues getting rapported on files we cannot rewrite/fix for a whole bunch of reasons

This is based on a review by @jurgenvinju (after #1122 was already merged) that remarked that using `isWritable` might not be the best heuristic.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
@@ -105,6 +105,7 @@ import org.eclipse.lsp4j.services.LanguageClient;
 import org.eclipse.lsp4j.services.LanguageClientAware;
 import org.rascalmpl.library.util.PathConfig;
 import org.rascalmpl.uri.URIResolverRegistry;
+import org.rascalmpl.uri.URIUtil;
 import org.rascalmpl.values.IRascalValueFactory;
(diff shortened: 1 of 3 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
```diff
@@ -245,8 +246,17 @@ public void didChange(DidChangeTextDocumentParams params) {
         triggerAnalyzer(changed, NORMAL_DEBOUNCE);
     }
 
+    private boolean isOpenInWorkspace(ISourceLocation loc) {
+        return availableWorkspaceServices().workspaceFolders()
+            .stream()
+            .map(f -> URIUtil.assumeCorrectLocation(f.getUri()))
```
toinehartman, line 252: Are `loc` and `f` always normalized/physical when possible here? So we're not checking if  `project://` is a parent of a `file://` for example?

toinehartman (approved):
Nice & compact improvement!

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
DavyLandman: yes, they're coming in from VS Code, and if `project` uris are coming from VS Code, we have bigger problems.

PR #1135, opened 2026-06-30
Test improvements and typos

toinehartman:
- Allow setting custom log level per test suite.
- Use notification tray for loading checks.
- Logging improvements.
- Small test fixes.

### DIFF
--- a/rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
+++ b/rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
@@ -29,7 +29,7 @@ import { expect } from 'chai';
 import * as fs from 'fs/promises';
 import * as path from 'path';
 import { TextEditor, until, ViewSection, VSBrowser, WebDriver, Workbench } from 'vscode-extension-tester';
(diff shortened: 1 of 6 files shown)


rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
```diff
@@ -98,7 +98,7 @@ describe('IDE', function () {
             if (statusBarSeen) {
                 console.log("Waiting for startup of rascal core");
                 for (let tries = 0; tries < 70; tries++) {
-                    if (!await checkRascalStatus()) {
+                    if (await checkRascalStatus() === false) {
```
DavyLandman, line 101: why the `=== false`?

DavyLandman (approved):

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
toinehartman: Because `checkRascalStatus()` is a `Promise<boolean, undefined>`, where `undefined` means "we could not determine whether it is loading or not".

PR #1136, opened 2026-06-30
Make sure path configs are not build on URIs we know are not legal

DavyLandman:
Sometimes (primarily caused by the `codeActions`) we get an request to build a `PathConfig` for a location where it's not legal to build a path config.

This PR at least stops the error messages that come up for the user by generating an empy path config.

Note that in the future we should change the actions code to maybe only run on workspace files.


toinehartman (commented):
Nice that you tracked this down. I'd prefer that we change the code action computation to not request a path config at all for files/projects that we cannot compute if for.

DavyLandman:
Agreed, that's why it's an error message. But for now, let's at least deal with it in a nicer way.

PR #1152, opened 2026-07-09
Replace `moveCursor` by `setCursor` in tests

### LINKED ISSUE #1150: Replace `moveCursor` with `setCursor` in UI tests
`setCursor` uses the VS Code command to jump to a position, whereas `moveCursor` presses the arrow keys one-by-one, checking the current position at each line, until it arrives at the requested position. The former is almost instant, whereas the latter might take a long time when the position to move to is far away.

**Caveat**
They have different semantics in the presence of tabs. See https://github.com/microsoft/vscode/issues/198780.

toinehartman:
Closes #1150.


toinehartman:
More performance could be gained when https://github.com/redhat-developer/vscode-extension-tester/pull/2407 is merged and released, since that will speed up the `selectText` call in `IO.rsc`, which has to scroll hundreds of lines before reaching the requested text. This will also fix `selectText` in tab-indented files.

DavyLandman (approved):

PR #1153, opened 2026-07-09
Trigger a type check for an entire Rascal project from VS Code

rodinaarssen:
This PR adds:
* a Rascal function that, given the root directory of a project, instructs the type checker to type check all Rascal files in the project (based on the PathConfig)
* context menu items to the Explorer view in VS Code to trigger said Rascal function
* machinery to connect these together

This PR may depend on a release of Rascal that contains https://github.com/usethesource/rascal/pull/2815


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricTextDocumentService.java
```diff
@@ -1051,4 +1053,9 @@ public void cancelProgress(String progressId) {
         contributions.values().forEach(plex ->
             plex.cancelProgress(progressId));
     }
+
+    public CompletableFuture<Void> checkProject(CheckProjectRequest req) {
+        // Do nothing
```
DavyLandman: maybe we should say it's not supported? maybe even throw an exception?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IBaseTextDocumentService.java
```diff
@@ -52,6 +54,7 @@ public interface IBaseTextDocumentService extends TextDocumentService, ITextDocu
     void initialized();
     void registerLanguage(LanguageParameter lang);
     void unregisterLanguage(LanguageParameter lang);
+    CompletableFuture<Void> checkProject(CheckProjectRequest req);
```
toinehartman: Why this additional, specific request to trigger this functionality? Can we not use `workspace/executeCommand` for this?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -265,18 +270,30 @@ IFunction makeParseTreeGetter(Evaluator e) {
         });
     }
 
+    private ISet getWorkspaceFolders(BaseWorkspaceService workspaceService) {
```
DavyLandman: why does this function get the workspace service as a parameter? as it's a field right?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -265,18 +270,30 @@ IFunction makeParseTreeGetter(Evaluator e) {
...
-        var workspaceFolders = workspaceService.workspaceFolders().stream().map(f -> Locations.toLoc(f.getUri())).collect(VF.setWriter());
 
         var shortModuleName = URIUtil.getLocationName(URIUtil.removeExtension(file));
         return runEvaluator("Rascal check (" + shortModuleName +")", compilerEvaluator,
-            e -> translateCheckResults((IMap) e.call("checkFile", file, workspaceFolders, makeParseTreeGetter(e), makePathConfigGetter(e))),
+            e -> translateCheckResults((IMap) e.call("checkFile", file, getWorkspaceFolders(workspaceService), makeParseTreeGetter(e), makePathConfigGetter(e))),
             Map.of(file, VF.set()), exec, false, client);
     }
 
+    public void checkProject(ISourceLocation projectRoot, IBool clean, Executor exec) {
```
DavyLandman: a bit misaligned to have `compileFile` and `checkProject` (I think maybe it should be `checkFile` as well?)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IBaseLanguageServerExtensions.java
```diff
@@ -64,10 +65,14 @@ default CompletableFuture<Two<String, URI[]>[]> supplyPathConfig(PathConfigParam
...
     @JsonRequest("vfs/schemes")
     CompletableFuture<String[]> fileSystemSchemes();
 
     @JsonRequest("vfs/logical/resolve")
     CompletableFuture<SourceLocationResponse> resolve(ISourceLocationRequest req);
+    
+    @JsonRequest
+    default CompletableFuture<Void> checkProject(CheckProjectRequest req) {
+        throw new UnsupportedOperationException();
+    }
```
DavyLandman: it should not be a default method, every implementor has to think about what to do with this request.

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/IDECheckerWrapper.rsc
```diff
@@ -84,20 +73,108 @@ map[loc, set[Message]] checkFile(loc l, set[loc] workspaceFolders, start[Module]
     rel[loc, loc] dependencies = {};
 
     step("Dependency graph", 1);
-    job("Building dependency graph", bool (void (str, int) step2) {
+    <dependencyMsgs, checkedForImports, dependencies> = buildDependencyGraph(checkForImports, openFileHeader.src, workspaceFolders, getParseTree, getPathConfig);
+
+    if ([] != dependencyMsgs) {
+        // Since we only reported errors on `l`, there is not need to analyze to which files the errors belong here.
```
DavyLandman, line 79: how is this true? can't we have error from our dependencies?

rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -313,12 +315,42 @@ export class RascalExtension implements vscode.Disposable {
...
+            const rascalMF = vscode.Uri.joinPath(projectRoot.uri, "META-INF", "RASCAL.MF");
+            if (!await this.fileExists(rascalMF)) {
+                void vscode.window.showErrorMessage(`Rascal check project: ${projectRoot} is not a Rascal project`);
+            }
+            try {
+                await (await this.rascal.rascalClient).sendRequest("rascal/checkProject", <CheckProjectRequest>{
+                    loc: toRascalUri(projectRoot.uri),
+                    clean: clean
+                });
+                void vscode.window.showInformationMessage(`Rascal check project: finished checking project ${projectRoot.uri}`);
```
DavyLandman: Is this true, is this only showed when it's finished?

DavyLandman (commented):
Very nice! I have some small remarks. But the most important one is that I'm missing a UI test for this feature.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/IBaseTextDocumentService.java
toinehartman: Particularly since we are extending a generic document service interface here, for something that we only intend to support for Rascal

rascal-vscode-extension/src/RascalExtension.ts
rodinaarssen: No, this line was not intended to make it into the PR

rascal-vscode-extension/package.json
```diff
@@ -59,6 +59,16 @@
...
+      {
+        "command": "rascalmpl.checkProject",
+        "title": "Rascal: check project",
+        "icon": "./assets/images/rascal-logo-v2.1.svg"
+      },
+      {
+        "command": "rascalmpl.cleanAndCheckProject",
+        "title": "Rascal: clean and check project",
+        "icon": "./assets/images/rascal-logo-v2.1.svg"
+      },
```
toinehartman, line 71: Very useful actions! The titles are a bit out of line with the existing commands. Maybe all of them should be `Rascal: ...`.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -265,18 +270,30 @@ IFunction makeParseTreeGetter(Evaluator e) {
...
+        logger.debug("Check Rascal project: {}", projectRoot);
+
+        var shortName = URIUtil.getLocationName(projectRoot);
+        runEvaluator("Rascal check project (" + shortName +")", compilerEvaluator,
```
toinehartman: Floating future. I think it should be returned, and the `supplyAsync` in `checkPorject` be replaced by a direct return.

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/IDECheckerWrapper.rsc
```diff
@@ -84,20 +73,108 @@ map[loc, set[Message]] checkFile(loc l, set[loc] workspaceFolders, start[Module]
...
                 for (i <- tree.top.header.imports, i has \module) {
                     modName = "<i.\module>";
                     for (ml <- locateRascalModules(modName, getPathConfig(currentProject), getPathConfig, workspaceFolders)) {
-                        if (<mlpt, importErrors> := getParseTreeOrErrors(ml, modName, openFileHeader.src)) {
+                        if (<mlpt, importErrors> := getParseTreeOrErrors(ml, modName, errorLocation, getParseTree)) {
                             if ({} !:= importErrors) {
-                                parseErrors += importErrors;
-                                checkedForImports += currentSrc; // do not check this module again
-                                continue; // since there is an error in this module, we do not recurse into its imports
+                                msgs += [*importErrors];
```
toinehartman: Maybe `getParseTreeOrErrors` could return a `list` of messages instead of a `set`, since I don't think we're using the properties of a set anywhere.

rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -313,12 +315,42 @@ export class RascalExtension implements vscode.Disposable {
...
+            if (!projectRoot) {
+                void vscode.window.showErrorMessage(`Rascal check project: could not deduce the project root directory for ${uri}`);
+                return;
+            }
+            const rascalMF = vscode.Uri.joinPath(projectRoot.uri, "META-INF", "RASCAL.MF");
+            if (!await this.fileExists(rascalMF)) {
+                void vscode.window.showErrorMessage(`Rascal check project: ${projectRoot} is not a Rascal project`);
+            }
+            try {
+                await (await this.rascal.rascalClient).sendRequest("rascal/checkProject", <CheckProjectRequest>{
```
toinehartman, line 335: As said in another comment: I think we could use `workspace/executeCommand` here, and save the addition of the specific endpoint. Downside is that we need to build an IValue-like string.

toinehartman (commented):
Nice feature! Looks good. I don't love the additional document service endpoint, since it's only for Rascal (and does not mean anything for parametric or the to-be-added routing server).

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/IDECheckerWrapper.rsc
rodinaarssen: This is the "check-on-save" path, where errors have never been reported on dependencies, but on the file itself

rascal-vscode-extension/package.json
rodinaarssen: The prefix is primarily there since it appears as a context menu entry. I agree that it's favorable to align the command names, but let's not do that in this PR.

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/IDECheckerWrapper.rsc
rodinaarssen: I made everything into a set, as there is no notion of order

rascal-vscode-extension/src/RascalExtension.ts
rodinaarssen: That's certainly a possibility; however, this requires special-casing in the `executeCommand` implementation in Java (i.c., check the command, redirect to a different evaluator). In a private discussion, it was agreed to not take this route at this point.

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/IDECheckerWrapper.rsc
rodinaarssen: Scratch that, the checker itself uses a list; I made everything into lists now

toinehartman (approved):

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
```diff
@@ -291,4 +292,29 @@ describe('IDE', function () {
...
+        await workspace.expand();
+        const projectRoot = <TreeItem> await workspace.findItem("test-project");
+        await (await projectRoot!.openContextMenu()).select("Rascal: clean and check project");
+        const repl = new RascalREPL(bench, driver);
+        await repl.start();
+        await repl.execute("import IO;");
+        await driver.wait(async () => {
+            await repl.execute("exists(|target://test-project/rascal/$Main.tpl|)");
+            return repl.lastOutput === "bool: true";
+        }, Delays.extremelySlow, "tpl for Main.rsc should exist by now");
```
DavyLandman, line 318: Why do we need to run this in the REPL?

For different test we directly use IO operations to check for a tpl file (the `TestWorkspace` has constants for the corresponding tpl locations).

rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -313,12 +315,41 @@ export class RascalExtension implements vscode.Disposable {
...
+    }
+
+    private checkProjectCommand(clean: boolean) {
+        return async (uri: vscode.Uri) => {
+            const projectRoot = vscode.workspace.getWorkspaceFolder(uri);
+            if (!projectRoot) {
+                void vscode.window.showErrorMessage(`Rascal check project: could not deduce the project root directory for ${uri}`);
+                return;
+            }
+            const rascalMF = vscode.Uri.joinPath(projectRoot.uri, "META-INF", "RASCAL.MF");
```
DavyLandman, line 330: I think we already have this logic, the whole is tehre a rascal.mf file thing? Can we reuse that?

DavyLandman (approved):
It looks good, I only think we can simplify the test a bit to avoid starting a REPL just to check for a tpl file.

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
```diff
@@ -292,7 +293,26 @@ describe('IDE', function () {
         await ide.checkNoDiagnosticsAnymore();
     });
 
-    it ("full file formatting works", async() => {
+    it("check project works", async function () {
+        // Context menu does not work on macOS
+        if (os.type() === "Darwin") {
+            this.skip();
+        }
+        // Fix type error to avoid a failing "after each" hook in CI
```
DavyLandman, line 301: does this comment still hold? like what is the type error?

rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -313,12 +316,41 @@ export class RascalExtension implements vscode.Disposable {
...
```
DavyLandman: this join path is not replaced yet?

DavyLandman (approved):
Two small things, if that is fixed/clarified, ready to  merge I would say.

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
rodinaarssen: This is to account for the "use editor content of open files" test, for which the `Importee.rsc` module is intentionally empty "on disk", but is filled with a declaration that `Importer.rsc` uses. To ensure that the project type check succeeds, this declaration is saved "on disk".

DavyLandman: hmm, but the protected files should take care of that right?

rascal-vscode-extension/src/RascalExtension.ts
rodinaarssen: I seem to have fixed the other occurrence of this pattern in the same file. Fixed, thanks

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
rodinaarssen: The file being protected means that after/before each test, its content is restored

PR #1154, opened 2026-07-10
Improved caching behavior by a small pre-job that just seeds the caches

DavyLandman:
Based on discussion in https://github.com/actions/setup-java/issues/705

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -15,12 +15,45 @@ permissions:
   contents: write
   
 jobs:
-  normal-tests:
+  seed-caches:
     runs-on: ubicloud-standard-2
     steps:
       - uses: actions/checkout@v7
-      - name: Assure different pom cache hit
-        run: echo "  " >> rascal-lsp/pom.xml
+      - uses: actions/setup-java@v5
+        id: java
+        with:
+          java-version: 11
+          distribution: 'temurin'
+          cache: 'maven'
+
+      - name: setup node
+        id: node 
+        uses: actions/setup-node@v6
+        with:
+          node-version: 20.x
+          cache: 'npm'
+          cache-dependency-path: rascal-vscode-extension/package-lock.json
+          registry-url: 'https://registry.npmjs.org'
+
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yaml
```diff
@@ -47,6 +74,7 @@ jobs:
           npm run test:normal
       
   ui-test:
+    needs: [seed-caches]
```
toinehartman, line 83: Since the OS of `seed-caches` is different to the OSes in the matrix, the cache key differs, I don't think that cache will ever be hit here. The UI tests do benefit from caches from a previous PR (if the POM/`package.json`) did not change, though.

toinehartman (commented):
Nice improvement! I am wondering if this fully works as expected.

For the UI tests, the cache is not found, since the OS is different to the one of the `seed-caches.

Image

.github/workflows/build.yaml
DavyLandman: True, I was just looking at that. And i've changed to OS, to only use the one that is shared (namely the `ubicloud-standard-2`, all the others only run the specific case they're interested in, so no problem if they fill their own caches).

And the lint also fills the `ubuntu-latest`.

.github/workflows/build.yaml
```diff
@@ -15,12 +15,39 @@ permissions:
...
+          node-version: 20.x
+          cache: 'npm'
+          cache-dependency-path: rascal-vscode-extension/package-lock.json
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: mvn download
+        working-directory: ./rascal-lsp
+        run: |
+          mvn -B dependency:go-offline dependency:resolve-plugins
+          mvn -Pchecker-framework -B dependency:go-offline dependency:resolve-plugins
```
toinehartman, line 44: Do we need `checker-framework` here? Isn't that only used in the separate `lint` run?

DavyLandman: ah yes :) i was copying too much.

toinehartman (approved):

PR #1156, opened 2026-07-13
[POM-leading for DSLs] Server lifetime, bugs and simplification

### LINKED ISSUE #1086: Fine-grained control of remote server lifetime
Work out details around remote server lifetime, e.g.:
- Whether/when to quit remote servers.
- How to clean up progress bars when remote servers quit/crash.

toinehartman:
- Re-enable `exitWhenEmpty`, backwards compatible with old versions. Make `registerLanguage` robust against dying remote servers.
- Initialize remote servers synchronously. This allows getting rid of servers that could not be initialized and removes the need of using futures in many places.
- Forward workspace changes to remotes.
- Other small fixes.

Closes #1086.


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -189,36 +205,60 @@ private static boolean isRascalLsp(Artifact art) {
...
+    private static Pair<@Nullable ComparableVersion, List<Path>> rascalLspDependencies(PathConfig pcfg, List<Artifact> mavenDependencies) {
+        // When loading a language server within the Rasal LSP project (e.g. in tests), we do not have a dependency on/JAR of LSP.
+        // Instead, we use its compiled classes and the JARs of all its dependencies.
+        var target = Path.of(Locations.toUri(Locations.toPhysicalIfPossible(pcfg.getBin())));
+        var depPaths = mavenDependencies.stream()
+            .map((Function<Artifact, @Nullable Path>) Artifact::getResolved)
+            .filter(Objects::nonNull)
+            .collect(Collectors.<@NonNull Path>toList());
+
+        var lspVersion = ActualRoutingLanguageServer.class.getPackage() == null
```
rodinaarssen: How can this be `null`? Does that happen when running from within `rascal-lsp`? Or is this just for CF

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -189,36 +205,60 @@ private static boolean isRascalLsp(Artifact art) {
...
+            .map(c -> c.getVersion())
+            .findFirst()
+            .map(ComparableVersion::new);
+
+        return Pair.of(
+            lspVersion.orElse(null),
+            deps.stream()
+                .filter(d -> isRascal(d) || isRascalLsp(d)) // We only need Rascal and LSP to run the language server
+                .map((Function<Artifact, @Nullable Path>) Artifact::getResolved)
+                .filter(Objects::nonNull)
```
rodinaarssen: Are we certain that both Rascal and rascal-lsp are available? Or is this just for CF

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -262,22 +309,28 @@ private void forwardLogs(InputStream logStream, String langName) {
...
+                , "-Drascal.remoteResolverRegistryPort=" + System.getProperty("rascal.remoteResolverRegistryPort")
+                , "-Drascal.customRemoteResolverRegistryClass=" + System.getProperty("rascal.customRemoteResolverRegistryClass")
+                , "-Xmx2048M"
+                , "-cp", classPath
+                , "org.rascalmpl.vscode.lsp.parametric.ParametricLanguageServer"
+            ));
+
+            var lspVersion = dependencies.getLeft();
+            if (lspVersion == null || lspVersion.compareTo(new ComparableVersion("2.22.6")) >= 0) {
+                serverArgs.add("--exitWhenEmpty");
```
rodinaarssen: Is it "bad" if this flag is used for older versions of rascal-lsp?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -409,7 +470,41 @@ public void write(JsonWriter writer, ProxiedIValue proxiedValue) throws IOExcept
...
+        logger.info("Static capabilities of {} are different to the ones registered with the client, which might lead to missing functionality.", language);
+        var routerVersionString = initialization.getServerInfo().getVersion();
+        var remoteVersionString = remoteInitialization.getServerInfo().getVersion();
+        if (routerVersionString != null && remoteVersionString != null) {
+            // Deployment mode; we can give the user useful hints here
+            var routerVersion = new ComparableVersion(routerVersionString);
+            var remoteVersion = new ComparableVersion(remoteVersionString);
+            if (routerVersion.compareTo(remoteVersion) > 0) {
+                // routerVersion > remoteVersion
+                logger.info("{} depends on version {} of Rascal-LSP, which is older than the version in the extension ({}).");
```
rodinaarssen: Missing arguments

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -409,7 +470,41 @@ public void write(JsonWriter writer, ProxiedIValue proxiedValue) throws IOExcept
...
+            } else if (routerVersion.compareTo(remoteVersion) < 0) {
+                // routerVersion < remoteVersion
+                logger.info("{} depends on version {} of Rascal-LSP, which is newer than the version in the extension ({}). Some DSL features might not be fully supported. To fix this, update the extension.", language, remoteVersion, routerVersion);
+            } else {
+                logger.debug("Router and remote version are equal ({}), but the capabilities differ. This should never happen. Please report this as a bug via https://github.com/usethesource/rascal-language-servers/issues/new?template=bug_report.md", routerVersion);
```
rodinaarssen: Perhaps the log level should be increased if this is actually to be reported

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
```diff
@@ -251,44 +245,46 @@ public void initialized() {
 
     @Override
     public void registerLanguage(LanguageParameter lang) {
-        // Nothing to do here
+        throw new UnsupportedOperationException("registerLanguage() should not be called on the routing document service, but only on delegate document services.");
```
rodinaarssen: This could be extracted into a separate function

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
```diff
@@ -251,44 +245,46 @@ public void initialized() {
...
+        throw new UnsupportedOperationException("executeCommand() should not be called on the routing document service, but only on delegate document services.");
     }
 
     @Override
     protected DiagnosticsReporter getDiagnosticsReporter(ISourceLocation file) {
-        return (l, msgs) -> {
-            // NOP; we delegate diagnostic reporting to our dedicated remote servers
-        };
+        // NOP; we delegate diagnostic reporting to our dedicated remote servers
+        return (l, msgs) -> {};
```
rodinaarssen, line 281: Is this method intentionally not throwing an exception?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/concurrent/CompletableFutureUtils.java
```diff
@@ -34,19 +34,42 @@
...
     }
 
+    public static <T> CompletableFuture<T> retry(Supplier<T> supplier, int times, Executor exec) {
+        return retry(CompletableFuture.supplyAsync(supplier, exec), times);
+    }
+
+    public static <T> CompletableFuture<T> retry(CompletableFuture<T> future, int times) {
+        var f = future;
+        for (int i = 0; i < times; i++) {
+            var time = i + 1;
```
rodinaarssen: NP: the loop could also run from `1` to `times` instead of `0` to `times - 1`

rodinaarssen (approved):
Nice work!

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -189,36 +205,60 @@ private static boolean isRascalLsp(Artifact art) {
...
+        // Instead, we use its compiled classes and the JARs of all its dependencies.
+        var target = Path.of(Locations.toUri(Locations.toPhysicalIfPossible(pcfg.getBin())));
+        var depPaths = mavenDependencies.stream()
+            .map((Function<Artifact, @Nullable Path>) Artifact::getResolved)
+            .filter(Objects::nonNull)
+            .collect(Collectors.<@NonNull Path>toList());
+
+        var lspVersion = ActualRoutingLanguageServer.class.getPackage() == null
+            ? null
+            : ActualRoutingLanguageServer.class.getPackage().getSpecificationVersion();
```
DavyLandman: what is the value and meaning of the specification version? is this related to pom.xml?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -189,36 +205,60 @@ private static boolean isRascalLsp(Artifact art) {
...
+            .filter(d -> isRascalLsp(d))
+            .map(Artifact::getCoordinate)
+            .map(c -> c.getVersion())
+            .findFirst()
+            .map(ComparableVersion::new);
+
+        return Pair.of(
+            lspVersion.orElse(null),
+            deps.stream()
+                .filter(d -> isRascal(d) || isRascalLsp(d)) // We only need Rascal and LSP to run the language server
```
DavyLandman: what if in the future this changes? like why do we need to limit ourselfs to only these? if the pom.xml lists more, let's put them on the classpath? like what if we stop shading the rascal.jar?

ideally it should be the same logic we use for starting the rascal repl, so that we don't have yet a slightly different way to calculate the class path here.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
DavyLandman: Let's only add backwards compatible code for stuff that has been released, not previous versions of this PR?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
DavyLandman: This should be more actionable, like, is this a problem?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
DavyLandman: agreed.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -304,10 +309,14 @@ public CompletableFuture<InitializeResult> initialize(InitializeParams params) {
 
             logger.info("LSP connection started (connected to {} version {})", params.getClientInfo().getName(), params.getClientInfo().getVersion());
             logger.debug("LSP client capabilities: {}", params.getCapabilities());
-            final InitializeResult initializeResult = new InitializeResult(new ServerCapabilities());
+            var serverInfo = BaseLanguageServer.class.getPackage() == null
```
DavyLandman: 2nd place where we have this logic of extracting the version from the package, can we share this somehow?

DavyLandman (approved):
Looks good, I have some small clarification questions.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
toinehartman: Yes, since previously released versions assume the first (and only) argument is the dedicated language name:
https://github.com/usethesource/rascal-language-servers/blob/3a051818d228d5461f15ba664237ab2ea8b45993/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/ParametricLanguageServer.java#L35-L42

toinehartman: This option does not work for previously released versions of LSP since they assume 1 positional argument. This option was added and then disabled in a previous PR indeed. It is now re-enabled in f20ae9e5bbe35c57351ccca0896c3b226d52c789 but we still need to be cautious about DSLs using an older LSP version.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
toinehartman: Just for CF.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
toinehartman: Yes, it typically comes from the POM. I now extracted this logic to a function named `getPOMVersion`.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
toinehartman: Since the user typically only sees the Rascal code of util::LanguageServer, I think that the extension being of a newer version than their Rascal-LSP dependency should not really lead to unexpected behaviour (until we start deprecating certain requests). Therefore I do not think action is required here. Maybe this log should be of less severity, so most users do not see it?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
toinehartman: Extracted to `getPomVersion()`

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
toinehartman: Extracted to `unsupportedRoutingOperation`

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
toinehartman: Yes. This function is used by `TextDocumentStateManager` and should just no-op in this case.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
DavyLandman: So can we support those? Like people using a LSP version before this feature was released? Can we start such an LSP and it just works? I think there is a minimal version, and that minimal version would be whatever we end up releasing this feature.

That minimal version will have to be bumped everytime we make some backwards incompatible change.

For example the whole remote IO feature is needed, so an version of 2 releases back won't work.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
DavyLandman: Yes, so it should be a lower level, and the text should make it clear we don't consider this a problem.,

But I do think we'll have minimal supported versions, and in the case it's too old, that should be an error.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
toinehartman: Put all libs on the class path ✅

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
toinehartman: Added a minimal compatible version check for the routing-remote server pair

toinehartman: Added a minimal version check that requires at least the current LSP version for the remote server/in the DSL POM

rascal-vscode-extension/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 We only list significant changes, for a full changelog please review the [commit history](https://github.com/usethesource/rascal-language-servers/commits/main/).
 
+## 0.13.6
```
DavyLandman: I expect this to become 0.14.0 it's quite a change.

DavyLandman (approved):
Yes, this looks good 👍🏼

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/RoutingTextDocumentService.java
```diff
@@ -146,38 +145,31 @@ private DocumentRouter<CompletableFuture<IBaseLanguageServerExtensions>> availab
...
-        } catch (InterruptedException e) {
-            Thread.currentThread().interrupt();
-        } catch (ExecutionException | TimeoutException e) {
-            logger.catching(e);
-        }
+        allRoutes().forEach(s -> {
+            try {
+                notify.accept(s, params);
+            } catch (Exception e) {
+                logger.error("Unexpected error while notifying all remote servers", e);
```
rodinaarssen, line 160: Would it be useful to let the error message contain which particular remote failed?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/LanguageRegistry.java
```diff
@@ -55,6 +55,22 @@ public interface LanguageRegistry {
     @JsonRequest("receiveUnregisterLanguage")
     CompletableFuture<Void> unregisterLanguage(LanguageParameter lang);
 
+    /**
+     * Extensible arguments related to language registration.
+     */
+    class RegistrationParameter {
```
rodinaarssen, line 61: Missing `toString`/`equals`

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -181,44 +198,42 @@ private static String extension(ISourceLocation doc) {
...
+        if (lsp.isPresent()) {
+            return Pair.of(new ComparableVersion(getJarVersion(lsp.get())), classPath);
         }
 
-        return deps.stream()
-            .filter(d -> isRascal(d) || isRascalLsp(d))
-            .map((Function<Artifact, @Nullable Path>) Artifact::getResolved)
-            .filter(Objects::nonNull)
-            .collect(Collectors.toList());
+        throw new IOException("No rascal-lsp dependency. Could not start language server");
```
DavyLandman, line 232: will this show up for a user in some way?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -228,71 +243,104 @@ private void forwardLogs(InputStream logStream, String langName) {
...
 
+    /**
+     * @throws IOException when the log writer is closed.
+     */
+    private void forwardLogLine(String langName, String line) throws IOException {
+        try {
+            var json = JsonParser.parseString(line);
+            prependThreadName(langName, json);
+            // Lock, so we can make sure our JSON is followed by a newline.
+            synchronized (System.err) {
```
DavyLandman: we should save this in a constant in this class, such that we dont have `system.err` spread everywhere. like where we construct the `logForwarder` is where we should save the reference to the underlying stream, and at the point respect that field.

(imagine if someone changes the stream that is passed into the `logForwarder`)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
```diff
@@ -228,71 +243,104 @@ private void forwardLogs(InputStream logStream, String langName) {
...
+    }
+
+    private static ISourceLocation tryResolveMavenJar(MavenRepositoryURIResolver mvn, ISourceLocation loc) {
+        if (!loc.getScheme().equals("mvn")) {
+            return loc;
+        }
+
+        try {
+            return mvn.resolveJar(loc);
+        } catch (IOException e) {
```
DavyLandman, line 286: let's at least log this exception for future debugging.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -299,8 +308,34 @@ public CompletableFuture<Two<String, URI[]>[]> supplyPathConfig(PathConfigParame
...
+            if (specVersion == null) {
+                throw new IOException("No specification version found for running LSP");
+            }
+            return specVersion;
+        }
+
+        protected static String getJarVersion(ISourceLocation jarFile) throws IOException {
+            jarFile = JarURIResolver.jarify(jarFile);
+            var manifestLoc = URIUtil.getChildLocation(jarFile, "META-INF/MANIFEST.MF");
+            var manifest = new Manifest(URIResolverRegistry.getInstance().getInputStream(manifestLoc));
```
DavyLandman: shouldn't the input stream be put in a try-with-resources block?

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -211,7 +211,13 @@ function calculateRascalMemoryReservation() {
     return "-Xmx800M";
 }
 
-function calculateDSLMemoryReservation(_dedicated: boolean) {
+export function calculateDSLMemoryReservation(isRouter: boolean) {
+    if (!isRouter) {
+        // The server is just a router. It needs ~650MB of base memory and hardly grows per language.
+        // We stay on the safe side.
+        return "-Xmx800M";
+    }
```
DavyLandman: shouldn't this be `if (isRouter)`?

toinehartman: Good catch. This was `isDedicated` before and not properly refactored.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/ActualRoutingLanguageServer.java
toinehartman: Extracted all occurrences to a single constant, and added comments to the logger configurations.

PR #1158, opened 2026-07-17
Redirect resolved URI to LSP based on original and resolved URIs

toinehartman:
Fixes a bug where `std://` would not be redirected to LSP, but `mvn://` would.

Since `std://` is resolved to `jar+file://` (in most cases) before checking if the file is managed by LSP, the `isFileManaged` check alway returns false: the open file key is `std://`, not `jar+file://`. This fix makes sure that both the original URI (as coming in from VS Code) as well as the resolved URI are checked for the need of redirecting to LSP.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileRedirector.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileRedirector.java
@@ -31,9 +31,9 @@ import java.io.IOException;
 import java.net.URISyntaxException;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
-
 import org.checkerframework.checker.initialization.qual.UnderInitialization;
 import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
+import org.rascalmpl.uri.URIResolverRegistry;
 import org.rascalmpl.uri.URIUtil;
 import org.rascalmpl.vscode.lsp.IBaseTextDocumentService;
 import org.rascalmpl.vscode.lsp.TextDocumentState;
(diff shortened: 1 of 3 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileRedirector.java
```diff
@@ -65,18 +65,26 @@ public boolean isFileManaged(ISourceLocation file) {
         return false;
     }
 
-    public ISourceLocation resolve(ISourceLocation input) {
-        if (isFileManaged(input)) {
-            try {
+    /**
+     * Redirect a URI to it's LSP-managed counterpart. Considers both the original and resolved URI.
```
rodinaarssen:
```suggestion
     * Redirect a URI to its LSP-managed counterpart. Considers both the original and resolved URI.
```

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/uri/LSPOpenFileRedirector.java
```diff
@@ -65,18 +65,26 @@ public boolean isFileManaged(ISourceLocation file) {
...
+     * @param resolved The resolution of the original URI.
+     * @param original The URI as it came in from VS Code (possibly logical). This is the one that will often be registered as the 'managed' file.
+     * @return A redirected URI if the file is open in LSP. If not return the resolved URI if non-null, and the original otherwise.
+     */
+    public ISourceLocation redirect(@Nullable ISourceLocation resolved, ISourceLocation original) {
```
rodinaarssen, line 74: That's definitely a better name

rodinaarssen (approved):

PR #1159, opened 2026-07-20
Remove `std` scheme

### LINKED ISSUE #1027: LSP version of extension not present in Maven directory
**Describe the bug**

Currently, the LSP checks and runs DSL projects against its embedded LSP version (in https://github.com/usethesource/rascal-language-servers/blob/fix/simplify-fact-computation/rascal-lsp/pom.xml). This version does not necessarily exist in the local Maven repository (which is only modified based on the project's pom.xml, which might have a different version).

This leads to strange behaviour and missing file errors.

**To Reproduce**

1. Create a project where the `rascal-lsp` version is different than the version of the current extension.
2. Clear the `rascal-lsp` JARs in the local Maven repo.
3. Open & type-check the project.
4. Observe errors.

**Expected behavior**
No errors. The extension should make sure that specific version exists and download it otherwise.

 - Context: VS Code extension
 - Rascal Version: `rascal-lsp 0.13.2`

### LINKED ISSUE #1147: Use standard library version from Maven dependencies in REPL
Change REPL (in VS Code) to respect the pom.xml and use the rascal from there (but check the version is new enough)

### LINKED ISSUE #1148: Quick fix projects without Rascal dependency in POM
Add a check/quickfix in vs code that helps users setup their rascal version in the pom.xml in case it's not there, or outdated.

### LINKED ISSUE #1149: Analyzer error on use of `std://` in location literals

### LINKED ISSUE #1151: Start VS Code REPL in Rascal project
- Class path uses `file://` path pointing to Rascal target folder + dependencies from Rascal POM.
- Source path uses Rascal source folders.

toinehartman:
Companion to https://github.com/usethesource/rascal/pull/2828

Closes #1151.
Closes #1147.
Closes #1148.
Closes #1149.
Closes #1027.

Includes the following PRs:
- #1163 
- #1168 
- #1169 
- #1185 
- #1189


rascal-lsp/src/test/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServicesTest.java
```diff
@@ -47,6 +46,7 @@ private void moduleNameTest(ISourceLocation sourcePath, String modPath, String m
 
     @Test
     public void stdModuleName() throws URISyntaxException {
+        // TODO Replace with locations that do not use the sunset std scheme
```
DavyLandman: looks like a TODO to still pickup?

rascal-vscode-extension/src/test/vscode-suite/dsl.test.ts
```diff
@@ -442,4 +438,28 @@ end
         // Dynamically registered capability
         await ide.hasSyntaxHighlighting(editor, Delays.slow);
     });
+
+    it("uses the standard library from LSP (TODO: the project POM)", async function () {
+        if (errorRecovery) { this.skip(); }
```
DavyLandman: needs to be updated wrt to changes on `main`.

rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -175,13 +173,28 @@ export class RascalExtension implements vscode.Disposable {
...
+                let rascalClasses: string[] | undefined = undefined;
+
+                if (uri !== undefined) {
+                    const lookupRes = await rascal.sendRequest<SourceLocationListResponse>("rascal/lookupRascalClasses", <ISourceLocationRequest>{
+                        loc: toRascalUri(uri)
+                    });
+
+                    rascalClasses = lookupRes.locs !== undefined
+                        ? lookupRes.locs.map(l => vscode.Uri.parse(l).fsPath)
+                        : undefined;
```
DavyLandman: nit: I think we can use the `?` deref operator. `lookupRes.locs?.map(l => vscode.Uri.parse(l).fsPath)`

rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -308,9 +321,11 @@ export class RascalExtension implements vscode.Disposable {
         );
         return shellArgs.concat(extraArgs || []);
     }
-    private buildTerminalJVMPath() :string {
-        const jars = ['rascal-lsp.jar', 'rascal.jar'];
-        return jars.map(j => path.join(this.jarRootPath, j)).join(path.delimiter);
+
+    private buildTerminalJVMPath(rascalClasses: string[] | undefined): string {
+        const lsp = path.join(this.jarRootPath, 'rascal-lsp.jar');
```
DavyLandman: I think this is wrong, as the  rascal-lsp dependency should come from the pom.xml not from the extension.

rascal-vscode-extension/package.json
```diff
@@ -43,11 +43,13 @@
...
         "command": "rascalmpl.runMain",
-        "title": "Create New Rascal Terminal, Import Module, and Run Main Function"
+        "title": "Create New Rascal Terminal, Import Module, and Run Main Function",
+        "enablement": "editorLangId == rascalmpl"
       },
       {
         "command": "rascalmpl.importModule",
-        "title": "Create New Rascal Terminal and Import Module"
+        "title": "Create New Rascal Terminal and Import Module",
+        "enablement": "editorLangId == rascalmpl"
```
DavyLandman: Why is this change in this PR?

toinehartman: Because I was looking for a way to start a REPL in the Pico project from the LSP testing setup, and figured, why can we not open a REPL in a project from a language file (instead of only a Rascal file)? I can find another way to open this REPL and offer these changes in a separate PR, since I still think they are valuable.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/BaseLanguageServer.java
```diff
@@ -399,23 +413,7 @@ public CompletableFuture<SourceLocationResponse> resolve(ISourceLocationRequest
...
-                        resolved = URIResolverRegistry.getInstance().logicalToPhysical(loc);
-                    } catch (IOException ignored) {
-                        logger.trace("Resolving {} failed, but we ignored it", loc, ignored);
-                    }
-                }
-                if (resolved == null) {
-                    resolved = loc;
-                }
-                return new SourceLocationResponse(resolved);
+                return new SourceLocationResponse(Locations.toPhysicalIfPossible(loc));
```
DavyLandman, line 416: 🎉

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -44,6 +44,26 @@ export async function activateLanguageClient(
...
+
+    // function casingMapKey(uri: vscode.Uri): string {
+    //     const normalizedScheme = uri.scheme.toLowerCase();
+    //     const normalizedAuthority = uri.authority.toLowerCase();
+    //     return `${normalizedScheme}/${normalizedAuthority}`;
+    // }
+
+    // // This regular expression does not match opaque URIs. That is not a problem, since those never have an authority to correct
+    // const uriMatcher = /^([^:]*):\/\/[^/]*\/(.*)$/;
+
```
DavyLandman: this can be dropped as well I think?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
```diff
@@ -155,6 +160,51 @@ public RascalTextDocumentService(ExecutorService exec) {
...
+    private static boolean isRascal(ISourceLocation loc) {
+        loc = Locations.toPhysicalIfPossible(loc);
+        return "rascal".equals(new RascalManifest().getProjectName(loc));
+    }
+
+    @Override
+    public List<ISourceLocation> lookupRascalClasses(ISourceLocation forFile) throws IOException, ModelResolutionError, URISyntaxException {
+        var pcfg = availableFacts().getPathConfig(forFile);
+        var mvn = new MavenRepositoryURIResolver(URIResolverRegistry.getInstance());
+
```
DavyLandman, line 173: what if this code is run while a user had a `mvn://...rascal.../IO.rsc` open?

rascal-vscode-extension/package.json
DavyLandman: since a DSL is not rascal? These 2 commands are specifically to open the current module (and optionally run it).

So is the goal to hide these 2 (and keep `Create New Rascal Terminal`) because they cannot work there?

DavyLandman (commented):
looking good

rascal-vscode-extension/src/RascalExtension.ts
toinehartman: Ths Rascal LSP dependency does not need to be on this class path at all. Fixed.

rascal-vscode-extension/package.json
toinehartman: The goal is to hide them indeed (instead of just being no-op, like it is currently). Together with this, I changed the `Create New Rascal Terminal` to work on non-Rascal files in Rascal projects.

toinehartman: I agree that it's confusing to do this in this PR, and we can change the tests to not need this. Let's discuss on a dedicated PR. https://github.com/usethesource/rascal-language-servers/pull/1184

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
toinehartman: Good catch, that won't work. The above was written with source projects in mind. For libraries, we should probably still start the REPL with the path config from the source project that depends on that library?

DavyLandman: Maybe not offer to start a REPL on a non `file` scheme?

rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -178,16 +179,35 @@ export class RascalExtension implements vscode.Disposable {
...
+                const options: vscode.TerminalOptions = {
                     iconPath: this.icon,
                     shellPath: await getJavaExecutable(this.log),
-                    shellArgs: await this.buildShellArgs(remoteIDEServicesConfiguration),
+                    shellArgs: await this.buildShellArgs(remoteIDEServicesConfiguration, rascalClasses),
                     isTransient: false, // right now we don't support transient terminals yet
                     name: `Rascal terminal (${this.getTerminalOrigin(uri, command??"")})`,
-                });
+                };
+                if (!uri) {
```
DavyLandman, line 216: Perhaps we should say that if there is only a single project open I the workspace, it's fine to open that one. As that would at least be predictable, and the same as is happening right now.

rascal-vscode-extension/test-workspace/test-lib/pom.xml
```diff
@@ -29,7 +29,7 @@
     <dependency>
       <groupId>org.rascalmpl</groupId>
       <artifactId>rascal</artifactId>
-      <version>0.40.3-RC3</version>
+      <version>0.43.0-RC13-PREVIEW-NO-STD</version>
```
DavyLandman: Shouldn't this also be updated to latest rc?

rascal-vscode-extension/test-workspace/test-project/pom.xml
```diff
@@ -29,13 +30,18 @@
     <dependency>
       <groupId>org.rascalmpl</groupId>
       <artifactId>rascal</artifactId>
-      <version>0.40.3-RC3</version>
+      <version>0.43.0-RC13-PREVIEW-NO-STD</version>
```
DavyLandman: Same here, update?

runUItests.sh
```diff
@@ -7,16 +7,24 @@
 set -e;
 set -x;
 
-cd rascal-vscode-extension
+# Find the current Rascal LSP version in the POM, and restore it when this script exits
+RESTORE_LSP_VERSION=$( ( cd rascal-vscode-extension/test-workspace/test-project && mvn dependency:tree -Dincludes=org.rascalmpl:rascal-lsp | grep rascal-lsp | cut -d ':' -f 4 ) )
+restore_versions() {
+    cd .. && ./update-test-dependencies.sh "$RESTORE_LSP_VERSION"
+}
+trap "restore_versions" EXIT
```
DavyLandman: We should also add `INT HUP TERM` otherwise it's only executed on normal exit.

update-test-dependencies.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Sets the version of Rascal LSP in the test project(s) to the first argument of this script. If no argument is given, it uses the current version of Rascal LSP from the POM.
+
+# Find the current version of Rascal LSP
+LSP_VERSION=${1:-$( ( cd rascal-lsp && mvn help:evaluate -Dexpression=project.version -q -DforceStdout ) )}
+
+# Set the version of Rascal LSP in the test project
+( cd rascal-vscode-extension/test-workspace/test-project && mvn versions:use-dep-version -q -Dincludes=org.rascalmpl:rascal-lsp -DdepVersion=$LSP_VERSION -DprocessProperties )
```
DavyLandman, line 9: Nice that this isn't a sed anymore 👍🏼

DavyLandman (approved):
Some small things to cleanup, but apart from that, let's merge

rascal-vscode-extension/test-workspace/test-lib/pom.xml
rodinaarssen: Not strictly required, but better indeed

PR #1161, opened 2026-07-22
Various code improvements

toinehartman:
- Clean up some unused declarations.
- Re-use checks and getters.

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
@@ -125,12 +125,12 @@ import org.rascalmpl.vscode.lsp.rascal.conversion.Message;
 import org.rascalmpl.vscode.lsp.rascal.conversion.SelectionRanges;
 import org.rascalmpl.vscode.lsp.rascal.conversion.SemanticTokenizer;
 import org.rascalmpl.vscode.lsp.rascal.model.FileFacts;
-import org.rascalmpl.vscode.lsp.rascal.model.SummaryBridge;
 import org.rascalmpl.vscode.lsp.uri.LSPOpenFileRedirector;
 import org.rascalmpl.vscode.lsp.util.Versioned;
 import org.rascalmpl.vscode.lsp.util.concurrent.CompletableFutureUtils;
 import org.rascalmpl.vscode.lsp.util.locations.Locations;
 import org.rascalmpl.vscode.lsp.util.locations.impl.TreeSearch;
+
 import io.usethesource.vallang.IConstructor;
 import io.usethesource.vallang.IList;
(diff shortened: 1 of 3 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
```diff
@@ -263,9 +263,7 @@ public void didClose(DidCloseTextDocumentParams params) {
         logger.debug("Close: {}", params.getTextDocument());
         var loc = Locations.toLoc(params.getTextDocument());
         closeFile(loc);
-        if (facts != null) {
-            facts.close(loc);
-        }
+        availableFacts().close(loc);
```
rodinaarssen, line 266: This (and similar locations) is now throwing if `facts == null`, is that intended?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/Locations.java
```diff
@@ -161,7 +157,7 @@ public static ISourceLocation toLoc(URI uri) {
      */
     public static URI toUri(ISourceLocation loc) {
         var uri = loc.getURI();
-        if (OPAQUE_SCHEME.equals(uri.getScheme())) {
+        if (isWrappedOpaque(loc)) {
```
rodinaarssen, line 160: 👍

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/Locations.java
```diff
@@ -66,8 +64,6 @@ public class Locations {
     // https://datatracker.ietf.org/doc/html/rfc3986#section-3.1
     private static final String OPAQUE_SCHEME = "opaque-lsp-";
 
-    public static final Set<String> TRANSLATED_SCHEMES = Set.of(OPAQUE_SCHEME, LSPOpenFileResolver.LSP_OPEN_SCHEME);
```
rodinaarssen, line 69: Is this set no longer relevant?

rodinaarssen (approved):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/locations/Locations.java
toinehartman: Indeed, it was introduced in #1101, but its use was refactored out before merging the PR.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalTextDocumentService.java
toinehartman: Yes. The `facts` should be permanently set when the client is connected, which happens before any document requests come in.

Issue #244, opened 2023-04-04
Project name suddenly uncapitalized at unknown location


jurgenvinju:
**Describe the bug**

Somewhere between summary construction and declaration resolution (for jump-to-def) I have `project://CapitalName` being renamed to `project://capitalName` which wreaks havoc on the editor features.

The bug is not in typepal. It happens purely when the summary is computed from the TModel and requested by the LSP and then used by the LSP.

jurgenvinju:
Ok it happens only after the "definition" message arrives at the client, still with the correct capitalization.

jurgenvinju:
Current hypothesis:
* URI is converted somewhere to a URL
* authority becomes a hostname because of that
* hostname is canonicalized to lowercase
* URL is either converted back to URI or used to open the editor directly

This would explain the behavior. Now looking for the proof.

jurgenvinju:
Ok. it's both worse and better than expected. `vscode.Uri` implements RFC 3986 (the URI standard), which says this:

<https://www.rfc-editor.org/rfc/rfc3986#page-38:~:text=6.2.2.1.%20%20Case%20Normalization>

> [6.2.2.1](https://www.rfc-editor.org/rfc/rfc3986#section-6.2.2.1).  Case Normalization
> For all URIs, the hexadecimal digits within a percent-encoding
> triplet (e.g., "%3a" versus "%3A") are case-insensitive and therefore
>   should be normalized to use uppercase letters for the digits A-F.
>
>   When a URI uses components of the generic syntax, the component
>   syntax equivalence rules always apply; namely, that the scheme and
>   host are case-insensitive and therefore should be normalized to
>   lowercase.  For example, the URI <HTTP://www.EXAMPLE.com/> is
>   equivalent to <http://www.example.com/>.  The other generic syntax
>   components are assumed to be case-sensitive unless specifically
>   defined otherwise by the scheme (see [Section 6.2.3](https://www.rfc-editor.org/rfc/rfc3986#section-6.2.3)).

Apparently `java.util.URI` does _not_ implement this part of the RFC, while `vscode.Uri` _does_.

Since all project URI go through the VFS system with our own VFS providers, they all must go through the `vscode.Uri` representation and thus both the schemes and the authorities are normalized to lowercase.

jurgenvinju:
* workaround: add an error/warning in the IDE for projects with non-lowercase names
[...]

PR #1162, opened 2026-07-22
Only compute path configs for files in workspace folders

### LINKED ISSUE #1160: Summary for `mvn://` location triggers path config computation
**Describe the bug**

When opening a `mvn` location, for example through go-to-def on an import, a path config is computed in order to make the summary. However, #1048 intended to prevent this (since path configs are only valid for source projects), and instead compute the summary using a known TPL location within the JAR.

**To Reproduce**

1. Depend on Rascal from https://github.com/usethesource/rascal/pull/2828
2. Typecheck a module with `import IO;`.
3. Go-to-def on `IO`.
4. Control-click something and observe the path config failing to compute.

**Expected behavior**
Go to the next def.

 - Context: VS Code extension
 - Rascal Version: LSP `2.22.6-SNAPSHOT`

**Additional context**
Ran into this while working on https://github.com/usethesource/rascal/pull/2828

toinehartman:
Closes https://github.com/usethesource/rascal-language-servers/issues/1160

### DIFF
--- a/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
+++ b/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
@@ -134,6 +134,13 @@ public class RascalLanguageServices {
         this.workspaceService = workspaceService;
     }
 
+    public boolean isOpenInWorkspace(ISourceLocation loc) {
+        return workspaceService.workspaceFolders()
+            .stream()
+            .map(f -> URIUtil.assumeCorrectLocation(f.getUri()))
+            .anyMatch(f -> URIUtil.isParentOf(f, loc));
+    }
+
     static String pathToModuleName(ISourceLocation l) {
         var p = l.getPath();
         if (isInsideJar(l)) {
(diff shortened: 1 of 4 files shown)


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/rascal/RascalLanguageServices.java
```diff
@@ -134,6 +134,13 @@ public RascalLanguageServices(RascalTextDocumentService docService, BaseWorkspac
         this.workspaceService = workspaceService;
     }
 
+    public boolean isOpenInWorkspace(ISourceLocation loc) {
+        return workspaceService.workspaceFolders()
+            .stream()
+            .map(f -> URIUtil.assumeCorrectLocation(f.getUri()))
+            .anyMatch(f -> f.toString().equalsIgnoreCase(loc.toString()) || URIUtil.isParentOf(f, loc));
```
DavyLandman: why is this `equalsIgnoreCase` on the `ISourceLocation`? don't we have equals for this?

And how come `isParentOf` does not need to use `equalsIgnoreCase`?

It looks like the function used to be somewhere esle, but it got moved here, and the extra check got added?

toinehartman: It was indeed moved. `isParentOf` works for files within workspace folders, but fails for workspace folders themselves. Hence the extra check.

The reason for the `equalsIgnoreCase` is unexpected case changes by VS Code (see: https://github.com/usethesource/rascal-language-servers/issues/244). However, in the meantime, we fixed that in #1168. I'll simplify this check to use `equals`.

toinehartman: Whether `URIUtil.isParentOf` is correct given the case-insensitivity of certain URI parts (scheme and authority) is a good question.

DavyLandman (approved):

PR #1166, opened 2026-07-29
Skip beforeEach for skipped UI tests

toinehartman:
Previously, the `beforeEach` hook would run for each test, even if it would be skipped (for example because of an incompatible error recovery setting). This PR changes the test setup to skip tests statically, saving some time (at the expense of skipped tests not being show in the test result overview).


rodinaarssen (approved):

DavyLandman (approved):
Maybe after merging we should add this commit to the blame ignore file?

PR #1168, opened 2026-07-30
Added cache storing schemes and authorities of incoming URIs in VS Code

rodinaarssen:
VS Code may convert schemes and authorities to lower case (which is allowed by the standard). For some Rascal locations (`|mvn://|` and `|project://|` come to mind) this is problematic, since during resolution the authority is used as a part of the path; this is problematic on case-sensitive file systems.

This PR adds a cache storing the casing of schemes and authorities, and reinstates this casing for the authority when serializing URIs to string.

h/t @toinehartman

### DIFF
--- a/rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
+++ b/rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
@@ -95,8 +95,10 @@ export async function activateLanguageClient(
 
                 if (parsed.scheme !== "file" && parsed.authority !== "") {
                     const cachedCasing = getCachedCasing(parsed);
-                    if (cachedCasing && !(cachedCasing[0] === parsed.scheme && cachedCasing[1] === parsed.authority)) {
-                        logger.warn(`Received scheme/authority casing |${parsed.scheme}://${parsed.authority}/| that is inconsistent with earlier |${cachedCasing[0]}://${cachedCasing[1]}/|. Round-tripping locations correctly is likely not possible`);
+                    if (cachedCasing) {
+                        if (!(cachedCasing[0] === parsed.scheme && cachedCasing[1] === parsed.authority)) {
+                            logger.warn(`Received scheme/authority casing |${parsed.scheme}://${parsed.authority}/| that is inconsistent with earlier |${cachedCasing[0]}://${cachedCasing[1]}/|. Round-tripping locations correctly is likely not possible`);
+                        }
                     } else {
                         storeCasing(parsed);
                     }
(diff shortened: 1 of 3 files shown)


rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -70,13 +72,17 @@ export async function activateLanguageClient(
                 // The casing of the schemes and autorities of incoming URIs is therefore stored, so the original casing
                 // can be restored when round-tripping (i.e., Rascal => VS Code => Rascal).
                 code2Protocol: (uri: vscode.Uri) : string => {
+                    const uriString = uri.toString();
                     const cachedCasing = getCachedCasing(uri);
 
                     if (cachedCasing) {
-                        uri = uri.with({ scheme: cachedCasing[0], authority: cachedCasing[1] });
+                        const match = uriMatcher.exec(uriString);
```
toinehartman, line 82: Nice! I think this could use a comment on why we do not use `uri.with`.

rascal-vscode-extension/src/lsp/RascalLSPConnection.ts
```diff
@@ -60,6 +60,8 @@ export async function activateLanguageClient(
         return `${normalizedScheme}/${normalizedAuthority}`;
     }
 
+    const uriMatcher = /^[^:]*:\/\/[^/]*\/(.*)&/;
```
toinehartman: This pattern does not match opaque URIs (like `untitled:Untitled-1`. Those never have an authority, so it's not that big of a deal. The scheme might still be normalized bij VS Code, although I hope we never introduce a non-lowercase scheme anywhere, so we shouldn't be vulnerable. At least, we should maybe document this trade-off as well.

toinehartman (approved):
Nice! The tests succeed again :). I only have some suggestions to document a bit more why we're doing this.

toinehartman:
Oh, and additionally, I think we can now remove the String-based `equalsIgnoreCase` workaround in the `SummaryBridge` in favor of just a hard equals between locations.

toinehartman:
This will also solve the unexpected uncapitalization reported in https://github.com/usethesource/rascal-language-servers/issues/244

toinehartman (approved):

DavyLandman:
I've already shared feedback on this to @toinehartman, but this work around can be broken in cases where the URI does not come from rascal-lsp side first. For example a click on a loc in the REPL, or a VS Code restart with files open, than a new `didOpen` is send with an URI that has not been seen before.

PR #1169, opened 2026-07-30
Tests for version/location of stdlib

toinehartman:


rodinaarssen (approved):

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
```diff
@@ -393,8 +392,6 @@ public static CompletableFuture<Evaluator> makeFutureEvaluator(LSPContext contex
                     services
                 );
 
-                eval.addClassLoader(RascalLanguageServer.class.getClassLoader());
```
DavyLandman, line 396: Why are we removing this as part of the `std://` PR? This is the typechecker&friends runners

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
```diff
@@ -165,7 +181,14 @@ describe('IDE', function () {
...
         const editor = await ide.openModule(TestWorkspace.mainFile);
-        await triggerTypeChecker(editor, TestWorkspace.mainFileTpl, true);
+        const output = await captureOutput("Rascal MPL Language Server",
+            () => triggerTypeChecker(editor, TestWorkspace.mainFileTpl, true));
+
+        const pcfg = matchPathConfig(output);
+
+        // Find Rascal version in POM
+        const pomRascalVersion = await getArtifactVersion("org.rascalmpl", "rascal", TestWorkspace.testProjectPom);
+        expect(pcfg.libs ?? []).to.include(`|mvn://org.rascalmpl--rascal--${pomRascalVersion}|`);
```
DavyLandman, line 191: can we add a message that is shown in case it isn't there?

rascal-vscode-extension/src/RascalExtension.ts
```diff
@@ -163,10 +163,6 @@ export class RascalExtension implements vscode.Disposable {
                 progress.report({message: "Starting rascal-lsp"});
                 const rascal = await this.rascal.rascalClient;
                 this.log.debug(`Starting Rascal REPL: on ${uri} and with command: ${command}`);
-                if (uri && !uri.path.endsWith(".rsc")) {
-                    // do not try to figure out a rascal project path when the focus is not a rascal file
-                    uri = undefined;
-                }
```
DavyLandman, line 169: why is this removed? this avoids an error when trying to build a REPL for a non rsc file?

DavyLandman (commented):
This PR has some changes that suprise me a bit based on the title.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/EvaluatorUtil.java
toinehartman: Because the evaluator also already has class loaders based on the libs in the path config. We already have the proper Rascal and LSP versions in the project path config, on which the class loaders will be based. This loader that was added here takes precedence over the others, which means that the Rascal version that LSP is built with takes precedence over the one from the path config. This is not correct for interpreted language contributions, which also use this function to build their evaluator.

rascal-vscode-extension/src/RascalExtension.ts
toinehartman: I wanted to be able to start REPLs from DSL files in a Rascal DSL project. We can also compute the project root from those, and this works fine for those cases (as is tested by the tests added here). Of course this does not work for the `Create Rascal terminal and import...` actions. Hence those are now only shown on Rascal files:
https://github.com/usethesource/rascal-language-servers/blob/b55a12872e33c8713c7c8d2522ca6231c8ee6664/rascal-vscode-extension/package.json#L44-L53

rascal-vscode-extension/src/test/vscode-suite/ide.test.ts
toinehartman: Do you mean to distinguish 'missing Rascal' test failure from 'wrong Rascal' test failure?

PR #1183, opened 2026-08-25
POM validator

toinehartman:


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/xml/PomAnalyzer.java
```diff
@@ -0,0 +1,99 @@
...
+    IBool hasDependency(ISourceLocation pomLoc, String groupId, String artifactId) {
+        try {
+            var mavenParser = new MavenParser(Path.of(pomLoc.getURI()));
+            var rootProject = mavenParser.parseProject();
+            var resolvedDependencies = rootProject.resolveDependencies(Scope.COMPILE, mavenParser);
+            return vf.bool(resolvedDependencies.stream()
+                .map(Artifact::getCoordinate)
+                .filter(a -> a.getGroupId().equals(groupId) && a.getArtifactId().equals(artifactId))
+                .findAny()
+                .isPresent());
```
DavyLandman: nit: these 3 can be replaced by `anyMatch`

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/xml/PomAnalyzer.java
```diff
@@ -0,0 +1,99 @@
...
+                if (version != null) {
+                    return vf.string(version);
+                }
+            }
+        } catch (IOException e) {
+            // Fall through
+        }
+
+        var fallbackVersion = "2.22.5";
+        return vf.string(fallbackVersion);
```
DavyLandman: How about we don't have a default? but throw an exception? as we really don't want to have this arb version in our code?

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/Analyzer.rsc
```diff
@@ -31,14 +31,17 @@ import IO;
...
 import util::PathConfig;
 import lang::rascal::\syntax::Rascal;
 import lang::rascal::lsp::Actions;
+import lang::xml::PomAnalyzer;
 
 
 @synopsis{A fast analyzer, is run on most parse trees, so it should be fast}
 list[Message] analyze(start[Module] tree, PathConfig(loc file) getPathConfig) {
     result = [];
+    pcfg = getPathConfig(tree.src.top);
```
DavyLandman: the reason this was done inline was to avoid running pathconfig calcualtions for every invocation of this functions, as it's called quite a lot. and initially we don't want to run a path config.

so let's keep that cloning in there, because it's a performance concern.

rascal-lsp/src/main/rascal/lsp/lang/xml/PomAnalyzer.rsc
```diff
@@ -0,0 +1,135 @@
...
+
+    indentation = inferIndentation(pom, pomLines);
+    newline = inferNewline(pomSrc, pomLines);
+
+    gId = "\<groupId\><groupId>\</groupId\>";
+    aId = "\<artifactId\><artifactId>\</artifactId\>";
+    v = "\<version\><version>\</version\>";
+
+    str makeDependencyXml(str baseIndentation)
+        = "<baseIndentation>\<dependency\><newline><baseIndentation><indentation><gId><newline><baseIndentation><indentation><aId><newline><baseIndentation><indentation><v><newline><baseIndentation>\</dependency\><newline>";
```
DavyLandman: can we make this into a multi-line string concat? or a function that builds elements of this? as this one is a bit painfull to read.

rascal-lsp/src/main/rascal/lsp/lang/xml/PomAnalyzer.rsc
```diff
@@ -0,0 +1,135 @@
...
+    }
+    throw IllegalArgument("Invalid pom at <pomLoc>");
+}
+
+TextEdit addRascalDependency(loc pomLoc, str version=getRascalVersion()) {
+    // getRascalVersion() returns "Not specified" in case it cannot find a MANIFEST.MF file with a version for Rascal
+    // This happens in second level
+    if (version == "Not specified") {
+        version = "0.43.0-RC14";
+    }
```
DavyLandman: Shall we just insert `???` in there? as in most cases people won't be running this code in 2nd level **with a snapshot dependency on rascal**.

rascal-vscode-extension/src/lsp/RascalLanguageServer.ts
```diff
@@ -71,6 +72,36 @@ export class RascalLanguageServer implements vscode.Disposable {
...
+                        }, error => {
+                            return client.handleFailedRequest(CodeActionRequest.type, token, error, null);
+                        });
+                    };
+                    const middleware = client.middleware;
+                    return middleware.provideCodeActions
+                        ? middleware.provideCodeActions(document, range, context, token, _provideCodeActions)
+                        : _provideCodeActions(document, range, context, token);
+                }
+            };
```
DavyLandman: let's move this to a class (in this file) that has a bit of documentation why it exists, as it's a bit big to have here in the middle of this constructor.

rascal-vscode-extension/src/lsp/RascalLanguageServer.ts
```diff
@@ -71,6 +72,36 @@ export class RascalLanguageServer implements vscode.Disposable {
...
+                            return client.protocol2CodeConverter.asCodeActionResult(values, token);
+                        }, error => {
+                            return client.handleFailedRequest(CodeActionRequest.type, token, error, null);
+                        });
+                    };
+                    const middleware = client.middleware;
+                    return middleware.provideCodeActions
+                        ? middleware.provideCodeActions(document, range, context, token, _provideCodeActions)
+                        : _provideCodeActions(document, range, context, token);
+                }
```
DavyLandman: can we document why this is needed?

rascal-vscode-extension/src/lsp/RascalLanguageServer.ts
```diff
@@ -71,6 +72,36 @@ export class RascalLanguageServer implements vscode.Disposable {
...
+                            context: client.code2ProtocolConverter.asCodeActionContextSync(context)
+                        };
+                        return client.sendRequest(CodeActionRequest.type, params, token).then(values => {
+                            if (token.isCancellationRequested || values === null || values === undefined) {
+                                return null;
+                            }
+                            return client.protocol2CodeConverter.asCodeActionResult(values, token);
+                        }, error => {
+                            return client.handleFailedRequest(CodeActionRequest.type, token, error, null);
+                        });
```
DavyLandman: since the function is async, we can replace these `.then` & `.error` with a `return await` and a `try .. catch`?

rascal-lsp/src/main/rascal/lsp/lang/rascal/lsp/Analyzer.rsc
rodinaarssen: Okay, that's fair

rascal-lsp/src/main/rascal/lsp/lang/xml/PomAnalyzer.rsc
```diff
@@ -0,0 +1,141 @@
...
+    gId = "\<groupId\><groupId>\</groupId\>";
+    aId = "\<artifactId\><artifactId>\</artifactId\>";
+    v = "\<version\><version>\</version\>";
+
+    str makeDependencyXml(str baseIndentation)
+        = "<baseIndentation>\<dependency\><newline>" +
+          "<baseIndentation><indentation><gId><newline>" +
+          "<baseIndentation><indentation><aId><newline>" +
+          "<baseIndentation><indentation><v><newline>" +
+          "<baseIndentation>\</dependency\><newline>";
```
DavyLandman, line 91: much better, thanks 👍🏼

rascal-lsp/src/main/rascal/lsp/lang/xml/PomAnalyzer.rsc
```diff
@@ -0,0 +1,141 @@
...
+    if (version == "Not specified") {
+        version = "???";
+    }
+    return addDependency(pomLoc, "org.rascalmpl", "rascal", version);
+}
+
+TextEdit addRascalLspDependency(loc pomLoc, str version="???") {
+    try {
+        version = getCurrentRascalLspVersion();
+    } catch IO(_):;
```
DavyLandman, line 130: now the exception we throw in the java side is ignored here, that way the error is never shown to the user (or even logged)?

rascal-vscode-extension/src/lsp/RascalLanguageServer.ts
```diff
@@ -171,3 +179,27 @@ export interface IRascalCoordinates {
...
 }
 
+/**
+ * Custom CodeActionProvider to be able to provide code actions specifically for `pom.xml` files, without registering a full-blown LSP server for XML files.
+ */
+class PomXmlCodeActionProvider implements vscode.CodeActionProvider {
+    constructor(private readonly client: BaseLanguageClient, private readonly logger: vscode.LogOutputChannel) { }
+
+    async provideCodeActions(document: vscode.TextDocument, range: vscode.Range | vscode.Selection, context: vscode.CodeActionContext, token: vscode.CancellationToken): Promise<(vscode.CodeAction | vscode.Command)[] | null | undefined> {
+        this.logger.debug(`[PomXmlCodeActionProvider] provideCodeActions: ${document.fileName} (${range})`);
```
DavyLandman: as this happens quite a lot, let's move this to trace

DavyLandman (approved):

rascal-lsp/src/main/rascal/lsp/lang/xml/PomAnalyzer.rsc
rodinaarssen: Indeed, that was intended. The error is not directly shown to the user. `"???"` will end up in the `pom.xml` though, so it's kind of visible at least

DavyLandman (approved):

PR #1185, opened 2026-08-28
Bootstrap Rascal LSP for VS Code tests

toinehartman:
Preliminary work to properly test language projects, since https://github.com/usethesource/rascal-language-servers/pull/1159 requires language projects to have an explicit LSP dependency.

### DIFF
--- a/rascal-vscode-extension/test-workspace/test-project/pom.xml
+++ b/rascal-vscode-extension/test-workspace/test-project/pom.xml
@@ -9,7 +9,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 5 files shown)


rascal-vscode-extension/test-workspace/test-project/pom.xml
```diff
@@ -9,6 +9,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <lsp.version>2.22.6-SNAPSHOT</lsp.version>
```
toinehartman, line 12: This might hurt in local development, since this snapshot will not generally be available.

rodinaarssen (approved):

PR #1194, opened 2026-09-07
fixed some java warnings

jurgenvinju:


DavyLandman (approved):
yup, they indeed un-deprecated the `MarkedString`.

PR #1195, opened 2026-09-08
Fix flaky UI tests

toinehartman:
- Update ExTester to 8.26.1.
- Change CI test path due to length restriction by VS Code.
- Make various stability and consistency changes to tests.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -122,7 +122,7 @@ jobs:
         id: cache-vscode
         uses: actions/cache@v6
         with:
-          path: ./rascal-vscode-extension/uitests
+          path: ./uitests
           key: "vscode-${{matrix.os}}"
 
       - name: "UI test (windows)"
@@ -132,7 +132,7 @@ jobs:
           DELAY_FACTOR: 10
           RASCAL_LSP_DEV_DEPLOY: true
           _JAVA_OPTIONS: '-Xmx5G'
-        run: npx extest setup-and-run out/test/vscode-suite/*.test.js --storage uitests
+        run: npx extest setup-and-run out/test/vscode-suite/*.test.js --storage ../uitests
 
       - name: "UI test (mac)"
         shell: bash
@@ -142,7 +142,7 @@ jobs:
           DELAY_FACTOR: 15
           RASCAL_LSP_DEV_DEPLOY: true
           _JAVA_OPTIONS: '-Xmx5G'
-        run: npx extest setup-and-run out/test/vscode-suite/*.test.js --storage uitests
+        run: npx extest setup-and-run out/test/vscode-suite/*.test.js --storage ../uitests
 
       - name: "UI test (ubuntu)"
         shell: bash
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yaml
```diff
@@ -122,7 +122,7 @@ jobs:
         id: cache-vscode
         uses: actions/cache@v6
         with:
-          path: ./rascal-vscode-extension/uitests
+          path: ./uitests
```
DavyLandman, line 125: is there a reason this path change happens? (as in, the PR says it's about bumping the version dependency?)

toinehartman: The CI failed on the first commit due to this path being too long. Probably a change in the underlying dependencies. This was a quick attempt to resolve that and indeed, the title did not (yet) reflect that. I'll set this PR to draft.

DavyLandman: ah right, yeah that makes a bit of sense 👍🏼  thanks for clarifying

rascal-vscode-extension/src/test/vscode-suite/utils.ts
```diff
@@ -537,16 +539,52 @@ async function setLogLevel(logLevel: LogLevel) {
...
+                return currentClip !== originalClipboard || currentClip.length > 0;
+            } catch {
+                return false;
+            }
+        },
+        2000, 'Clipboard copy operation did not complete', 50
+    );
+    if (originalClipboard.length > 0) {
+        clipboard.writeSync(originalClipboard);
+    }
```
DavyLandman: this looks like we should have a function called: `copyClipboard` or something?

rascal-vscode-extension/src/test/vscode-suite/utils.ts
```diff
@@ -273,6 +277,7 @@ export class IDEOperations {
             await sleep(Delays.fast); // give it some time for the diagnostic to clear
         }
         expect(allVisibleMarkers, "Not all error diagnostics have been cleared").to.deep.equal([]);
+        await bottomBar.closePanel();
```
DavyLandman: shouldn't we ignore fails for this one as well?

rascal-vscode-extension/src/test/vscode-suite/utils.ts
```diff
@@ -470,6 +475,8 @@ export class IDEOperations {
     }
 
     async clickCodeLens(editor: TextEditor, name: string, timeout = Delays.slow, message = `Cannot click code lens: ${name}`): Promise<void> {
+        // Always scroll up, where the lenses typically are
+        await editor.setCursor(1, 1);
```
DavyLandman: that is not true, they can be everywhere, for example with the DSL tests. maybe it should be a param to guess the line where it should be roughly? and maybe in most cases the default of `1` is correct?

rascal-vscode-extension/src/test/vscode-suite/utils.ts
```diff
@@ -537,16 +544,25 @@ async function setLogLevel(logLevel: LogLevel) {
...
-export async function getOutput(channel: OutputChannel): Promise<string> {
-    const output = await new Workbench().getBottomBar().openOutputView();
+export async function getOutput(channel: OutputChannel, driver: WebDriver): Promise<string> {
+    const bottomBar = new Workbench().getBottomBar();
+    const output = await bottomBar.openOutputView();
     await output.selectChannel(channel);
-    return await output.getText();
+    await output.waitForStable();
+    await driver.wait(async () => (await output.getText()).length > 0, Delays.fast, "Output channel should load");
+    const text = await output.getText();
```
DavyLandman: we can also do:

```typescript
const text = await driver.wait(() => output.getText(),....
```

since `driver.wait` waits for a truthy value, and an empty string is false

rascal-vscode-extension/src/test/vscode-suite/utils.ts
```diff
@@ -563,24 +579,30 @@ function parseRascalList(input: string | undefined): Array<string> {
     return input.slice(1, -1).split(',').map(el => el.trim());
 }
 
-export function matchPathConfig(input: string) {
+export function matchPathConfig(input: string): {projectRoot: string, srcs: string[], ignores: string[], libs: string[], bin: string, resources: string[], messages: string[]} {
```
DavyLandman: let's introduce a small interface definition for this return type?

DavyLandman (approved):

rascal-vscode-extension/src/test/vscode-suite/utils.ts
toinehartman: Comment is confusing here. It should probably say; if we scroll to the top of the file, we can typically see all lenses in test files.

DavyLandman: but what if the lens is at the bottom of a long file? and I just scrolled there?

rascal-vscode-extension/src/test/vscode-suite/utils.ts
toinehartman: Rewritten. This one is a bit tricky, since it seems that sometimes, we get the text before the whole output window is loaded. Let's see if this works.

DavyLandman (approved):

PR #1200, opened 2026-09-14
[POM-leading] Capability proxy

toinehartman:
Improve maintainability and thread-safety of the client router, specifically operations related to (un)registering dynamic capabilities.

Also see:
- https://github.com/usethesource/rascal-language-servers/pull/1090
- https://github.com/usethesource/rascal-language-servers/pull/1131


rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -75,17 +77,24 @@
...
+     * with a set of registrations with those exact options. A set, since we do not care about order and do not
+     * need to consider duplicates.
+     */
+    private final Map<String, CompletableFuture<Map<Object, Set<Registration>>>> registrations = new ConcurrentHashMap<>();
+
+    /**
+     * The current registrations to the actual client.
      */
-    private final Map<String, CompletableFuture<Map<Object, List<Registration>>>> registrations = new ConcurrentHashMap<>();
+    private final Map<Pair<String, Object>, Registration> proxyRegistrations = new ConcurrentHashMap<>();
```
sungshik: Nit: Perhaps keep the same map-of-maps structure as in `registrations`?

Nit: The JavaDoc could clarify the design a bit more (i.e., the idea that capability registrations from remote LSP servers to VS Code need to be multiplexed)

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -71,25 +80,23 @@
...
     private final IBaseLanguageClient client;
     private final ExecutorService exec;
-
-    /**
-     * The current registrations.
-     * Map of method names to current registrations.
-     * The inner map is keyed by registration options, with a list of registrations with those exact options. The first registration in this list is always registered with the actual client, while the others are kept for internal administration.
-     */
-    private final Map<String, CompletableFuture<Map<Object, List<Registration>>>> registrations = new ConcurrentHashMap<>();
+    private final CapabilityRegistry capabilityRegistry;
```
DavyLandman: nit: class & field have exact the same name? so `capabilties` or `registry` is verbose enough?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -214,130 +221,385 @@ public CompletableFuture<ShowDocumentResult> showDocument(ShowDocumentParams par
...
+    *
+    * <p>
+    * To coordinate calls of {@code registerCapability} and {@code unregisterCapability} and avoid races, instances of
+    * this class internally use a basic lock-free scheduler. Essentially, the scheduler ensures that the work of each
+    * next call of {@code registerCapability} or {@code unregisterCapability} will begin only when the work of the
+    * current call, <em>including its asyncronous completion</em>, has ended. See the JavaDoc of {@link Scheduler} for
+    * details.
+    * </p>
+    */
+    class CapabilityRegistry {
```
DavyLandman: can this not be a private class?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -214,130 +221,385 @@ public CompletableFuture<ShowDocumentResult> showDocument(ShowDocumentParams par
...
+ * When a new task is submitted, and if no existing task is in progress yet, then the new task is started immediately.
+ * In contrast, if an existing task is in progress already, then the new task is started when the current task and all
+ * other pending existing tasks in the queue have signaled their completion. To this end, attempts to start tasks are
+ * made in two places: in the method to submit a new task, and in the closure that is run when an existing task has
+ * signaled its completion.
+ * </p>
+ *
+ * @param R Result type of tasks
+ */
+class Scheduler<R> {
```
DavyLandman: I think all these classes (also the helper one) should become private static classes, such that no external class will be tempted to use them.

DavyLandman: Can rename this to be more specific than schedular? Maybe: `MonotonicSchedular` or `FairSingleThreadExecutor`?

Also, can we document why `Executors.newSingleThreadExecutor()` is not the right fit?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -214,130 +221,385 @@ public CompletableFuture<ShowDocumentResult> showDocument(ShowDocumentParams par
...
+            logger.trace("Register capability {} ({}): Submitting to scheduler...", method, id);
+            return scheduler.submit(result -> {
+                var options = fromServer.getRegisterOptions();
+                var remaining = MapOfMapsOfSets.size(sentByServers, method, options);
+
+                // Case: Must forward registration
+                if (remaining == 0) {
+                    logger.trace("Register capability {} ({}): Forwarding registration to client...", method, id);
+                    var toClient = new Registration(UUID.randomUUID().toString(), method, options);
+                    forwardRegistration(toClient).whenCompleteAsync((v, t) -> {
```
DavyLandman: nit: how about we make `forwardRegistration` take the `method` and `options` as parameters and it returns the `Registration` it just created, such that the `v` in the completable future can be used by the `whenComplete`.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -214,130 +221,385 @@ public CompletableFuture<ShowDocumentResult> showDocument(ShowDocumentParams par
...
-                    var idx = regs.indexOf(unreg.get());
-                    if (idx != 0) {
-                        // This method is registered with the client, but not with this exact ID.
-                        // We remove this ID from our administration, but do not need to inform the client, since nothing changed for them.
-                        regs.remove(idx);
-                        logger.trace("Ignoring registration for {} ({}), since it is still supported by other languages.", method, u.getId());
-                        return CompletableFuture.completedFuture(currentRegs);
+                // Case: Must not forward
+                else {
+                    logger.trace("Register capability {} ({}): Not forwarding registration to client (remaining other registrations for same capability: {})", method, id, remaining);
```
DavyLandman: Can we rephrase this message such that it's clear that that is intended?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -214,130 +221,385 @@ public CompletableFuture<ShowDocumentResult> showDocument(ShowDocumentParams par
...
+
+            logger.trace("Unregister capability {} ({}): Submitting to scheduler...", method, id);
+            return scheduler.submit(result -> {
+
+                // Find the corresponding registration previously sent by a server
+                var fromServer = MapOfMapsOfSets.findAny(sentByServers, r -> matches(r, u));
+                if (fromServer == null) {
+                    var t = new IllegalStateException("Cannot unregister a capability for which no registration was sent by a server");
+                    logger.trace("Unregister capability {} ({}). Failed: {}", method, id, t);
+                    result.completeExceptionally(t);
```
DavyLandman: is that truly a failure? and not a lost race that is fine?

like if this happens, do we need to debug a problem?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -214,130 +221,385 @@ public CompletableFuture<ShowDocumentResult> showDocument(ShowDocumentParams par
...
+         * registration for that capability sent by a server. This method, together with
+         * {@link #registerCapability(Registration)}, ensures the registrations successfully received by the client are
+         * the sum of the registrations sent by the servers.
+         */
+        public CompletableFuture<Void> unregisterCapability(Unregistration u) {
+            var method = u.getMethod();
+            var id = u.getId();
+
+            logger.trace("Unregister capability {} ({}): Submitting to scheduler...", method, id);
+            return scheduler.submit(result -> {
```
DavyLandman: after reading this code and how a user needs to think about calling `result.` before exiting, I'm wondering if maybe the signature should change that the closure should return a completable future? like the arguments is the scheduler and the result is a future? that way no exit out of these functions leads to a hanging future.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -214,130 +221,385 @@ public CompletableFuture<ShowDocumentResult> showDocument(ShowDocumentParams par
...
+                // `t` are ignored; it is the responsibility of other calls on `task.result` to handle them.
+                task.result.whenComplete((v, t) -> {
+                    busy.set(false);
+                    exec.submit(this::attemptStartTask);
+                });
+                task.action.accept(task.result);
+                // Don't unset `busy` yet. Instead, doing so is the responsibility of the closure on the previous lines
+                // and should happen only when the task has signaled its completion.
+            } else {
+                busy.set(false);
```
DavyLandman: Can we document in what case can we have a scheduled `attemptStartTask` and still have an empty `task` queu?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -214,130 +221,385 @@ public CompletableFuture<ShowDocumentResult> showDocument(ShowDocumentParams par
...
+    public Scheduler(ExecutorService exec) {
+        this.exec = exec;
+        this.busy = new AtomicBoolean(false);
+        this.tasks = new ConcurrentLinkedQueue<>();
+    }
+
+    public CompletableFuture<R> submit(Consumer<CompletableFuture<R>> action) {
+        var result = new CompletableFuture<R>();
+        var task = new Task<>(action, result);
+        tasks.offer(task);
```
DavyLandman: nit: these can be inlined, and the future might be created by the task?

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -214,130 +221,385 @@ public CompletableFuture<ShowDocumentResult> showDocument(ShowDocumentParams par
...
+    private final AtomicBoolean busy;
+
+    public Scheduler(ExecutorService exec) {
+        this.exec = exec;
+        this.busy = new AtomicBoolean(false);
+        this.tasks = new ConcurrentLinkedQueue<>();
+    }
+
+    public CompletableFuture<R> submit(Consumer<CompletableFuture<R>> action) {
+        var result = new CompletableFuture<R>();
```
DavyLandman: there is a subtle bug where a future should be connected to an executure, else any next element in the chain won't happen on the executor, but on the global pool.

we've made some classes that work around this issue, maybe those are nice to use?

although if we rewrite this to always return a closure, than we get rid of that problem in this specific class.

DavyLandman (commented):
Looks good, I have some small clarififcation questions and a few improvements to the code.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
sungshik: It is truly a failure. Added some comments to clarify:
https://github.com/usethesource/rascal-language-servers/blob/715f35aadf1c5bc2259cb1f959b9a1287b20bce7/rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java#L405-L408

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/parametric/routing/MultipleClientProxy.java
```diff
@@ -214,130 +222,383 @@ public CompletableFuture<ShowDocumentResult> showDocument(ShowDocumentParams par
...
+            var id = u.getId();
+
+            logger.trace("Unregister capability {} ({}): Submitting to scheduler...", method, id);
+            return scheduler.submit(() -> {
+
+                // Find the corresponding registration previously sent by a server
+                var fromServer = MapOfMapsOfSets.findAny(sentByServers, r -> matches(r, u));
+                if (fromServer == null) {
+                    var ex = new IllegalStateException("Cannot unregister a capability for which no registration was sent by a server");
+                    logger.trace("Unregister capability {} ({}). Failed: {}", method, id, ex);
```
DavyLandman: I think the errors should be higher log level than trace? Similar in other places where something failed and we might care when debugging an issue.

rascal-lsp/src/main/java/org/rascalmpl/vscode/lsp/util/concurrent/CompletableFutureUtils.java
```diff
@@ -47,10 +47,25 @@ private CompletableFutureUtils() {/* hidden */ }
...
+    }
+
     public static <T> CompletableFuture<T> completedFuture(T value, Executor exec) {
         return CompletableFuture.supplyAsync(() -> value, exec);
     }
 
+    public static <T> CompletableFuture<T> failedFuture(Throwable ex, Executor exec) {
+        var f = CompletableFutureUtils.<T> create(exec);
+        f.completeExceptionally(ex);
+        return f;
```
DavyLandman, line 68: maybe we can also rewrite `completedFuture` with this style?

DavyLandman (approved):

